PR #50, opened 2019-03-29
DS-14 card component

dprzybylek:
image


src/interfaces/card.d.ts
```diff
@@ -0,0 +1,5 @@
+export interface ICardProps {
```
kamilmateusiak: This Card component is pretty simple so we can easily add support for html props:
`export interface ICardProps  extends React.HTMLAttributes<HTMLDivElement> {}`

Also, please add:
```
export var Card: React.ComponentType<ICardProps>;
```

src/components/Card/Card.js
```diff
@@ -0,0 +1,33 @@
...
+
+const Card = ({ img, title, children }) => {
+  const hasComponents = React.Children.count(children) >= 1;
+  const mergedClassNames = getMergedClassNames(styles[baseClass]);
+  const titleClass = classNames({
+    [`${baseClass}__title`]: hasComponents,
+    [`${baseClass}__empty`]: !hasComponents
+  });
+  return (
+    <div className={mergedClassNames}>
```
kamilmateusiak: `<div {...restProps} className={mergedClassNames}>...</div>`

src/components/Card/Card.js
```diff
@@ -0,0 +1,33 @@
...
+      <div className={styles[titleClass]}>
+        {img && <img src={img} className={styles[`${baseClass}__img`]} />}
```
kamilmateusiak: there is no way to add e.g alt to this image

src/components/Card/Card.js
```diff
@@ -0,0 +1,33 @@
...
+import styles from './style.scss';
+
+const baseClass = 'card';
+
+const Card = ({ img, title, children }) => {
+  const hasComponents = React.Children.count(children) >= 1;
+  const mergedClassNames = getMergedClassNames(styles[baseClass]);
+  const titleClass = classNames({
+    [`${baseClass}__title`]: hasComponents,
+    [`${baseClass}__empty`]: !hasComponents
```
kamilmateusiak: can you describe any use case for this empty Card?

src/components/Card/Card.md
```diff
@@ -0,0 +1,54 @@
...
+      </p>
+      <Divider />
+      <div style={{ paddingRight: "10px", display: "inline-block" }}>
+        <Button>Action 1</Button>
+      </div>
+      <Button>Action 2</Button>
+    </div>
+  </Card>
+</div>
+```
```
kamilmateusiak: please add HTML snippet

src/components/Card/style.scss
```diff
@@ -0,0 +1,51 @@
...
+  &__img {
+    height: 24px;
+    width: 24px;
+    margin-right: 10px;
+    border-radius: 4px;
+    display: inline-block;
+  }
+
+  &:last-of-type {
+    margin-bottom: 10px;
```
kamilmateusiak: hmm, could you elaborate on this?

src/components/index.js
```diff
@@ -60,3 +62,5 @@ export { Dropdown, DropdownList, DropdownListItem } from './Dropdown';
 export { PopperTooltip, CssTooltip } from './TooltipNew';
 
 export { Loader } from './Loader';
+
+export { Divider } from './Divider';
```
kamilmateusiak: Please add typescript types definitions if it's exported component

styleguide.config.js
```diff
@@ -33,6 +33,15 @@ module.exports = {
             './src/components/Toast/ToastWrapper.js'
           ]
         },
+        {
+          name: 'Card',
+          components: [
+            './src/components/Card/Card.js',
+            './src/components/Loader/Loader.js',
+            './src/components/Button/Button.js',
+            './src/components/Divider/Divider.js'
```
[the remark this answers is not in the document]
kamilmateusiak: There is no `Divider.md` so it's pointless to add it here. Also, I agree with @quarties about adding those components here.

src/components/Card/style.scss
dprzybylek: tbh it's copy paste syndrome 😅

src/components/Card/Card.js
dprzybylek: So, this case occurs in customer details, if someone doesn't provide content for the card.

src/components/index.js
dprzybylek: But it hasn't any prop types

src/components/Card/Card.js
dprzybylek: should I provide all img props or only alt?

src/interfaces/card.d.ts
dprzybylek: done at e57e61ea1e6ba66aa5c16156a890aba32e1c1e42

src/components/Card/Card.js
dprzybylek: done at e57e61ea1e6ba66aa5c16156a890aba32e1c1e42

src/components/Card/Card.md
dprzybylek: done at e57e61ea1e6ba66aa5c16156a890aba32e1c1e42

src/components/Card/style.scss
dprzybylek: done at e57e61ea1e6ba66aa5c16156a890aba32e1c1e42

styleguide.config.js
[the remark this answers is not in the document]
dprzybylek: done at e57e61ea1e6ba66aa5c16156a890aba32e1c1e42

src/components/Card/Card.md
```diff
@@ -0,0 +1,54 @@
+### Cards
+
+Cards are used to group related content together to make LiveChat easier to scan and comprehend.
+
+```js
+<div style={{ width: "300px" }}>
+  <Card title="Title goes here" img="https://via.placeholder.com/100">
+    <div>
+      <p>
+        {" "}
```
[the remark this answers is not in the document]
dprzybylek: done at e57e61ea1e6ba66aa5c16156a890aba32e1c1e42

src/components/index.js
kamilmateusiak: so you need to export a ComponentType with empty props :P

src/components/Card/Card.js
kamilmateusiak: forget about it :P there is no need. I wonder why we are forcing img usage here. Wouldn't it be useful to use svg here?

src/components/Card/Card.js
kamilmateusiak: I designer is ok with that... :P

src/components/Divider/Divider.js
```diff
@@ -0,0 +1,6 @@
+import * as React from 'react';
+import styles from './style.scss';
+
+const Divider = () => <div className={styles.divider} />;
```
kamilmateusiak: Shouldn't we use `restProps` here? For instance, there is no way to change `width` of this component. Devs will need wrapper div to handle it.

src/components/Card/Card.js
dprzybylek: In agent app we handle images, changing it now can causes issues. But I can extend this with svg, what do you think?

src/components/index.js
dprzybylek: ada5c0d4ab1a71b4d2fe5edbf16b8421399e55dd

src/components/Divider/Divider.js
dprzybylek: ada5c0d4ab1a71b4d2fe5edbf16b8421399e55dd

src/components/Card/Card.js
dprzybylek: This comes from the customer details section 🤔 maybe they don't love it, but should we change it now?

src/components/Divider/Divider.js
kamilmateusiak: can u use `getMergedClassNames` or cx here to unlock passing className without overriding our styles?

dprzybylek: done at f9f60b89b7e70ef2936e66f6786773861f232de9

src/components/Card/Card.js
kamilmateusiak: IMHO there should be  always an empty state when content is not provided, but we can't force it

src/components/Card/Card.js
kamilmateusiak: https://caniuse.com/#feat=svg-img svg in image tag is widely supported but i thing it will require some styling. Can we add `` ?

src/components/Card/Card.js
dprzybylek: What do you suggest?

src/components/Card/Card.js
dprzybylek: I checked it, it doesn't require additional styling. Example `src="https://placeholder.pics/svg/300x300/30FFFF-9458FF"`
image

kamilmateusiak: ok, let's omit those `imgProps`, 'cause this implementation is sufficient for now :)

src/components/Card/Card.js
```diff
@@ -0,0 +1,35 @@
...
+        <div className={styles[`${baseClass}__text`]}>{title}</div>
+      </div>
+      {children}
+    </div>
+  );
+};
+
+Card.propTypes = {
+  title: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
+  children: PropTypes.oneOfType([PropTypes.string, PropTypes.element])
```
kamilmateusiak: u could use `PropTypes.node` imho

src/components/Card/Card.js
```diff
@@ -0,0 +1,35 @@
...
+
+const baseClass = 'card';
+
+const Card = props => {
+  const { img, title, children, ...restProps } = props;
+  const hasComponents = React.Children.count(children) >= 1;
+  const mergedClassNames = getMergedClassNames(styles[baseClass]);
+  const titleClass = classNames({
+    [`${baseClass}__title`]: hasComponents,
+    [`${baseClass}__empty`]: !hasComponents
```
kamilmateusiak: `${baseClass}__empty` and hasComponents - leftovers :)

src/interfaces/card.d.ts
```diff
@@ -1,6 +1,6 @@
 export interface ICardProps extends React.HTMLAttributes<HTMLDivElement> {
   title: string;
-  children?: React.ReactNode | string;
+  children: React.ReactNode | string;
```
kamilmateusiak: `React.ReactNode` handles type `string`

src/interfaces/index.d.ts
```diff
@@ -1,4 +1,5 @@
 export * from "./buttons";
+export * from './card';
```
quarties: @kamilmateusiak do we want to export `Divider` interface here?

kamilmateusiak: woaaah, nice catch :) @dprzybylek can you add `export * from './divider';` here?

dprzybylek: yup, on the way

dprzybylek: done at 9f8670cd9d06aeb8e3f8825b50603d60fca22f62

PR #70, opened 2019-05-21
Handling Dropdown content resize

kamilmateusiak:
dropdown-resize

We noticed the issue in Dropdown component. When the number of dropdown list items changes or items are removed on click (screen) Dropdown is not updating its position. Dynamic content of dropdown could also trigger this bug.
We decided to listen on Dropdown content wrapper resize to properly handle those cases. 

References:
- similar issue solution in the other library: https://github.com/palantir/blueprint/pull/2718
- ResizeObserver polyfill: https://github.com/que-etc/resize-observer-polyfill
- popper issue (closed): https://github.com/FezVrasta/popper.js/issues/85 

Also I added props for custom keycodes setup and autofocus first item on creating component and list items count change (filtering)

### DIFF
--- a/package.json
+++ b/package.json
@@ -141,7 +141,8 @@
     "react-day-picker": "^7.2.4",
     "react-material-icon-svg": "1.7.0",
     "react-popper": "^1.3.3",
-    "react-transition-group": "^2.4.0"
+    "react-transition-group": "^2.4.0",
+    "resize-observer-polyfill": "^1.5.1"
   },
   "postcss": {},
   "jest": {
--- a/src/components/Dropdown/DropdownList.js
+++ b/src/components/Dropdown/DropdownList.js
@@ -9,9 +9,13 @@ import findNextFocusableItem from '../../helpers/find-next-focusable-item';
 const baseClass = 'dropdown';
 
 class DropdownList extends React.PureComponent {
-  state = {
-    focusedElement: null
-  };
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      focusedElement: props.defaultFocusedItemId
+    };
+  }
 
   componentDidMount() {
     document.addEventListener('keydown', this.onKeydown);
@@ -31,8 +35,8 @@ class DropdownList extends React.PureComponent {
       this.handleArrowKeyUse(event);
     }
 
-    if (keyCode === KeyCodes.enter) {
-      this.handleEnterKeyUse();
+    if (this.isItemSelectKeyCode(keyCode)) {
+      this.handleSelectKeyUse();
     }
   };
 
@@ -48,7 +52,7 @@ class DropdownList extends React.PureComponent {
     return this.hoverCallbacks[itemKey];
   };
 
-  handleEnterKeyUse = () => {
+  handleSelectKeyUse = () => {
     const { focusedElement } = this.state;
 
     if (focusedElement !== null) {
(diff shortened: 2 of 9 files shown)


src/components/Switch/Switch.js
```diff
@@ -111,6 +94,38 @@ class Switch extends React.PureComponent {
   }
 }
 
-export default React.forwardRef((props, ref) => (
-  <Switch innerRef={ref} {...props} />
+const basePropTypes = {
```
kamilmateusiak: @dprzybylek small refactor to make default props and prop types of Switch visible in the docs
We should avoid defining it as a static property on a component class.

src/components/Switch/Switch.js
```diff
@@ -111,6 +94,38 @@ class Switch extends React.PureComponent {
...
+  defaultOn: false,
+  onChange: noop,
+  size: 'basic',
+  name: baseClass
+};
+
+SwitchComponent.propTypes = {
+  ...basePropTypes,
+  innerRef: PropTypes.instanceOf(
+    typeof Element === 'undefined' ? () => {} : Element
```
kamilmateusiak: @dprzybylek take a look at https://github.com/livechat/design-system/pull/43
We need to figure out a better way to avoid SSR issues.

package.json
```diff
@@ -141,7 +141,8 @@
     "react-day-picker": "^7.2.4",
     "react-material-icon-svg": "1.7.0",
     "react-popper": "^1.3.3",
-    "react-transition-group": "^2.4.0"
+    "react-transition-group": "^2.4.0",
+    "resize-observer-polyfill": "^1.5.1"
```
kamilmateusiak: https://github.com/que-etc/resize-observer-polyfill 
It has really good browsers support.

src/components/Dropdown/DropdownList.js
```diff
@@ -9,9 +9,27 @@ import findNextFocusableItem from '../../helpers/find-next-focusable-item';
...
 
 class DropdownList extends React.PureComponent {
-  state = {
-    focusedElement: null
-  };
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      focusedElement: (props.items[0] && props.items[0].itemId) || null,
```
kamilmateusiak: autofocus first item on component create

src/components/Switch/Switch.js
dprzybylek: ok 👍

src/components/Dropdown/__snapshots__/DropdownList.test.js.snap
```diff
@@ -34,7 +34,7 @@ exports[`Archives | Components | FiltersMenu renders correctly 1`] = `
       />
     }
     isDisabled={false}
-    isFocused={false}
+    isFocused={true}
```
dprzybylek: Ok, so since now dropdown is default focused?
btw `Archives | Components | FiltersMenu` this test should have another name ;)
and we should avoid `renders correctly` tests

kamilmateusiak: it's related to autofocus feature, this is the first available dropdown list item (not disabled)

src/components/Dropdown/DropdownList.js
```diff
@@ -188,7 +233,12 @@ class DropdownList extends React.PureComponent {
 }
 
 DropdownList.propTypes = {
+  autoFocusOnItemsCountChange: PropTypes.bool,
   className: PropTypes.string,
+  defaultFocusedItemId: PropTypes.oneOfType([
```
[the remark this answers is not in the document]
kamilmateusiak: left over, good catch :)

PR #74, opened 2019-06-05
Adding types definitions to tabs components

kamilmateusiak:


src/components/Tab/Tab.js
```diff
@@ -28,14 +44,19 @@ const Tab = ({ onSelect, href, isSelected, children, description }) => {
 
 Tab.propTypes = {
   children: PropTypes.node.isRequired,
+  className: PropTypes.string,
   onSelect: PropTypes.func,
```
Aideann: should be required? I think so

src/components/Tab/Tab.js
```diff
@@ -28,14 +44,19 @@ const Tab = ({ onSelect, href, isSelected, children, description }) => {
...
-  description: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+  onClick: PropTypes.func,
```
Aideann: What is the `onClick` purpose? is `onSelect` not enough?

src/components/Tab/Tab.js
kamilmateusiak: Ok, what to do... I need to extend ButtonHTMLAttributes/AnchorHTMLAttributes where onSelect is already declared. I think about removing it from here cause it should not be the main handler.

src/components/Tab/Tab.js
kamilmateusiak: I think`onClick` should be the main function to handle tabs.

src/components/Tab/Tab.js
kamilmateusiak: Also It's not required 'cause `Tab` can be rendered as anchor element (there are some use cases in AA)

dprzybylek:
@kamilmateusiak exports are sorted now, that's why you have some conflicts 😅

src/components/Tab/Tab.js
Aideann: Ok then remove onSelect and leave only onClick ;) that will resolve every problem here

src/components/Tab/Tab.js
Aideann: So remove onSelect ;d

kamilmateusiak: already did it ;)

PR #77, opened 2019-06-06
Release 0.4.0

kamilmateusiak:
## v0.4.0

 _Release: 2019-05-20_

 ##### New Features

 - `Card` and `Divider` components

 ##### Refactors

 - documentation styles updated
- adding typescript types and support for other html properties to `Tabs`, `TabsWrapper` and `TabsList` components
- `footer` props of `Modal` component is now optional

 ##### Deprecations

 - `Tabs` main handler `onSelect` changed to `onClick`. Right now `onSelect` property is handling HTML `onselect` event

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,25 @@
 All notable changes to this project will be documented in this file.
 
+## v0.4.0
+
+_Release: 2019-05-20_
+
+##### New Features
+
+- `Card` and `Divider` components
+
+##### Refactors
+
+- documentation styles updated
+- adding typescript types and support for other html properties to `Tabs`, `TabsWrapper` and `TabsList` components
+- `footer` props of `Modal` component is now optional
+
+##### Deprecations
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,25 @@
 All notable changes to this project will be documented in this file.
 
+## v0.4.0
+
+_Release: 2019-05-20_
+
+##### New Features
+
+- `Card` and `Divider` components
```
dprzybylek: divider is not listed in docs (cause it was developed ad hoc), should we annouced this in changelog?

kamilmateusiak: it doesn't have a dedicated page but AFAIR it's used in `Card` page. IMHO we should include it here. What ya think?

dprzybylek: i have mixed feelings, could stay but we should provide a dedicated page in docs in the nearest future.

kamilmateusiak: Ok, I will add it tomorrow :)

PR #80, opened 2019-06-13
Passing event as argument of `onItemSelect` property

kamilmateusiak:
- we should always pass `event` as argument of event handlers in component API 
- in this case, the event helps with recognizing of keyboard or mouse usage (we need to add different amplitude event in AA for keyboard and mouse)

### DIFF
--- a/src/components/Dropdown/DropdownList.js
+++ b/src/components/Dropdown/DropdownList.js
@@ -67,7 +67,7 @@ class DropdownList extends React.PureComponent {
 
     if (this.isItemSelectKeyCode(keyCode)) {
       event.preventDefault();
-      this.handleSelectKeyUse();
+      this.handleSelectKeyUse(event);
     }
   };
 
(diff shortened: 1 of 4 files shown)


src/components/Dropdown/DropdownList.js
```diff
@@ -83,7 +83,7 @@ class DropdownList extends React.PureComponent {
     return this.hoverCallbacks[itemKey];
   };
 
-  handleSelectKeyUse = () => {
+  handleSelectKeyUse = event => {
```
dprzybylek: maybe we should keep convention and decide to use `e` or `event`?

kamilmateusiak: yup, good idea. Let's use `event` :)

kamilmateusiak: changed in https://github.com/livechat/design-system/pull/80/commits/056f8e8d2dba917e061ded9e4c6bdabc2db8bc0c

PR #85, opened 2019-06-18
Add badge

fedyk:
DS Workshop: add badge support for Design System


src/scss/variables/colors.scss
```diff
@@ -31,3 +31,9 @@ $error-color: #d64646;
 $success-color: #4bb678;
 $warning-color: #efa842;
 $info-color: #5ca8f5;
+$light-color: #cbd4de;
+
```
dprzybylek: we should discuss colors naming convention with @kamilmateusiak ;)

fedyk: Just removed the controversial color from variables, https://github.com/livechat/design-system/pull/85/commits/f883115cc7135279d240b0f31438e25bb244de80

src/components/Badge/Badge.md
```diff
@@ -0,0 +1,56 @@
...
+## Color
+
+- Red reserved to display a sense of urgency or indicate state.
+- Light Variation is best for non-essential element on the screen and should recede from the primary content.
+
+# Primary
+
+Errors and important content, like unread messages, new tickets.
+
+```js
```
dprzybylek: Please add exmaples with HTML code (use`js noeditor`)

src/interfaces/badge.d.ts
```diff
@@ -0,0 +1,9 @@
+// <reference types="react" />
+export type BadgeType = "primary" | "light";
+
+export interface IBadgeProps {
```
dprzybylek: it probably should `extends React.HTMLAttributes<HTMLSpanElement>` @kamilmateusiak wdyt?

src/components/Badge/Badge.js
```diff
@@ -0,0 +1,23 @@
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
+import styles from './style.scss';
+
+function Badge(props) {
+  const badgeType = `badge--${props.type || 'primary'}`;
```
dprzybylek: Other components provides `className` prop, I'm not sure if badge should provide it to. cc: @kamilmateusiak

kamilmateusiak: yup, we are trying to make our component as flexible as possible. Usually, we are using {...restProps}. That way developer will be able to pass e.g `data-test` attribute. Also, we add a possibility to extend/change our styles by merging our className with incoming. You can use `classnames` library for this.

src/components/Badge/Badge.md
kamilmateusiak: there is `ComponentHtmlMarkup` component . which will render the component and extract HTML from it :) We are using this when the component is simple and can be used without React by simply using correct classNames

src/components/Badge/style.scss
```diff
@@ -0,0 +1,31 @@
+@import '../../scss/variables/colors';
+
+.badge {
+  position: relative;
+  display: inline-block;
+  vertical-align: middle;
+  height: 16px;
```
kamilmateusiak: imho instead of height u should use `padding` to control size of the component. Also, padding left and right doesn't great when content is longer http://take.ms/qifur

src/interfaces/badge.d.ts
kamilmateusiak: yup, if we use `...restProps `

Neob91:
Please check https://app.zeplin.io/project/5b4fb0903176d0d90e5b04f8/screen/5d63ad582dba3f9bc1176211 for reference

styleguide.config.js
```diff
@@ -42,6 +42,10 @@ module.exports = {
           content:
             './packages/design-system/src/components/NotificationSystem/NotificationSystem.md'
         },
+        {
+          name: 'Badge',
+          components: './src/components/Badge/Badge.js'
```
kamilmateusiak: the url is wrong and the component is not visible in the docs
it should be: `./packages/design-system/src/components/Badge/Badge.js`

packages/design-system/src/components/Badge/style.scss
```diff
@@ -0,0 +1,30 @@
...
+  text-align: center;
+  border-radius: 18px;
+  overflow: hidden;
+
+  & + & {
+    margin-left: 4px;
+  }
+}
+
+.badge--primary {
```
kamilmateusiak: I think this class is not necessary, those colors could be used in `.badge` class

packages/design-system/src/components/Badge/Badge.js
```diff
@@ -0,0 +1,23 @@
...
+    </span>
+  );
+}
+
+Badge.propTypes = {
+  children: PropTypes.node.isRequired,
+  /**
+   * Type of badge
+   */
+  type: PropTypes.oneOf(['primary', 'light'])
```
kamilmateusiak: I would rather avoid using words like 'dark' or 'light'. We have started to develop themes in the design-system (and soon it will be used in AA) and this named can be problematic. I think the word `secondary` is much better here.

packages/design-system/src/components/Badge/Badge.js
```diff
@@ -0,0 +1,23 @@
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
+import styles from './style.scss';
+
+function Badge(props) {
+  const badgeType = `badge--${props.type || 'primary'}`;
+
+  return (
+    <span className={`${styles.badge} ${styles[badgeType]}`}>
```
kamilmateusiak:
```suggestion
className={cx(styles.badge, {
   [styles[`badge--${props.type`]]: props.type && props.type !== 'primary'
 })}
```

kamilmateusiak: Also we could add a support for custom classname from props

packages/design-system/src/interfaces/badge.d.ts
```diff
@@ -0,0 +1,5 @@
+interface IBadgeProps
+  type?: 'primary' | 'light';
+}
+
+export const Badge: React.FC<IBadgeProps>;
```
kamilmateusiak: not sure about this (I encountered a lot of problems when tried to use `const` and `React.FC` instead of `ComponentType`, could you check if it's working with the Agent App. You can simply use `npm link`

kamilmateusiak: Also, take a look at https://github.com/livechat/design-system/pull/85/files#diff-6d2c4a9bb6e4b0a30e4717337472a95bR2 - previous declaration of types for Badge
Please add `extends React.HTMLAttributes<HTMLSpanElement>` to the IBadgeProps to support span element attributes

packages/design-system/src/components/Badge/Badge.js
dprzybylek: Are you sure that we have only 3 types of badge? Please, ask @Jolama for more colors.
image

packages/design-system/src/interfaces/badge.d.ts
Neob91: @kamilmateusiak Why do we need to support span element attributes?

kamilmateusiak: Quick example: your tester needs to write some automatic tests (if the badge is appearing) and you need to add data-test attribute to the component). How will you accomplish the task? 
It's a common pattern in the lib. We don't want to limit components to only a few props which we guess that will be needed.

kamilmateusiak: `className` is a must here. It would be great to use this badge in `Avatar` component of Agent App. We need to position it correctly (`position: absolute` with a few styles).

packages/design-system/src/components/Badge/Badge.js
Neob91: I do not believe we should permit the usage of a custom class name - this would tempt devs to customise these components on their own, while we don't want them to.

packages/design-system/src/interfaces/badge.d.ts
```diff
@@ -0,0 +1,5 @@
+interface IBadgeProps extends React.HTMLAttributes<HTMLSpanElement> {
+  secondary?: boolean;
+}
+
+export const Badge: React.FC<IBadgeProps>;
```
kamilmateusiak: did you check is the interface works correctly?

src/interfaces/badge.d.ts
```diff
@@ -0,0 +1,9 @@
+// <reference types="react" />
+export type BadgeType = "primary" | "light";
```
kamilmateusiak: duplication to remove

packages/design-system/src/interfaces/badge.d.ts
```diff
@@ -0,0 +1,5 @@
+interface IBadgeProps extends React.HTMLAttributes<HTMLSpanElement> {
```
kamilmateusiak: omit className, use custom `type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;` because Agent App has old ts version

packages/design-system/src/components/Badge/Badge.js
```diff
@@ -0,0 +1,23 @@
...
+import cx from 'classnames';
+import styles from './style.scss';
+
+function Badge({ children, className: extraClassName, secondary, ...props }) {
+  const className = cx(styles.badge, {
+    [styles['badge--secondary']]: secondary
+  });
+
+  return (
+    <span className={className} {...props}>
```
kamilmateusiak: If you don't want to include className in the component api you need to change the order:
```suggestion
    <span {...props} className={className}>
```

packages/design-system/src/components/Badge/Badge.md
```diff
@@ -0,0 +1,56 @@
...
+- Red reserved to display a sense of urgency or indicate state.
+- Light Variation is best for non-essential element on the screen and should recede from the primary content.
+
+# Primary
+
+Errors and important content, like unread messages, new tickets.
+
+```js
+<Badge type="primary">1</Badge>
+<Badge type="primary">11</Badge>
```
kamilmateusiak: not updated docs (the type was changed)

packages/design-system/src/components/Badge/Badge.js
kamilmateusiak: Did you ask the designer about used badges colors?

packages/design-system/src/interfaces/badge.d.ts
Neob91: Why is this necessary?

packages/design-system/src/components/Badge/Badge.js
Neob91: It's not needed with the way it's currently done.

packages/design-system/src/interfaces/badge.d.ts
kamilmateusiak: because a developer might think it supports `className` property, omit it and he/she will get an error when passing className

packages/design-system/src/components/Badge/Badge.js
kamilmateusiak: ok

packages/design-system/src/interfaces/badge.d.ts
Neob91: Yes, it does.

packages/design-system/src/interfaces/badge.d.ts
Neob91: I decided to add `className` support after all.

packages/design-system/src/components/Badge/style.scss
```diff
@@ -0,0 +1,28 @@
...
+  padding: 0 3px;
+  height: 18px;
+  min-width: 18px;
+
+  font-size: 10px;
+  font-weight: 600;
+  line-height: 18px;
+  text-align: center;
+  border-radius: 18px;
+  overflow: hidden;
```
kamilmateusiak: why do we need `overflow: hidden`? when used with text it's not centered and this overflow change position of the badge https://take.ms/peZyq

packages/design-system/src/components/Badge/Badge.js
kamilmateusiak: @Neob91 ping

kamilmateusiak:
it's almost done!

packages/design-system/src/components/Badge/Badge.js
Neob91: Yes, this is something else than badge.

packages/design-system/src/components/Badge/style.scss
Neob91: It was already there when I started work on this PR - I can remove it.

Neob91:
@kamilmateusiak Can I get an approve now? :-)

packages/design-system/src/components/Badge/Badge.js
dprzybylek: @Neob91  in fact we want to give them opportunity to customise components.

PR #87, opened 2019-06-18
ButtonGroup component

walaszczykm:


src/components/ButtonGroup/ButtonGroup.js
```diff
@@ -0,0 +1,74 @@
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
+import styles from './style.scss';
+
+const baseClass = 'btn-group';
+
+class ButtonGroup extends React.Component {
+  state = {
+    currentIndex: this.props.currentIndex || -1
```
dprzybylek: please, move this into `const defaultIndex= -1`

src/components/ButtonGroup/ButtonGroup.js
```diff
@@ -0,0 +1,74 @@
...
+  };
+
+  makeClickHandler = index => event => {
```
dprzybylek: `handleClick`

src/components/ButtonGroup/ButtonGroup.js
```diff
@@ -0,0 +1,74 @@
...
+    this.props.onChange(event, index);
+    this.setState({ currentIndex: index });
+  };
+
+  render() {
+    const { size, fullWidth, children, ...restProps } = this.props;
+    const currentIndex =
+      typeof this.props.currentIndex === 'number'
```
dprzybylek: I think we should force number value in this props

styleguide.config.js
```diff
@@ -243,7 +246,7 @@ module.exports = {
           .view-html-btn--active {
             border-bottom: 2px #f28a25 solid;
           }
-          
+
           pre.prettyprint { display: block; background-color: #0a001f; font-size: 13px; font-family: monospace; margin: 0; padding: 8px 16px;}
```
dprzybylek: extra space?

src/components/ButtonGroup/ButtonGroup.js
```diff
@@ -0,0 +1,74 @@
...
+
+ButtonGroup.defaultProps = {
+  onChange: () => {}
+};
+
+ButtonGroup.propTypes = {
+  /**
+   * If `true`, the buttons in group will take up the full width of its container
+   */
+  fullWidth: PropTypes.bool,
```
dprzybylek: Maybe we should pass `className` instead of `fullWidth`?

styleguide.config.js
walaszczykm: removed additional tab, autoformat 🤷‍♂

src/components/ButtonGroup/ButtonGroup.js
walaszczykm: The component checks for the type of this prop to determine if it was passed and if it contains valid index value to make a decision if the component should use forced index passed through props or use its internal `state.currentIndex`. Of course, I could check if it is not `undefined` but I don't like to use `undefined` keyword explicit 🤔

src/components/ButtonGroup/ButtonGroup.js
walaszczykm: I would like to allow users to use the same props like the ones on `Button` component because after all they are assigned to those buttons rendered inside the group while rendering. But if you have in mind a better way of handling this behaviour  I am open for suggestions 🙂

styleguide.config.js
dprzybylek: 🤷‍♀

src/components/ButtonGroup/ButtonGroup.js
dprzybylek: so, maybe use isNaN? :D

src/components/ButtonGroup/ButtonGroup.js
walaszczykm: done 👍

src/components/ButtonGroup/ButtonGroup.js
walaszczykm: done 👍

src/components/ButtonGroup/ButtonGroup.js
dprzybylek: but w8, you initialize component with state
```
state = {
    currentIndex: this.props.currentIndex || defaultIndex
  };
```

so do you need this check in render?

walaszczykm: that was my first idea but then eslint comes on the way 🤷‍♂ 
```
Unexpected use of 'isNaN'.eslint(no-restricted-globals)
```

walaszczykm: Yep because of updates in props, possibly the `currentIndex` prop could be removed or added after component initialization. Now this component is "always ready to update" 😄. The state initialization using this prop is for properly syncing the starting state value: If I pass for example `currentIndex={2}` and then remove it the component will handle tracking by itself from proper index `2` 🙂

walaszczykm: I have updated the implementation by using the derived state to handle forcing `currentIndex` value from props. That way this variable in the render is not needed and component always relies on its state (but the state could be changed from props) 🙂

src/components/ButtonGroup/ButtonGroup.js
```diff
@@ -0,0 +1,84 @@
...
+    const { size, fullWidth, children, ...restProps } = this.props;
+
+    const mappedChildren = React.Children.map(children, (child, index) =>
+      React.cloneElement(child, {
+        size,
+        fullWidth,
+        primary: false,
+        destructive: false,
+        onClick: this.handleClick(index),
+        className: index === currentIndex ? styles.active : null
```
dprzybylek: @walaszczykm look what happens when the middle button is active.
image

src/components/ButtonGroup/ButtonGroup.js
dprzybylek: I think we should provide `className` and `fullWidth`. As I see both are used in `Button` component.

walaszczykm: ok, I have updated the component to use also `className` 👍

src/components/ButtonGroup/ButtonGroup.js
walaszczykm: I found a solution to this problem and implemented it 👍

walaszczykm:
@dprzybylek I have updated the PR and rebased it. Can you please check again if everything is alright now?

src/components/ButtonGroup/ButtonGroup.js
```diff
@@ -0,0 +1,93 @@
...
+    const { currentIndex } = this.state;
+    const { size, fullWidth, children, className, ...restProps } = this.props;
+
+    const mappedChildren = React.Children.map(children, (child, index) =>
+      React.cloneElement(child, {
+        size,
+        fullWidth,
+        primary: false,
+        destructive: false,
+        onClick: this.handleClick(index),
```
kamilmateusiak: It gonna be confusing if some1 pass the `onClick` or any other property which is changed here, but I understand why you did it like that :) We could add a note in the docs

src/interfaces/buttonGroup.d.ts
```diff
@@ -0,0 +1,13 @@
+// <reference types="react" />
+import { ButtonSize } from "./buttons";
+
+export interface IButtonGroupProps {
```
kamilmateusiak: How about extending `React.HTMLAttributes<HTMLDivElement>` here? You are using `restProps` in the render method of the component so it will do no harm to allow data attributes and so on.

walaszczykm: Good point will do 🙂

src/components/ButtonGroup/ButtonGroup.js
walaszczykm: well, I've added wrapper lambda which calls `this.handleClick` and then original `child.props.onClick` if provided 🙂

src/interfaces/buttonGroup.d.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export interface IButtonGroupProps
+  extends React.HTMLAttributes<HTMLDivElement> {
+  fullWidth?: boolean;
+  className?: string;
+  currentIndex?: number;
+  size: ButtonSize;
+  children: React.ReactNode;
+  onIndexChange?: (
+    event: React.MouseEvent<HTMLDivElement>,
```
kamilmateusiak: should `event` be here? :> You have `this.props.onIndexChange(index)` in the code

kamilmateusiak: IMHO it a good thing to pass `event`. Please change it in the component

walaszczykm: 🙈

src/components/ButtonGroup/ButtonGroup.js
```diff
@@ -0,0 +1,98 @@
...
+    return (
+      <div role="group" className={styles[baseClass]} {...restProps}>
+        {mappedChildren}
+      </div>
+    );
+  }
+}
+
+ButtonGroup.defaultProps = {
+  onIndexChange: () => {}
```
kamilmateusiak: I wonder if `onIndexChange` shouldn't be required. What do you think?

walaszczykm: well actually it is not required to work If you just want to visually identify the current state and assign handlers to `onClick` prop for individual buttons

src/interfaces/buttonGroup.d.ts
walaszczykm: ok, updated 👍

walaszczykm:
@kamilmateusiak I think I have resolved all your suggestions. Please let me know if there is anything more or if I can merge this PR 🙂

kamilmateusiak:
Imho it's ready to be merged. Great work 💪

Did we give you a design-system t-shirt already? You deserve it :)

walaszczykm:
> Imho it's ready to be merged. Great work 💪
> 
> Did we give you a design-system t-shirt already? You deserve it :)

Yeah sure, I have a design-system t-shirt 😄

PR #96, opened 2019-07-16
fix radio button on ie11

dprzybylek:


src/components/RadioButton/style.scss
```diff
@@ -51,7 +52,7 @@ $base-class: 'radio';
     font-size: 15px;
     color: $font-primary-color;
     line-height: 20px;
-    flex: 1;
+    flex: 1 0 auto;
```
kamilmateusiak: can you fix `.lc-checkbox__text` as well? there is the same problem

kamilmateusiak: and `.lc-checkbox__label`

kamilmateusiak:
After the merge, it would be great to test it with beta release. In AA we are using RadioButtons and AFAIR there are some hacks which we need to fix

src/components/RadioButton/style.scss
dprzybylek: @kamilmateusiak done

PR #98, opened 2019-07-16
fixed btn line height

dprzybylek:


src/components/Button/Button.js
```diff
@@ -66,7 +66,7 @@ const Button = React.forwardRef((props, ref) => {
       {loading && (
         <Loader size="small" className={styles[`${baseClass}__loader`]} />
       )}
-      {children}
+      <div className={styles[`${baseClass}__content`]}>{children}</div>
```
kamilmateusiak: woaaah :D that will be a big change 
I saw that many components libraries did it like this and it's acceptable. Can you use `span` instead of `div`? It seems more correct.

src/components/Button/style.scss
```diff
@@ -23,7 +23,13 @@
   padding: 0 16px;
   min-width: 36px;
   min-height: 36px;
-  line-height: 36px;
+  line-height: 1.3;
+
+  &:after{
```
kamilmateusiak: can u explain to me why we need it?

src/components/Button/style.scss
```diff
@@ -129,4 +133,7 @@
     transform: translate(-50%, -50%);
     left: 50%;
   }
+  .btn__content {
+    margin: auto;
```
kamilmateusiak: if you will use `span` please remove `margin: auto` and add `display: inline-block;`

kamilmateusiak: `.btn` has `text-align: center;` so it will be centered

src/components/Button/style.scss
dprzybylek: yup, flex in ie 11. 
https://stackoverflow.com/questions/19371626/flexbox-not-centering-vertically-in-ie

kamilmateusiak: omg :D Didn't other solutions work?

dprzybylek: didn't because we cannot set constant btn height  😢

kamilmateusiak:
please add `box-sizing: border-box` to lc-btn class https://take.ms/Ca7Bg

kamilmateusiak:
it appears that inline-block instead of inline-flex works perfectly fine: https://take.ms/M66N2

src/components/Button/style.scss
```diff
@@ -15,15 +14,15 @@
...
   background-color: #fff;
   color: $font-primary-color;
   border-color: $border-primary-color;
   font-size: 15px;
-  padding: 0 16px;
+  padding: 7px 16px;
   min-width: 36px;
   min-height: 36px;
-  line-height: 36px;
+  line-height: 1.3;
```
quarties: I think we should use `px` here to keep consistency - we use pixels for `line-height` [almost everywhere](https://github.com/livechat/design-system/search?p=3&q=%22line-height%3A%22&unscoped_q=%22line-height%3A%22).

For `1.3` the pixel equivalent would be `19px / 20px`

PR #102, opened 2019-07-26
added missing interfaces for form

dprzybylek:


src/interfaces/forms.d.ts
```diff
@@ -78,7 +93,8 @@ export interface IMultiSelectFieldProps extends IMultiSelectProps {
   description?: React.ReactNode;
 }
 
-export interface IInputProps extends React.InputHTMLAttributes<HTMLInputElement> {
+export interface IInputProps
+  extends React.InputHTMLAttributes<HTMLInputElement> {
```
kamilmateusiak: could you add `ref` property to types definitions of components with `forwardRef`? Button component is a good example how it should be done

src/interfaces/forms.d.ts
```diff
@@ -1,5 +1,20 @@
 // <reference types="react" />
 
+export interface IFormProps extends React.InputHTMLAttributes<HTMLFormElement> {
```
kamilmateusiak: `InputHTMLAttributes` ?

src/interfaces/forms.d.ts
dprzybylek: only `input` uses forwardRef 😱

src/interfaces/forms.d.ts
dprzybylek: changed to `React.HTMLAttributes<HTMLFormElement>`

dprzybylek:
@kamilmateusiak fixes done, please check.

src/interfaces/forms.d.ts
kamilmateusiak: ohh 😮 we need to add forwardRef to those component in one of next PRs

dprzybylek:
@kamilmateusiak removed 🤦‍♀

src/interfaces/forms.d.ts
```diff
@@ -88,9 +105,13 @@ export interface IInputFieldProps extends IInputProps {
   inline?: boolean;
   description?: React.ReactNode;
   fieldClassName?: string;
+  ref?:
+    | React.Ref<HTMLButtonElement>
```
kamilmateusiak: it should be React.Ref<HTMLInputElement> :)

kamilmateusiak: Also you can fix it only in `IInputFieldProps`. Ref is not needed here because of `IInputFieldProps extends IInputProps` :D

PR #105, opened 2019-08-05
handle error in field group

dprzybylek:
Design discussed with jo, changes requested for company details modal.

image


src/components/FieldGroup/FieldGroup.js
```diff
@@ -22,15 +23,17 @@ const FieldGroup = props => {
   return (
     <div {...restProps} className={mergedClassNames}>
       {children}
+      {error && <FieldError>{error}</FieldError>}
```
kamilmateusiak: take a look at TextField component - maybe it's worth to add also the description :)

dprzybylek: added ;)

PR #109, opened 2019-08-16
allow passing refs down to primitive element in TextArea and TextArea…

mattleonowicz:
…Field


src/interfaces/forms.d.ts
```diff
@@ -110,6 +110,7 @@ export interface IInputFieldProps extends IInputProps {
 export interface ITextAreaProps
   extends React.TextareaHTMLAttributes<HTMLTextAreaElement> {
   error?: string;
+  ref?: React.Ref<HTMLTextAreaElement>;
```
kamilmateusiak: add `| React.Ref<React.Component<ITextAreaProps>>` - it's helpful when someone uses `styled` from `react-emotion` library

PR #115, opened 2019-08-29
close modal on mousedown

dprzybylek:


dprzybylek:
nope, try to click on modal (don't let the button up!) and move cursor to the overlay. Leave mouse alone, modal will close.

PR #117, opened 2019-08-30
Guide Tooltips > Make spotlight non-clickable

norbertprocak:


src/components/TooltipNew/SpotlightOverlay.js
```diff
@@ -33,6 +33,14 @@ const SpotlightOverlay = ({ gap, isVisible, slide }) => {
     height: `calc(100% - ${gap.bottom}px)`
   };
 
+  const spotlight = {
+    top: `${gap.top}px`,
+    left: `${gap.left}px`,
+    width: `${gap.width}px`,
+    height: `${gap.height}px`,
+    opacity: 0
```
kamilmateusiak: since `visibility` is still `visible` it will take over all click event right?

src/interfaces/tooltipNew.d.ts
```diff
@@ -76,6 +76,7 @@ export interface IUserGuideTooltipProps
   theme?: TooltipTheme;
   placement?: PopperJS.Placement;
   containerName?: string;
+  disableSpotlightClicks?: boolean;
```
kamilmateusiak: `disabledSpotlightPointerEvents` maybe
hmm, still it will be possible to do some action with the keyboard. Is this solution good enough for us?

norbertprocak: I changed the prop name.

Yes, I think disabling the mouse events is good enough.

src/components/TooltipNew/SpotlightOverlay.js
norbertprocak: ok

src/components/TooltipNew/UserGuideTooltip.js
```diff
@@ -127,6 +128,7 @@ class UserGuideTooltip extends React.PureComponent {
           gap={rect}
           isVisible={isVisible}
           slide={shouldSlide}
+          disableClicks={disableSpotlightPointerEvents}
```
kamilmateusiak: can you also use `disableSpotlightPointerEvents` property name here?

src/components/TooltipNew/SpotlightOverlay.js
```diff
@@ -81,7 +99,8 @@ SpotlightOverlay.propTypes = {
     height: PropTypes.number.isRequired
   }).isRequired,
   isVisible: PropTypes.bool.isRequired,
-  slide: PropTypes.bool.isRequired
+  slide: PropTypes.bool.isRequired,
+  disableClicks: PropTypes.bool.isRequired
```
dprzybylek: There is a lot of required props, in my opinion we should provide default value here or in higher component. Btw, `disableClicks` is required here, but it isn't required at `UserGuideTooltop`.

kamilmateusiak: ooops, you are right. No need for required and default props for `disableClicks` property

src/components/TooltipNew/UserGuideTooltip.js
norbertprocak: ok

src/components/TooltipNew/SpotlightOverlay.js
norbertprocak: ok

PR #119, opened 2019-09-02
Guide Tooltips > Overlay z-index

norbertprocak:


norbertprocak:
> Why do you want to set `zIndex` via prop instead of using `style` and passing custom JSS?

There is a legacy problem with z-index because it's required by PopperTooltip for some reason. So to properly provide the z-index to the tooltip I must use it as a prop across the whole UserGuideTooltip component and its descendants.

src/components/Modal/ModalPortal.js
```diff
@@ -36,6 +37,7 @@ class ModalPortal extends React.Component {
 ModalPortal.propTypes = {
   className: PropTypes.string,
   children: PropTypes.node.isRequired,
+  zIndex: PropTypes.number,
```
kamilmateusiak: it's required in PopperTooltip but it should not be required here

kamilmateusiak: `PopperTooltip` is not the only component which uses ModalPortal

norbertprocak: The `zIndex` prop itself for `ModalPortal` is not required (can be omitted). But if what you mean is that there should be no prop `zIndex` for `ModalPortal` at all, then I need your advice on how to provide z-index to the overlay.
My case is that when I open UserGuideTooltip, the black overlay gets covered by any background element that has z-index (LiveChat navigation menu for example). I need to provide z-index to the overlay somehow.
My case is that when I

kamilmateusiak: I don't know why but I assumed that you are requiring it here. 🤔The way it is now is fine.

src/components/Modal/ModalPortal.js
```diff
@@ -11,9 +11,10 @@ class ModalPortal extends React.Component {
     super(props);
 
     this.modalContainer = document.createElement('div');
-    if (this.props.className) {
-      this.modalContainer.className = this.props.className;
+    if (props.className) {
+      this.modalContainer.className = props.className;
     }
+    this.modalContainer.style.zIndex = props.zIndex;
```
kamilmateusiak: please add check:
```suggestion
  if (props.zIndex) {
    this.modalContainer.style.zIndex = props.zIndex;
  }
```

norbertprocak: I provided the extra check

src/components/Modal/ModalPortal.js
```diff
@@ -11,8 +11,11 @@ class ModalPortal extends React.Component {
     super(props);
 
     this.modalContainer = document.createElement('div');
-    if (this.props.className) {
-      this.modalContainer.className = this.props.className;
+    if (props.className) {
+      this.modalContainer.className = props.className;
+    }
+    if (props.zIndex != null) {
```
kamilmateusiak: shouldn't you check `null` or `undefined`?

norbertprocak: if I used triple equality (`===`/`!==`) then I would have to check for both `null` and `undefined`, but since `null == undefined` is `true` then only null-check is required

I can change that to strict equality if you prefer

kamilmateusiak: x rozne null
yup, please use `!==`

norbertprocak: wait, `x != null` is `true` which is fine (the z-index will be assigned '0')

kamilmateusiak: omg ;D sorry, you are right

PR #124, opened 2019-09-13
Tooltips > Close with X button

norbertprocak:


src/components/TooltipNew/style.scss
```diff
@@ -15,16 +15,38 @@
...
+    }
+  }
+
   &--invert {
     background-color: #fff;
     border-color: #fff;
     color: $font-primary-color;
+
+    .popper-tooltip__close svg {
+      fill: $font-primary-color;
```
kamilmateusiak: you can use `fill: currentColor;` - `svg` would inherit color from the parent :)

src/components/TooltipNew/style.scss
```diff
@@ -15,16 +15,38 @@
...
+    }
   }
 
   &--important {
     background-color: $ui-tooltip-important-color;
     border-color: $ui-tooltip-important-color;
     color: $font-primary-color;
+
```
kamilmateusiak: can u add not required `onClose` property to `CssTooltip` and use it to show close button there?

src/components/TooltipNew/style.scss
norbertprocak: awesome, thanks!

src/components/TooltipNew/style.scss
norbertprocak: sure

kamilmateusiak:
https://github.com/livechat/design-system/blob/e43e2cb3313bc1e61deecb44f40bcf93bbe611c1/src/interfaces/tooltipNew.d.ts#L6 
Since you are adding changes in tooltips can you remove this 'normal' from ts types for Tooltips? Some devs were confused because of it.

src/components/TooltipNew/style.scss
```diff
@@ -15,6 +15,20 @@
   text-align: left;
   transition-property: opacity, visibility;
 
+  &__close {
```
kamilmateusiak: you are repeating the same code two times. Can you store it in the mixin?
```suggestion
@mixin close-button {
  &__close {
    display: inline-block;
    position: absolute;
    right: 0;
    top: 0;
    padding: 8px;
    cursor: pointer;
    line-height: 0;
    
    svg {
      fill: currentColor;
    }
  }
}
```
or you can add the following line at the bottom of the file:
```
.popper-tooltip, css-tooltip {
  &__close {
    ...
  }
}
```

src/components/TooltipNew/PopperTooltip.js
```diff
@@ -347,6 +361,7 @@ PopperTooltip.propTypes = {
    * If you are using `triggerActionType='managed'` event handler will call provided onClose prop.
    */
   closeOnOutsideClick: PropTypes.bool,
+  closeWithX: PropTypes.bool,
```
kamilmateusiak: I not going to complain about it this time, but it's important to add comments to props. If you add a comment to property it will be rendered in the docs.

src/components/TooltipNew/style.scss
norbertprocak: done

src/components/TooltipNew/PopperTooltip.js
norbertprocak: no problem I added some docs

also I added another way to close tooltip - with esc key (request by jo)

norbertprocak:
> https://github.com/livechat/design-system/blob/e43e2cb3313bc1e61deecb44f40bcf93bbe611c1/src/interfaces/tooltipNew.d.ts#L6
> 
> 
> Since you are adding changes in tooltips can you remove this 'normal' from ts types for Tooltips? Some devs were confused because of it.

removed

src/components/TooltipNew/PopperTooltip.js
```diff
@@ -33,8 +37,13 @@ class PopperTooltip extends React.PureComponent {
       document.addEventListener('click', this.handleDocumentClick);
     }
 
+    if (this.props.closeWithEsc && this.getIsVisible()) {
+      document.addEventListener('keydown', this.handleDocumentKeyDown);
+    }
+
     if (didHide) {
       document.removeEventListener('click', this.handleDocumentClick);
+      document.removeEventListener('keydown', this.handleDocumentKeyDown);
```
kamilmateusiak: remove event listener also in `componentWillUnmount`

norbertprocak: right

kamilmateusiak:
current: https://take.ms/GlyGD 
maybe? (padding: 4px): https://take.ms/T7Zjf
Shouldn't `x` be moved closer to the edge?

src/components/TooltipNew/PopperTooltip.js
```diff
@@ -274,6 +296,14 @@ class PopperTooltip extends React.PureComponent {
           [className]: className
         })}
       >
+        {closeWithX && (
+          <div
```
kamilmateusiak: one last thing: can we use `button` instead of div? I would like to focus on this close button with `tab` key

norbertprocak:
> current: https://take.ms/GlyGD
> maybe? (padding: 4px): https://take.ms/T7Zjf
> Shouldn't `x` be moved closer to the edge?

Yes, that was also my concern. Currently you can create a tooltip like this:
Zrzut ekranu 2019-09-18 o 12 32 37

The problem is with padding-right. We can move X closer to the corner but it will still be too close to the text. We can add extra padding-right to the tooltip but that wouldn't look good on some tooltips.

For example this is fine (X is closer to corner and extra padding-right is provided):
Zrzut ekranu 2019-09-18 o 12 37 03
But this is not (same tooltip, longer text):
Zrzut ekranu 2019-09-18 o 12 37 19


So I guess it depends on the tooltip content so I left the problem to the developer using the tooltip.

src/components/TooltipNew/PopperTooltip.js
norbertprocak: I changed it to button

norbertprocak: And I provided padding `4px`

src/components/TooltipNew/style.scss
```diff
@@ -331,6 +331,26 @@
...
+    position: absolute;
+    right: 0;
+    top: 0;
+    padding: 4px;
+    cursor: pointer;
+    line-height: 0;
+    background: none;
+    border: none;
+    color: inherit;
+    outline: none;
```
kamilmateusiak:
```suggestion
  &:focus {
    box-shadow: 0 0 1px 2px rgba($primary-color, 0.7);
  }

  &:active {
    box-shadow: 0 0 0 1px $border-primary-color;
    transition-duration: 100ms;
  }
```

norbertprocak: done

PR #125, opened 2019-09-18
Colours proof of concept

dprzybylek:


src/themes/themes/default.js
```diff
@@ -0,0 +1,7 @@
+import { grey, grey70, white, grey10 } from '../colours/colours';
```
kamilmateusiak: how will you export it from the lib?

kamilmateusiak: will I be able to import it as follows:
```
import { DefaultTheme } from '@livechat/design-system';

const { colors } = DefaultTheme;
const color = colors.text01;
```
or maybe you are planning something else? ;D

PR #133, opened 2019-10-04
Switch ts types fix

kamilmateusiak:
Switch extended wrong interface `HTMLAttributes` instead of `InputHTMLAttributes`. It was not possible to use ie. `disabled` attribute in ts projects.


src/interfaces/switch.d.ts
```diff
@@ -1,5 +1,8 @@
 export type SwitchSize = "basic" | "compact";
-export interface ISwitchProps extends React.HTMLAttributes<HTMLInputElement> {
+
+type HTMLInputProps = Omit<React.InputHTMLAttributes<HTMLInputElement>, "size">;
```
kamilmateusiak: need to omit size because by default it has type number

PR #134, opened 2019-10-08
Fixing chrome 47 issue - upload bar

kamilmateusiak:
upload-bar-SC
https://livechatinc.atlassian.net/browse/DS-103

available at https://my.labs.livechatinc.com/feature/DS-103


src/components/UploadBar/style.scss
```diff
@@ -105,6 +105,8 @@ $file-upload-progress: "file-upload-progress";
       padding-top: 16px;
       width: 100%;
       max-height: inherit;
+      display: flex;
```
dprzybylek: you can add `flex: 0 1 auto` just in case

PR #140, opened 2019-10-10
Release v0.5.5

kamilmateusiak:
## v0.5.5

_Release: 2019-10-10_

##### Bug Fixes

- UploadBar - fixing scroll issue on old browsers (#134)

---

### DIFF
--- a/src/components/UploadBar/style.scss
+++ b/src/components/UploadBar/style.scss
@@ -105,6 +105,8 @@ $file-upload-progress: "file-upload-progress";
       padding-top: 16px;
       width: 100%;
       max-height: inherit;
+      display: flex;
+      flex-direction: column;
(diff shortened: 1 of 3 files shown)


src/components/UploadBar/style.scss
```diff
@@ -105,6 +105,8 @@ $file-upload-progress: "file-upload-progress";
       padding-top: 16px;
       width: 100%;
       max-height: inherit;
+      display: flex;
+      flex-direction: column;
```
dprzybylek: `flex: 0 1 auto;` ?

kamilmateusiak: it was tested and seems that there is no need for it
I'm totally against adding not needed code

PR #142, opened 2019-10-14
Export TextField component

mattleonowicz:


src/components/index.js
```diff
@@ -27,6 +27,7 @@ export {
   ModalBody
 } from './Modal';
 
+export { default as TextField } from './TextField';
```
kamilmateusiak: You need to add some ts typings in `forms.d.ts`.  AFAIK you are using typescript in the HelpDesk so it could be pretty useful :)

mattleonowicz: Right! Done.

mattleonowicz:
Ok, done.

PR #144, opened 2019-10-15
Lerna

dprzybylek:
## v0.6.0

_Release: 2019-10-24_

##### Refactors

- Introducing `Lerna` to LiveChat design system repository.
  The `livechat/design-system` repository was converted to monorepo. All new packages related to our design system ecosystem will be added here. This approach gives as a couple of advantages:
  - more maintainable packages structure
  - all dependencies in a single folder, simple releases with Lerna
  - faster development across dependencies
  - consistent linting and configuration across all packages
  - reusable code (ie. common `helpers` can be moved to standalone package)

---


package.json
```diff
@@ -1,164 +1,18 @@
...
-    "styleguide:build": "cross-env NODE_ENV=production styleguidist build",
-    "predeploy": "npm run-script styleguide:build",
-    "deploy": "gh-pages -d styleguide",
-    "test": "jest",
-    "test:update": "jest -u",
-    "test:watch": "jest --watchAll",
-    "test:coverage": "jest --coverage",
-    "eslint": "eslint --ext .js,.ts src/ ",
-    "eslint:fix": "eslint --ext .js,.ts src/ --fix "
+    "test": "lerna run test"
```
kamilmateusiak: unfortunately, the script is not working for me ;x

kamilmateusiak: do you think we should add something else here?
wouldn't it be worth to add `"docs": "lerna run styleguide"`?

packages/colors/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "@livechat/design-system-colors",
+  "version": "0.0.1",
+  "description": "Design system colors",
+  "author": "Dobrochna <d.przybylek@livechatinc.com>",
+  "homepage": "https://github.com/livechat/design-system",
```
kamilmateusiak: "https://github.com/livechat/design-system/tree/master/packages/colors"

kamilmateusiak:
`packages/design-system/CONTRIBUTING.md` - shouldn't it be moved to root? also the content needs to be changed

kamilmateusiak:
what about docs? why is it still in the packages/design-system directory? all packages should be documented in one documentation IMHO

package.json
dprzybylek: > unfortunately, the script is not working for me ;x

`npm run test` in root dir doesn't work?

packages/design-system/.lintstagedrc.json
```diff
@@ -0,0 +1,13 @@
+{
+  "linters": {
+    "src/**/*.{js,jsx,tsx,ts,json}": [
+      "prettier --write",
+      "eslint --fix --ext .js,.jsx,.tsx,.ts",
```
sgraczyk: ts / tsx ?

lerna.json
```diff
@@ -0,0 +1,6 @@
+{
```
Andarist: I found yarn to perform **much** better for monorepo installations, with npm lerna has to enter each package and run npm commands in each. yarn has monorepo support builtin so everything happens automatically within a single command. When I've switched to yarn in redux-saga my install times has dropped quite a bit

packages/colors/README.md
```diff
@@ -0,0 +1,11 @@
+# `colors`
```
Andarist: this header usually name full package name which in this case is `@livechat/design-system-colors`

packages/colors/lib/colors.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
+module.exports = colors;
```
Andarist: please don't use commonjs in 2019 😢

packages/design-system/.lintstagedrc.json
```diff
@@ -0,0 +1,13 @@
+{
```
Andarist: this should be moved to the root directory, git hooks should be repository-level and not package-level

packages/design-system/.gitignore
```diff
@@ -0,0 +1,26 @@
+# Logs
```
Andarist: similarly - it's not worth to keep .gitignore in each package, u can configure it "globally" in the root - unless there are some package-level stuff you want to ignore, but in here most of the stuff is generic (like logs, dist, node_modules etc)

.lintstagedrc.json
```diff
@@ -1,13 +1,10 @@
...
-      "eslint --fix --ext .js,.jsx,.tsx,.ts",
-      "git add"
-    ],
-    "*.md": [
-      "prettier --write",
-      "git add"
-    ]
-  }
-}
+  "extends": [
```
Andarist: are u sure this is correct? seems like config for stylelint, and not for linstaged

packages/design-system/package.json
```diff
@@ -0,0 +1,159 @@
...
+    "build:cdn": "npm run build && mkdir -p dist/cdn && cp dist/design-system.css dist/cdn/styles.css && cp dist/cdn/styles.css dist/cdn/styles-v$npm_package_version.css",
+    "prepare": "npm run build",
+    "test": "jest",
+    "test:update": "jest -u",
+    "test:watch": "jest --watchAll",
+    "test:coverage": "jest --coverage",
+    "eslint": "eslint --ext .js,.ts src/ ",
+    "eslint:fix": "eslint --ext .js,.ts src/ --fix "
+  },
+  "husky": {
```
Andarist: as mentioned elsewhere - this config should be moved to the root directory

packages/colors/lib/colors.js
dprzybylek: `lerna create` did it 🙈this package is empty, because we will implement colors after switching to lerna.

lerna.json
dprzybylek: Good point, I think we could migrate to yarn in another PR, but I need to discuss it with the DS team.

kamilmateusiak: Yup, we can give it a try since our build and installation time is large. I don't see any cons of using yarn here.

packages/colors/package.json
dprzybylek: done

lerna.json
```diff
@@ -0,0 +1,6 @@
...
+  "packages": [
+    "packages/*"
```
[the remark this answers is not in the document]
dprzybylek: done

packages/colors/README.md
dprzybylek: renamed

packages/design-system/.lintstagedrc.json
dprzybylek: moved

packages/design-system/.gitignore
dprzybylek: moved

.lintstagedrc.json
dprzybylek: please take a look at new config

packages/design-system/package.json
dprzybylek: moved

dprzybylek:
@sgraczyk @kamilmateusiak @Andarist fixes done, please review :)

docs/customized-components/Button.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Button as DSButton } from '../../src/index';
+import { Button as DSButton } from '../../packages/design-system/src/index'
```
Andarist: IMHO you shouldn't reach into packages likes this, monorepo is somewhat an implementation detail and you should access particular packages only through their respective names (in this example - `@livechat/design-system`).

You have two basic approaches that I know of to consume packages with build steps within your monorepo:
1. running watch to rebuild dist files which is really "meh" to me
2. aliasing, this unfortunately brings some extra config into the picture but at least each package can just reach for the newest version of other package and everything can somewhat be contained within a single process (in example webpack). We are using this approach and I've seen other people using it as well, I've also created a small helper package which can help with producing those aliases automatically - https://github.com/Andarist/lerna-alias

.babelrc
```diff
@@ -1,5 +1,5 @@
 {
```
Andarist: what is consuming this babelrc right now? in general babel@6 (which you are still using) had some quirky behavior regarding babelrc resolution algorithm and most likely this can fail once you upgrade to babel@7, so worth considering how to prepare a more bullet-proof solution right now - we can discuss this in person if u want (in general you should avoid hierarchical babelrc files with babel@6)

packages/design-system/package.json
```diff
@@ -0,0 +1,153 @@
...
+    "eslint-import-resolver-webpack": "^0.10.0",
+    "eslint-plugin-babel": "^5.1.0",
+    "eslint-plugin-import": "^2.12.0",
+    "eslint-plugin-jsx-a11y": "^6.0.3",
+    "eslint-plugin-prettier": "^2.6.2",
+    "eslint-plugin-react": "^7.8.2",
+    "eslint-plugin-typescript": "^0.12.0",
+    "extract-text-webpack-plugin": "^3.0.2",
+    "gh-pages": "^1.2.0",
+    "husky": "^1.3.1",
```
Andarist: husky & lint-staged were moved to the root, so they should be removed from here - especially that those version here are now different than what you have in the root

packages/design-system/package.json
```diff
@@ -0,0 +1,153 @@
...
+    "normalize.css": "^8.0.0",
+    "postcss-flexbugs-fixes": "^4.0.0",
+    "postcss-global-import": "^1.0.6",
+    "postcss-import": "^11.1.0",
+    "postcss-load-config": "^2.0.0",
+    "postcss-loader": "^2.1.6",
+    "postcss-modules": "^1.3.2",
+    "postcss-url": "^7.3.2",
+    "precss": "^3.1.2",
+    "prettier": "1.14.2",
```
Andarist: similarly prettier is a monorepo-level tool, so it shouldnt be kept here

packages/design-system/package.json
```diff
@@ -0,0 +1,153 @@
...
+    "babel-preset-es2015-rollup": "^3.0.0",
+    "babel-preset-react": "^6.3.13",
+    "caniuse-lite": "1.0.30000974",
+    "cross-env": "^5.2.0",
+    "css-loader": "^2.1.1",
+    "cssnano": "^4.0.5",
+    "enzyme": "^3.3.0",
+    "enzyme-adapter-react-16": "^1.1.1",
+    "enzyme-to-json": "^3.3.4",
+    "eslint": "^4.19.1",
```
Andarist: same story for the eslint (move it to the root)

packages/design-system/package.json
```diff
@@ -0,0 +1,153 @@
...
+    "sass-loader": "^7.1.0",
+    "style-loader": "^0.23.1",
+    "stylelint": "^9.6.0",
+    "stylelint-config-recommended-scss": "^3.2.0",
+    "stylelint-config-standard": "^18.2.0",
+    "stylelint-scss": "^3.3.1",
+    "ts-loader": "^3.5.0",
+    "typescript": "^2.9.1",
+    "typescript-eslint-parser": "^19.0.2",
+    "webpack": "^4.28.4",
```
Andarist: this package right now doesnt depend anyhow on webpack, so it should be removed from here (along with any other webpack-related devDeps)

docs/customized-components/Button.js
dprzybylek: done at 3638c3e1c30f75a3930047a8ae5c620a3bb40423

packages/design-system/package.json
dprzybylek: done at 860e6b7d0907e3e55330982f896d15ba70f95cc5

packages/design-system/package.json
dprzybylek: done at 860e6b7d0907e3e55330982f896d15ba70f95cc5

packages/design-system/package.json
dprzybylek: done at 13352591223f2ba618e2916566e485b92d3e402f

packages/design-system/package.json
dprzybylek: done at 538a33315631085f7d995318f66c3a90ddf9c327

dprzybylek: done at 538a33315631085f7d995318f66c3a90ddf9c327

.babelrc
dprzybylek: I discussed with @kamilmateusiak - it consumes rollup for now, we will consider upgrade to babel@7 after this PR

kamilmateusiak:
published as beta release to check the build in a real-life app https://www.npmjs.com/package/@livechat/design-system/v/0.6.0-beta.0

PR #147, opened 2019-10-30
fix for CDN styles build after lerna implementation

quarties:


packages/design-system/package.json
```diff
@@ -11,19 +11,19 @@
     "url": "https://github.com/livechat/design-system"
   },
   "typings": "src/interfaces/index.d.ts",
-  "main": "dist/design-system.cjs.js",
-  "module": "dist/design-system.es.js",
-  "jsnext:main": "dist/design-system.es.js",
-  "style": "dist/design-system.css",
+  "main": "../../dist/core/design-system.cjs.js",
```
kamilmateusiak: woaaah, are you sure? shouldn't we change the path on bamboo CDN? `design-system.cjs.js` should stay in the package directory imho

quarties: Well, with monorepo like this one created with `lerna` there should be single folder for artifacts.

kamilmateusiak: but this `main` property is package only, monorepo thing doesn't matter here

kamilmateusiak: https://take.ms/Z0ynv - take a look at this. This is a published design-system library in AA. If you change main etc. it will break the package probably

packages/design-system/package.json
```diff
@@ -11,19 +11,19 @@
...
-    "dist",
+    "../../dist/core",
     "src"
   ],
   "scripts": {
     "start": "rollup -c --watch",
     "build": "rollup -c",
     "prebuild": "rimraf dist",
-    "build:cdn": "npm run build && mkdir -p dist/cdn && cp dist/design-system.css dist/cdn/styles.css && cp dist/cdn/styles.css dist/cdn/styles-v$npm_package_version.css",
+    "build:cdn": "npm run build && mkdir -p ../../dist/cdn && cp ../../dist/core/design-system.css ../../dist/cdn/styles.css && cp ../../dist/cdn/styles.css ../../dist/cdn/styles-v$npm_package_version.css",
```
kamilmateusiak:
```suggestion
    "build:cdn": "npm run build && mkdir -p ../../dist/cdn && cp dist/design-system.css ../../dist/cdn/styles.css && cp ../../dist/cdn/styles.css ../../dist/cdn/styles-v$npm_package_version.css",
```
1st needed change

package.json
```diff
@@ -2,6 +2,8 @@
   "name": "design-system",
   "private": true,
   "scripts": {
+    "postinstall": "lerna bootstrap",
+    "build:cdn": "lerna run --scope @livechat/design-system build:cdn",
```
kamilmateusiak: "postinstall" and  "build:cdn" - second needed change

quarties: what's wrong here?

packages/design-system/package.json
quarties: done in a0846f4

packages/design-system/package.json
quarties: good catch - fixed in a0846f4

package.json
kamilmateusiak: nothing, I meant that this change is needed xD

PR #150, opened 2019-11-13
Controlled button styles

dprzybylek:
new controlled (and group) button styles:
group button

### DIFF
--- a/packages/design-system/src/components/ButtonGroup/ButtonGroup.js
+++ b/packages/design-system/src/components/ButtonGroup/ButtonGroup.js
@@ -36,6 +36,7 @@ class ButtonGroup extends React.Component {
         primary: false,
         secondary: true,
         destructive: false,
+        type: 'button',
         onClick: event => {
           this.handleClick(i, event);
           if (child.props.onClick) {
--- a/packages/design-system/src/components/ButtonGroup/ButtonGroup.js
+++ b/packages/design-system/src/components/ButtonGroup/ButtonGroup.js
@@ -8,26 +8,26 @@ const baseClass = 'btn-group';
 
 class ButtonGroup extends React.Component {
   state = {
-    currentIndex: defaultIndex
+    currentindex: defaultIndex
   };
 
   static getDerivedStateFromProps(props, state) {
     return {
-      currentIndex:
-        typeof props.currentIndex === 'number'
-          ? props.currentIndex
-          : state.currentIndex
+      currentindex:
+        typeof props.currentindex === 'number'
+          ? props.currentindex
+          : state.currentindex
     };
   }
 
   handleClick = (index, event) => {
     this.props.onIndexChange(index, event);
-    this.setState({ currentIndex: index });
+    this.setState({ currentindex: index });
   };
 
   render() {
-    const { currentIndex } = this.state;
-    const { size, fullWidth, children, className, ...restProps } = this.props;
+    const { currentindex } = this.state;
+    const { size, fullWidth, children, className, onIndexChange, ...restProps } = this.props;
 
     const mappedChildren = React.Children.map(children, (child, index) =>
       React.cloneElement(child, {
@@ -44,7 +44,7 @@ class ButtonGroup extends React.Component {
         },
         className: cx({
           [className]: !!className,
-          [styles.active]: index === currentIndex
+          [styles.active]: index === currentindex
         })
       })
     );
@@ -58,7 +58,7 @@ class ButtonGroup extends React.Component {
 }
 
 ButtonGroup.defaultProps = {
-  onIndexChange: () => {}
+  onIndexChange: (index, event) => {}
 };
 
 ButtonGroup.propTypes = {
@@ -73,14 +73,14 @@ ButtonGroup.propTypes = {
   className: PropTypes.string,
 
   /**
-   * Callback fired when the value of `currentIndex` changes
+   * Callback fired when the value of `currentindex` changes
    */
   onIndexChange: PropTypes.func,
 
   /**
    * The index of button in group which is currently active
    */
-  currentIndex: PropTypes.number,
+  currentindex: PropTypes.number,
 
   /**
    * Size of buttons in group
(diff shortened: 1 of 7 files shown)


packages/design-system/src/components/ButtonGroup/ButtonGroup.js
```diff
@@ -8,26 +8,26 @@ const baseClass = 'btn-group';
 
 class ButtonGroup extends React.Component {
   state = {
-    currentIndex: defaultIndex
+    currentindex: defaultIndex
```
marcinsawicki: Why you removed the camelcase format?

packages/design-system/src/components/ButtonGroup/ButtonGroup.md
```diff
@@ -22,22 +22,29 @@ By default ButtonGroup use internal state to track currently selected button. It
...
 
-If you want to use ButtonGroups as controlled component you can set `currentIndex` props to set which button is active.
+If you want to use ButtonGroups as controlled component you can set `currentindex` props to set which button is active.
 
 ```js
-<ButtonGroup currentIndex={1}>
+initialState = { currentindex: 0 };
+
+handleClick = (index, e) => {
+  console.log('index', index)
```
walaszczykm: this `console.log` probably should not be here

packages/design-system/src/components/ButtonGroup/ButtonGroup.js
walaszczykm: Yes I also opt for the original one that I can't find any reasoning for not making it in camel case as they are two separate words 🤔

packages/design-system/src/components/ButtonGroup/ButtonGroup.js
```diff
@@ -58,7 +58,7 @@ class ButtonGroup extends React.Component {
 }
 
 ButtonGroup.defaultProps = {
-  onIndexChange: () => {}
+  onIndexChange: (index, event) => {}
```
walaszczykm: It is actually the same: still `noop` function but now you have there not used arguments which could be bad in case of some linting rule in future.

packages/design-system/src/components/ButtonGroup/ButtonGroup.js
dprzybylek: `const {... currentIndex, ...restProps } = this.props;` this fixed my issue, reverted curentIndex at b1005187a0d51f7e88f2f764c0701d21ca564965

packages/design-system/src/components/ButtonGroup/ButtonGroup.md
dprzybylek: removed at b1005187a0d51f7e88f2f764c0701d21ca564965

packages/design-system/src/components/ButtonGroup/ButtonGroup.js
dprzybylek: deleted this noop, and added check in call b1005187a0d51f7e88f2f764c0701d21ca564965

packages/design-system/src/components/ButtonGroup/ButtonGroup.md
```diff
@@ -25,19 +25,25 @@ By default ButtonGroup use internal state to track currently selected button. It
...
+    borderRadius: '50%',
+    display: 'inline-block',
+    position: 'relative',
+    bottom: '5px'
+  }}></span></Button>
   <Button>Third Button</Button>
 </ButtonGroup>
 ```
-
-```js noeditor
```
kamilmateusiak: why did you remove html example?

packages/design-system/src/components/ButtonGroup/ButtonGroup.md
```diff
@@ -25,19 +25,25 @@ By default ButtonGroup use internal state to track currently selected button. It
...
+handleClick = (index, e) => {
+  setState({
+    currentIndex: index
+  })
+}
+
+<ButtonGroup currentIndex={state.currentIndex} onIndexChange={handleClick}>
   <Button>First Button</Button>
-  <Button>Second Button</Button>
+  <Button>Second Button <span style={{
```
kamilmateusiak: could you declare this span as separate component (here, in doc file). It would be more readable that way
Is it element of design-system?

packages/design-system/src/components/ButtonGroup/ButtonGroup.js
```diff
@@ -8,43 +8,43 @@ const baseClass = 'btn-group';
...
+    const mappedChildren = React.Children.map(children, (child, i) =>
       React.cloneElement(child, {
         size,
         fullWidth,
         primary: false,
         secondary: true,
         destructive: false,
         onClick: event => {
-          this.handleClick(index, event);
+          this.handleClick(i, event);
```
kamilmateusiak: it should have property `type='button'`

packages/design-system/src/components/ButtonGroup/ButtonGroup.js
```diff
@@ -8,43 +8,43 @@ const baseClass = 'btn-group';
...
-    this.props.onIndexChange(index, event);
+    this.props.onIndexChange && this.props.onIndexChange(index, event);
     this.setState({ currentIndex: index });
   };
 
   render() {
-    const { currentIndex } = this.state;
-    const { size, fullWidth, children, className, ...restProps } = this.props;
+    const { index } = this.state;
+    const { size, fullWidth, children, className, onIndexChange, currentIndex, ...restProps } = this.props;
```
walaszczykm: you do not need `currentIndex` here as it is not used in the component `render` method

kamilmateusiak: it's not used, but there is `restProps` from the spread operator. Below div will receive currentIndex as property, but there is no property like this on div element.

dprzybylek: ☝️

packages/design-system/src/components/ButtonGroup/ButtonGroup.md
dprzybylek: no, this was just example for designer, removed

packages/design-system/src/components/ButtonGroup/ButtonGroup.js
kamilmateusiak: what about that?

PR #154, opened 2019-12-04
[AA-7213] Allow autofocusedItemId prop

kamilmateusiak:
Focus can’t be controlled at the moment, but in some cases, it could be helpful - Jakub Sikora, from the Developer Platform team, needs to automatically focus the current version of docs on dropdown open.

Take a look at the comment:
https://livechatinc.atlassian.net/browse/DPS-1258?focusedCommentId=63598&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-63598

Done in PR:
- fixing keydown and hover handlers - items hover handlers were disabled after keyboard keys use when there was no list scroll
- allow autofocusedItemId prop

### DIFF
--- a/packages/design-system/src/components/Dropdown/DropdownList.js
+++ b/packages/design-system/src/components/Dropdown/DropdownList.js
@@ -14,7 +14,8 @@ class DropdownList extends React.PureComponent {
     super(props);
 
     this.state = {
-      focusedElement: getFirstFocusableItemId(props.items),
+      focusedElement:
+        this.props.autoFocusedItemId || getFirstFocusableItemId(props.items),
       itemsCount: props.items.length
     };
   }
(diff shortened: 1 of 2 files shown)


packages/design-system/src/components/Dropdown/DropdownList.js
```diff
@@ -74,6 +75,7 @@ class DropdownList extends React.PureComponent {
   getFocusedItemCallback = itemKey => {
     if (!this.hoverCallbacks[itemKey]) {
       this.hoverCallbacks[itemKey] = () => {
+        debugger; // eslint-disable-line
```
mreszke: please remove

packages/design-system/src/components/Dropdown/DropdownList.js
```diff
@@ -177,10 +177,24 @@ class DropdownList extends React.PureComponent {
           itemOfsetTop - (itemOfsetTop % itemHeigth);
       }
     }
+
+    this.enableHoverOnItems(150);
   };
 
+  enableHoverOnItems(delayInMs) {
+    if (this.hoverEnablerTimeout) clearTimeout(this.hoverEnablerTimeout);
```
mreszke: I'd like to use handlebars, imo it's much more clear

PR #157, opened 2019-12-11
@livechat/design-system-colors setup

kamilmateusiak:
docs available at: https://livechat-design-system-dev.netlify.com/#/Colors

mostly package-locks, do not be scared :)

package json — design-system 2020-01-10 00-48-29


kamilmateusiak:
I added package-lock to design-system package because the build on netlify was failing without it (common problem with CI tools)
I don't really see a reason to not have the package-lock file.

docs/colors/helpers.js
```diff
@@ -0,0 +1,62 @@
...
+  // Set non-editable to avoid focus and move outside of view
+  el.setAttribute('readonly', '');
+  el.style = { position: 'absolute', left: '-9999px' };
+  document.body.appendChild(el);
+  // Select text inside element
+  el.select();
+  // Copy text to clipboard
+  document.execCommand('copy');
+  // Remove temporary element
+  document.body.removeChild(el);
```
dprzybylek: If we want to support iOS there is some magic code for that
```
export const copyContentToClipboardOnIOs = (element: HTMLInputElement): void => {
  const maxRange = element.value.length;
  const range = document.createRange();
  /* eslint-disable no-param-reassign */
  element.contentEditable = 'true';
  element.readOnly = false;
  range.selectNodeContents(element);
  const selection = window.getSelection();
  selection.removeAllRanges();
  selection.addRange(range);
  element.setSelectionRange(0, maxRange);
  element.contentEditable = 'false';
  element.readOnly = true;
  /* eslint-enable no-param-reassign */
  document.execCommand('copy');
};
```

docs/colors/helpers.js
```diff
@@ -0,0 +1,62 @@
...
+}
+
+function getLuminanace(r, g, b) {
+  const a = [r, g, b].map(value => {
+    const divided = value / 255;
+    return divided <= 0.03928
+      ? divided / 12.92
+      : ((divided + 0.055) / 1.055) ** 2.4;
+  });
+  return a[0] * 0.2126 + a[1] * 0.7152 + a[2] * 0.0722;
```
dprzybylek: Ok, so I think all this crazy numbers should go to descriptive consts.

kamilmateusiak: omg, it will require some research xD

dprzybylek: 😂 you can also add a comment "based on ..."

walaszczykm:
Did you consider to use a `polished` library (https://github.com/styled-components/polished) as a toolset for colour operations? The functions defined in `helpers.js` will not be required then. The maintainability of the codebase and its readability will be much better then.

What do you think about that? It is a really solid library with a lot of well splited and named utility functions for manipulating colors and styles. Maybe it is worth to give it a try 🙂

kamilmateusiak:
@walaszczykm Good idea, I will research this package and check it out in package.
I am really concerned with @livechat/design-system-colors package size. Without using `polished` it really small (I used https://www.npmjs.com/package/package-size to measure it):
```
package                  size         minified  gzipped
./packages/colors  5.03 KB  2.02 KB   815 B
```
I need to check the size after using `polished. It's tree-shakeable so the package size shouldn't be much higher.

docs/colors/helpers.js
kamilmateusiak: I used one method from `polished` and it is not needed anymore ;D

kamilmateusiak:
@walaszczykm thank you, `getContrast` did the job :) I removed previous logic

walaszczykm:
@kamilmateusiak I am happy that my suggestion turns out to be helpful 🙂

packages/colors/tools/create-css.ts
```diff
@@ -0,0 +1,56 @@
...
+const colorTextStyles = Object.keys(colors).map(colorName => {
+  const className = `.lcds-text-${colorName
+    .replace(/([a-z])([A-Z])/g, '$1-$2')
+    .toLowerCase()}`;
+  const cssProperties = `{ color: ${(colors as any)[colorName]}; }`;
+
+  return `${className} ${cssProperties}`;
+});
+
+const colorBgStyles = Object.keys(colors).map(colorName => {
```
mjurkowski: A lot of repeats - make a factory :)

packages/colors/README.md
```diff
@@ -1,11 +1,49 @@
...
 
 ## Usage
 
+There are three ways to use LiveChat design colors palette:
+- use JavaScript implementation of colors
 ```
-const colors = require('colors');
+import Colors from '@livechat/design-system-colors'
+```
+- import `design-system-colos.css` file and style UI elements with classes
```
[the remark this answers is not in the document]
kamilmateusiak: I decided to declare vars on `:root` instead of a special class. Although I don't find useful using additional, special class, I remember we talked about 'cascade' property of CSS styles. Could you please remind me why it was important in this case?

fedyk:
`@livechat/design-system-colors` would be a public package?

kamilmateusiak:
@fedyk Yes, it will be published as a public package. It's a very good idea to use it in mobile apps cause it doesn't depend on any HTML/DOM properties - just pure JS/TS with additional styles files  :)

packages/colors/src/design-system-colors.test.ts
```diff
@@ -0,0 +1,73 @@
+import { tint, shade } from 'polished';
+import { pipe } from 'ramda';
+import Colors, { baseColors } from './design-system-colors';
```
kamilmateusiak: with libs
Decided to use scripts from Michał Jawinski only in tests because adding polished and ramda increased size of package 6 times (when gzipped)

PR #158, opened 2019-12-13
added warn about deprecated tooltip

dprzybylek:


packages/design-system/src/components/Tooltip/Tooltip.js
```diff
@@ -9,6 +9,7 @@ import { DIRECTION, ALIGMENT } from '../../constants/tooltip';
 
 class Tooltip extends React.Component {
   constructor(props) {
+    console.warn('[Design System] deprecated component - Tooltip, please use CSSTooltip or PopperTooltip')
```
jdrzejb: Can we ensure that it wouldn't land in the production build?

kamilmateusiak: Yup, please add check for `process.env`

kamilmateusiak: move the console.warn and check to helper - `noticeAboutDeprecation(message: string)`

dprzybylek: done

packages/design-system/src/helpers/notice-about-deprecation.js
```diff
@@ -0,0 +1,6 @@
+
+export const noticeAboutDeprecation = (message) => {
+    const env = process.env.NODE_ENV;
+    const isDevelopmetEnv = env === 'dev' || env === 'development'  || env === 'labs';
+    isDevelopmetEnv && console.error(message)
+}
```
walaszczykm: No EOL ☝️

walaszczykm: I would also suggest using `console.warn` instead of `console.error` for deprecation warnings.

packages/design-system/src/components/Tooltip/TooltipContent.js
```diff
@@ -23,6 +25,8 @@ const TooltipContent = ({
     }),
     className
   );
+  
+  noticeAboutDeprecation('[Design System] deprecated component - Tooltip, please use CSSTooltip or PopperTooltip instead');
```
walaszczykm: maybe this `[Design System]` prefix could be added by the `noticeAboutDeprecation` function itself? So it will be guaranteed that all depreciation warnings will have it.

kamilmateusiak: yup, a very good idea! :)

packages/design-system/src/helpers/notice-about-deprecation.js
kamilmateusiak: here some discussion about deprecation notices in React: https://github.com/facebook/react/issues/9395

walaszczykm: Oh, it' a good one for sure. So by using errors, there are benefits even if it is a little bit less semantically correct 🤔 Good to know 👍

packages/design-system/src/components/Tooltip/TooltipContent.js
dprzybylek: done ;)

dprzybylek: done 🌮

PR #159, opened 2019-12-17
Buttons Wrapper Component

jdrzejb:
Buttons wrapper component.

It helps when you need to inline some actions. It automatically adds spacing to direct children elements.

### DIFF
--- a/packages/design-system/src/components/ButtonsWrapper/style.css
+++ b/packages/design-system/src/components/ButtonsWrapper/style.css
@@ -1,0 +1,10 @@
+.btns-wrapper {
+  display: flex;
+  flex-wrap: wrap;
+  align-items: center;
+  margin-top: -8px;
+}
+.btns-wrapper > * {
+  margin-right: 8px;
+  margin-top: 8px;
+}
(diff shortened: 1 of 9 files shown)


packages/design-system/src/components/ButtonsWrapper/style.css
```diff
@@ -0,0 +1,10 @@
+.btns-wrapper {
```
kamilmateusiak: only `scss` file is fine

packages/design-system/src/components/ButtonsWrapper/style.scss
```diff
@@ -0,0 +1,11 @@
+.btns-wrapper {
+  display: flex;
+  flex-wrap: wrap;
+  align-items: center;
+  margin-top: -8px;
+
+  & > * {
+    margin-right: 8px;
```
kamilmateusiak: is it fine to have always 8px margin on the right side? I know it's not much :)

jdrzejb: It should be zero after last element. I'll fix it

PR #164, opened 2020-02-03
added type number as default value, fix input type

dprzybylek:


packages/design-system/src/components/NumericInputField/NumericInput.js
```diff
@@ -146,12 +146,14 @@ class NumericInput extends React.PureComponent {
       className
     );
 
+    const inputValue = typeof value === 'string' ? parseInt(value, 10) : value;
```
walaszczykm: It is still a little bit weak guard because value with type `string` can be not a number string. In that case a `parseInt` function will return `NaN` value. If you then pass a `NaN` value to a input with type number the React will give a warning:
```
Warning: Received NaN for the `value` attribute. If this is expected, cast the value to a string.
```

dprzybylek: @walaszczykm what do you propose to do? I prefer to remove the string, but don't want to loose backward compatibility :/

walaszczykm: use the fact that for all non-numeric strings the `parseInt` function will return `NaN` and use it for fallback. The only question is what fallback value should be. A `0` or the empty string `''` which will result in the valid empty input 🤔

walaszczykm: @dprzybylek  You also are not handling al other types of values like `object`, `array`, `boolean`, `symbol`. I think all of them should be transformed into a default fallback value (maybe empty string).

Additionally, you can add `useMemo` hook as it is a perfect example for it: generating `inputValue` using provided `value` 🙂.

The implementation of those three suggestions could look like this:
```js
  const inputValue = React.useMemo(() => {
    switch (typeof value) {
      case "number":
        return value;

      case "string": {
        const parsed = parseInt(value, 10);
        return isNaN(parsed) ? "" : value;
      }

      default:
        return "";
    }
  }, [value]);
```

in that way, you can handle specific values types and always fall back to default for all other types. That way you have 100% cover for possible variants 🙂

walaszczykm: But this is only a suggestion. You should think about desired behaviour for wrong values. Maybe relying on default React warning is fine I don't say it is not but it is good to be aware off this.

dprzybylek: please look at 04f9d89997f4e85af874c527bcfee0f1670183d1

packages/design-system/src/components/NumericInputField/NumericInput.js
```diff
@@ -122,6 +122,14 @@ class NumericInput extends React.PureComponent {
     this.removeKeyboardEventListeners();
   };
 
+  parseValue = () => {
+    const { value } = this.props;
+    if(isNaN(value)) {
```
opif: strictly this isn't correct, a string is never a NaN (a really bad name for the concept of "not a Real number"), though this will look like it works since value will be coerced into a number before being tested.
@walaszczykm 's suggestion didn't have this issue, and you should probably use Number.isNaN instead of the global isNaN to avoid the coercion issue I mentioned above.

dprzybylek:
@opif @walaszczykm  I've talk with @kamilmateusiak. String needs to stay, I reverted number type and fix min max handling.

packages/design-system/src/components/NumericInputField/NumericInput.js
```diff
@@ -146,25 +146,28 @@ class NumericInput extends React.PureComponent {
...
         {!noControls && (
           <React.Fragment>
             <button
               tabIndex="-1"
               disabled={
                 disabled ||
-                (this.props.max &&
-                  parseInt(this.props.value, 10) === this.props.max)
+                (max &&
+                  parseInt(value, 10) === max)
```
opif: I can see that the logic didn't change here, but is it correct that we these buttons don't treat `0` as either a valid min or max value?

dprzybylek: @kamilmateusiak could you comment? I don't know what did you have in mind :D

kamilmateusiak: @opif means that `max && parseInt(value, 10) === max` will turn out to be falsy here when the max is 0. You could check is `max !== undefined` (let's not be to strict here, there is no need to check for null and etc.)

kamilmateusiak: there is a couple of checks for `min` and `max` props in the code so you could add a helper `hasMax => this.props.max !== undefined`

kamilmateusiak:
I think it's not worth to change the type of NumericInput value. Of course, it's possible and many libraries support strings and numbers as value (like in carbon: https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/NumberInput/NumberInput.js or https://github.com/react-component/input-number/blob/master/src/index.js where it's a little too complicated)

packages/design-system/src/components/NumericInputField/NumericInput.js
dprzybylek: @kamilmateusiak  please look at 4f781f0dc390a57e0b0c8f5de6e8171a7e5c3448

packages/design-system/src/components/NumericInputField/NumericInput.js
```diff
@@ -9,12 +9,15 @@ const cx = classNames.bind(styles);
 const baseClass = 'numeric-input';
 
 class NumericInput extends React.PureComponent {
+  hasMin = () => this.props.min !== undefined;
+  hasMax = () => this.props.max !== undefined;
+  
   componentDidMount() {
-    if (this.props.max && parseInt(this.props.value, 10) > this.props.max) {
+    if (this.hasMax() && parseInt(this.props.value, 10) > this.props.max) {
```
walaszczykm: I found it a little bit weird that this `min` `max` values validation is happening on `componentDidMount` only. There is actually possibility that those values will change meanwhile component lifecycle. So, in that case, it could be done additionally in `componentDidUpdate` method or by using the derived state to keep those validated values in sync with raw props.

PR #166, opened 2020-02-05
AA-7719 ButtonGroup > Disable box-shadow for :focus and :active states

norbertprocak:
I used `ButtonGroup` in the agent app and saw that the `box-shadow` styles were placed in a different order. `.lc-btn:focus` was placed after `.lc-btn-group .lc-btn` which made the `box-shadow` of the regular button more important than disabled `box-shadow: none` for the grouped buttons. I wanted the styles of the group to be more specific.

### DIFF
--- a/packages/design-system/src/components/ButtonGroup/style.scss
+++ b/packages/design-system/src/components/ButtonGroup/style.scss
@@ -12,7 +12,6 @@
     &.active {
       border-color: $border-primary-color;
       background-color: $ui-background-color-hover;
-      z-index: 1;
       text-shadow: 0 0 0.5px black;
 
       &:hover {
@@ -25,7 +24,7 @@
     }
 
     &:focus, &:active {
-      box-shadow: none;
+      z-index: 1;
     }
 
     &:not(:first-child) {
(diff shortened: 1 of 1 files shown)


packages/design-system/src/components/ButtonGroup/style.scss
```diff
@@ -25,6 +24,10 @@
       background-color: $ui-background-color;
     }
 
+    &:focus, &:active {
+      box-shadow: none;
```
kamilmateusiak: yeah, it resets box-shadow but we don't have any alternative in ButtonGroup to indicate the button was focused. WCAG requires that the user is notified about a focused element

norbertprocak: So I kept the `box-shadow` and simply improved its display to avoid "cuts" like these:

Zrzut ekranu 2020-02-7 o 15 44 53

is that good ?

norbertprocak: Now it's this:
Zrzut ekranu 2020-02-7 o 15 47 39

kamilmateusiak: Awesome! I am glad you were able to improve it like that.

PR #167, opened 2020-02-05
Fix inner-shadow appearing on Safari mobile

klarzynskik:
It appears that the native behavior of Safari related to the `-webkit-appearance` has caused the inner shadows to be displayed inside of the `InputField`, `NumericInputField` and `TextAreaField `:

image

Using `-webkit-appearance: none;` might be an overkill, but after digging a bit it seems that `background-clip: padding-box;` does the trick.

I have also updated the snapshots for the NumericInput, as they seemed not to be up-to-date.

### DIFF
--- a/packages/design-system/src/components/InputField/style.scss
+++ b/packages/design-system/src/components/InputField/style.scss
@@ -10,7 +10,8 @@
   box-sizing: border-box;
   transition: border 0.25s ease-in-out;
   will-change: border-color;
-
+  background-clip: padding-box;
+ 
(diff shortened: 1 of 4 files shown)


dprzybylek:
@kamilmateusiak maybe we should merge it to master? Numeric input will block release from dev  :/

kamilmateusiak:
Can't do, 

> @kamilmateusiak maybe we should merge it to master? Numeric input will block release from dev :/

I cannot do that. This branch has recent changes from develop branch. I will release it today together with https://github.com/livechat/design-system/pull/164 (no need to wait for tests ion AA imho)

PR #168, opened 2020-02-06
Run npm test on push

kamilmateusiak:


kamilmateusiak:
It failed because of not updated snapshots. After merging https://github.com/livechat/design-system/pull/167 it should pass.

quarties:
@kamilmateusiak what do you think about using just [`husky`](https://www.npmjs.com/package/husky) package?

Then setup would look like this: https://github.com/livechat/customer-center/blob/master/package.json#L82

kamilmateusiak:
Yeah, we could use husky, but I do want to ask the PR author if tests passed. It should be visible in the PR. 
Actually we have it configured like that in the colors package: https://github.com/livechat/design-system/blob/develop/packages/colors/package.json

quarties:
> Yeah, we could use husky, but I do want to ask the PR author if tests passed. It should be visible in the PR.

Sure, I just think it's redundant as we have those checks in Bamboo. If we want to have them visible here, then I think we should ask our admins for help with setting up bamboo checks in GH PRs.

kamilmateusiak:
IMHO there is no need for setting this up from scratch on bamboo. We are only using bamboo to publish CSS styles on CDN. 
If you want to, feel free to ask them about that. Here, on GitHub, we have easy to set up an alternative and we have full control of it anytime.

PR #171, opened 2020-02-10
adjust inputs

dprzybylek:
Some changes in input fields:

1.  Changed inputs height to fit the designs. All should be 36px. -> https://app.zeplin.io/project/5b4fb0903176d0d90e5b04f8/screen/5be9ac5eccaa613e8652da64
2. Added optional header in select list, -> https://app.zeplin.io/project/5def6564c8bd560826d74d74/screen/5df0d5fb49d73eaeaf831fe0
3. Do not render clear button, instead of hiding it.

### DIFF
--- a/packages/design-system/src/components/MultiSelectField/MultiSelect.js
+++ b/packages/design-system/src/components/MultiSelectField/MultiSelect.js
@@ -396,7 +396,7 @@ class MultiSelect extends React.PureComponent {
       }),
       className
     );
-
+    const headItemsStyles = selectedItems && selectedItems.length > 0 ? `${baseClass}-head__items` : `${baseClass}-head__items-empty`;
     return (
       <div ref={this.containerRef} className={mergedClassNames} id={id}>
         <MultiSelectHead
@@ -408,7 +408,7 @@ class MultiSelect extends React.PureComponent {
           disabled={disabled}
         >
           <div
-            className={styles[`${baseClass}-head__items`]}
+            className={styles[headItemsStyles]}
             style={{ maxHeight: maxItemsContainerHeight }}
             ref={this.selectedItemsContainerRef}
           >
--- a/packages/design-system/src/components/MultiSelectField/style.scss
+++ b/packages/design-system/src/components/MultiSelectField/style.scss
@@ -6,7 +6,8 @@ $base-class: 'multiselect';
   position: relative;
   width: inherit;
   color: $font-primary-color;
-  font-size: 14px;
+  font-size: 15px;
+  line-height: 1.2;
 
   &-head {
     position: relative;
@@ -56,14 +57,13 @@ $base-class: 'multiselect';
 
   &-head__search {
     display: block;
-    width: 5px;
-    margin: 5px 10px;
-    line-height: 1.3;
+    margin: 2px 3px;
     color: $font-primary-color;
     border: none;
     background-color: transparent;
     box-sizing: border-box;
     outline: none;
+    line-height: 1.2;
 
     ::placeholder,
     ::-webkit-input-placeholder,
@@ -182,9 +182,20 @@ $base-class: 'multiselect';
     visibility: visible;
   }
 
+  &-head__items-empty {
+    padding: 5px 12px;
+    height: 100%;
+    display: flex;
+    flex-wrap: wrap;
+    align-items: center;
+    overflow-y: auto;
+    flex: 1;
+    opacity: 1;
+    visibility: visible;
+  }
+
   &__placeholder {
-    margin: 2px 3px;
-    line-height: 1.3;
+    margin: 3px 3px;
     opacity: 0.4;
     color: $font-primary-color;
   }
@@ -198,7 +209,7 @@ $base-class: 'multiselect';
     border-radius: 5px;
     padding: 2px 6px 4px 6px;
     line-height: normal;
-    margin: 3px 3px 1px 3px;
+    margin: 3px 3px 2px 3px;
     cursor: default;
     overflow: hidden;
     display: flex;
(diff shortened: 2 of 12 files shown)


packages/design-system/src/components/MultiSelectField/style.scss
```diff
@@ -6,6 +6,7 @@ $base-class: 'multiselect';
   position: relative;
   width: inherit;
   color: $font-primary-color;
+  font-size: 14px;
```
jdrzejb: is it 14px in ds? please consult with @jo

dprzybylek: yup, it's on all designs

dprzybylek: ok, I talk with @Jolama, should be 15 😢

packages/design-system/src/components/MultiSelectField/style.scss
```diff
@@ -57,9 +57,8 @@ $base-class: 'multiselect';
   &-head__search {
     display: block;
     width: 5px;
-    margin: 2px 3px;
-    font-size: 16px;
-    line-height: 28px;
+    margin: 5px 10px;
+    line-height: 1.3;
```
kamilmateusiak: I believe that ie11 or safari will give you a hard time after this change

packages/design-system/src/components/SelectField/ClearButton.js
```diff
@@ -1,21 +1,16 @@
...
 import CloseIcon from 'react-material-icon-svg/dist/CloseIcon';
-import classNames from 'classnames/bind';
 import styles from './style.scss';
 
 const baseClass = 'select-head';
-const cx = classNames.bind(styles);
 
 const ClearButton = React.forwardRef((props, ref) => {
-  const { isVisible, clearSelectedOption } = props;
+  const { clearSelectedOption } = props;
```
kamilmateusiak: you removed `isVisible` but it's still used in the `Select` component

packages/design-system/src/components/SelectField/Select.js
```diff
@@ -310,12 +312,17 @@ class Select extends React.PureComponent {
             onChange={this.onSearchChange}
             disabled={disabled}
           />
-          <ClearButton
-            isVisible={!!selectedItemModel && !isOpen && !required}
-            ref={this.clearButtonRef}
-            clearSelectedOption={this.clearSelectedOption}
-          />
-          <MenuDownIcon width="24px" height="24px" fill="#424d57" />
+          { shouldRenderClearButton &&
```
kamilmateusiak: if you do it like that we will lose fade-in transition ;x

packages/design-system/src/components/SelectField/SelectField.md
```diff
@@ -49,6 +49,7 @@ const getSelectedItemBody = props => {
     getSelectedItemBody={getSelectedItemBody}
     selected={state.selectedItem}
     searchPlaceholder='Search...'
+    selectHeader="Select from items below"
```
kamilmateusiak: https://take.ms/ZeJzK - should the scroll on keypress behave like that?

packages/design-system/src/components/SelectField/Select.js
kamilmateusiak: however maybe we don't need it so much, dunno

packages/design-system/src/components/MultiSelectField/style.scss
dprzybylek: you mean line height?

packages/design-system/src/components/MultiSelectField/style.scss
dprzybylek: done

packages/design-system/src/components/SelectField/SelectField.md
dprzybylek: I think so, you cannot select the header. What do you propose?

packages/design-system/src/components/SelectField/Select.js
dprzybylek: In my opinion we don't need it. As I see it makes no difference.

packages/design-system/src/components/MultiSelectField/MultiSelect.js
```diff
@@ -396,7 +396,7 @@ class MultiSelect extends React.PureComponent {
       }),
       className
     );
-
+    const headItemsStyles = selectedItems && selectedItems.length > 0 ? `${baseClass}-head__items` : `${baseClass}-head__items-empty`;
```
kamilmateusiak: don't you need selectHeader in the MUltiselect as well?

packages/design-system/src/components/SelectField/SelectField.md
kamilmateusiak: I tested it one more time and this scrolling with the keyboard doesn't work very well so let's leave it now
I expected that when the first element is selected it will be scrolled to the top and I will be able to see the header

dprzybylek:
@kamilmateusiak sure, this will be tested with agent app branch, after that I will merge it to develop.

packages/design-system/src/components/MultiSelectField/style.scss
kamilmateusiak: let's just test it, maybe it will work just fine xD

packages/design-system/src/components/MultiSelectField/MultiSelect.js
dprzybylek: for now not, in multiselect we have "select/deselect all" row, adding header could mess up.

PR #176, opened 2020-04-16
Tab - styling fix

jkrolak999:
There were wrong hover and selected styles on `Tabs`, they were using `#000`.

### DIFF
--- a/packages/design-system/src/components/Tab/style.scss
+++ b/packages/design-system/src/components/Tab/style.scss
@@ -5,7 +5,7 @@
   display: flex;
   flex-direction: row;
   align-items: center;
-  padding: 0 24px;
+  padding: 0 12px;
   border: none;
   background: none;
   outline: none;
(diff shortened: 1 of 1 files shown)


jkrolak999:
> Looks good. Have you had a chance to verify if there are places where we use wrong text color?

yes, I've checked it and this tabs was the last place where #000 was used

packages/design-system/src/components/Tab/style.scss
```diff
@@ -38,9 +38,10 @@
 
   &--selected {
     font-weight: 700;
+    color: $primary-color;
 
     .tab__description {
-      color: #8e949a;
+      color: $primary-color;
```
kamilmateusiak, line 44: Should't it be $secondary-color and only on hover $primary-color?

jkrolak999: this is description in selected state and I talked about it with Jo -> she decided to go for primary-color

kamilmateusiak:
@jdrzejb we are preparing themes in the design-system and all colors should be fixed anyway. A small preview for you is here: https://livechat-design-system-dev.netlify.app/#/Themes . I promise it will be available till the end of sprint 8.1 :)

PR #179, opened 2020-04-28
(DPS-1607) New props in button - loaderLabel

SamborGornicz:
Added new props that allow adding a label while the button is in the loading state.

Zrzut ekranu 2020-04-28 o 08 42 20

### DIFF
--- a/packages/design-system/src/components/Button/Button.js
+++ b/packages/design-system/src/components/Button/Button.js
@@ -75,7 +75,6 @@ const Button = React.forwardRef((props, ref) => {
           label={loaderLabel}
           className={styles[`${baseClass}__loader`]}
           labelClassName={styles[`${baseClass}__loader-label`]}
-          spinnerWrapperClassName={styles[`${baseClass}__loader-spinner`]}
         />
       )}
       {icon && <i className={styles[`${baseClass}__icon`]}>{icon}</i>}
--- a/packages/design-system/src/components/Button/Button.js
+++ b/packages/design-system/src/components/Button/Button.js
@@ -70,14 +70,13 @@ const Button = React.forwardRef((props, ref) => {
       aria-expanded={ariaExpanded}
     >
       {loading && (
-        <div className={styles[`${baseClass}__loader`]}>
-          <Loader size="small" />
-          {loaderLabel && (
-            <div className={styles[`${baseClass}__loader-label`]}>
-              {loaderLabel}
-            </div>
-          )}
-        </div>
+        <Loader
+          size="small"
+          label={loaderLabel}
+          className={styles[`${baseClass}__loader`]}
+          labelClassName={styles[`${baseClass}__loader-label`]}
+          spinnerWrapperClassName={styles[`${baseClass}__loader-spinner`]}
+        />
       )}
       {icon && <i className={styles[`${baseClass}__icon`]}>{icon}</i>}
       {children && <div>{children}</div>}
--- a/packages/design-system/src/components/Button/Button.js
+++ b/packages/design-system/src/components/Button/Button.js
@@ -73,9 +73,9 @@ const Button = React.forwardRef((props, ref) => {
         <div className={styles[`${baseClass}__loader`]}>
           <Loader size="small" />
           {loaderLabel && (
-            <span className={styles[`${baseClass}__loader-label`]}>
+            <div className={styles[`${baseClass}__loader-label`]}>
               {loaderLabel}
-            </span>
+            </div>
           )}
         </div>
       )}
--- a/packages/design-system/src/components/Button/Button.js
+++ b/packages/design-system/src/components/Button/Button.js
@@ -15,6 +15,7 @@ const Button = React.forwardRef((props, ref) => {
     destructive,
     disabled,
     loading,
+    loaderLabel,
     icon,
     size,
     fullWidth,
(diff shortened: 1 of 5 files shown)


packages/design-system/src/components/Button/Button.js
```diff
@@ -69,10 +70,17 @@ const Button = React.forwardRef((props, ref) => {
       aria-expanded={ariaExpanded}
     >
       {loading && (
-        <Loader size="small" className={styles[`${baseClass}__loader`]} />
+        <div className={styles[`${baseClass}__loader`]}>
+          <Loader size="small" />
```
kamilmateusiak, line 74: https://livechat.github.io/design-system/#/Components/Loader?id=loader-1 - on the bottom of the page you can find a loader with label:
```suggestion
          <Loader size="small" label={loaderLabel} labelClassName={styles[`${baseClass}__loader-label`]} />
```
Could you use it?  Or if it's hard there is LoaderLabel component which could be used instead of this custom div element

packages/design-system/src/components/Button/Button.md
```diff
@@ -179,6 +179,47 @@ Button expanded to the full width to fill the parent container (max-width: 320px
 </ComponentHtmlMarkup>
 ```
 
+### Loading with label
+
+```js
+  initialState = { loading: true };
+
+  toggleLoading = () => {
```
kamilmateusiak, line 187: I think you could use already existing section `### Loading` and add only one example to show that it's possible to use a label for loader

packages/design-system/src/components/Button/style.scss
```diff
@@ -131,9 +134,25 @@
...
     position: absolute;
-    top: 50%;
-    transform: translate(-50%, -50%);
-    left: 50%;
+    top: 0;
+    left: 0;
+    opacity: 1;
+    width: 100%;
+    height: 100%;
+    padding: 6px 16px;
```
kamilmateusiak, line 142: Can't you use `top: 50%` + `transform: translateY(-50%);` ? I mean this padding doesn't look good in a component with poissible various sizes (button can be small, normal and large)

packages/design-system/src/components/Button/style.scss
```diff
@@ -131,9 +134,25 @@
...
+    justify-content: center;
+    box-sizing: border-box;
+  }
+
+  .btn__loader-label {
+    text-align: left;
+    overflow: hidden;
+    white-space: nowrap;
+    text-overflow: ellipsis;
+    margin-left: 5px;
```
kamilmateusiak, line 154: if you decide to use Loader property instead of custom div remember to remove those margins

packages/design-system/src/components/Button/style.scss
kamilmateusiak: https://github.com/livechat/design-system/pull/179/files#diff-ce3c4a2a8c708ff0a84602f45be811a5R102 - different padding for smaller button

packages/design-system/src/components/Button/style.scss
```diff
@@ -131,9 +134,29 @@
...
+    height: 100%;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    box-sizing: border-box;
+    padding: inherit;
+  }
+
+  .btn__loader-spinner {
+    min-width: 15px;
```
kamilmateusiak, line 150: Why do you need it? Small loader has `width: 15px` by default.

kamilmateusiak: Design System - LiveChat API Developers Docs and Guides 2020-04-28 13-20-45
Ok, I understand xD I used width 100px for button and its height changed also ;/

kamilmateusiak: Looks like you found a bug in Loader with label. Please fix the Loader component spinner by adding `flex-shrink: 0;` to `.lc-loader-spinner-wrapper` should solve issue for all loaders

packages/design-system/src/components/Button/style.scss
```diff
@@ -131,9 +134,29 @@
...
+  }
+
+  .btn__loader-label {
+    text-align: left;
+    overflow: hidden;
+    white-space: nowrap;
+    text-overflow: ellipsis;
+    color: inherit;
+    margin: 0 5px 1px;
```
kamilmateusiak, line 159: do not understand this `margin-bottom: 1px;`, the label is centered with flexbox. Is it not working on some browser?

SamborGornicz: I've had a feeling that the label wasn't aligned on the y-axis, but it seems to be good now. I will push a fix in sec

packages/design-system/src/components/Button/style.scss
SamborGornicz: done

packages/design-system/src/components/Button/Button.js
```diff
@@ -69,10 +70,15 @@ const Button = React.forwardRef((props, ref) => {
...
+        <Loader
+          size="small"
+          label={loaderLabel}
+          className={styles[`${baseClass}__loader`]}
+          labelClassName={styles[`${baseClass}__loader-label`]}
+        />
       )}
       {icon && <i className={styles[`${baseClass}__icon`]}>{icon}</i>}
-      {children}
+      {children && <div>{children}</div>}
```
quarties, line 81: Why do we change it? It may break some implementations, tests, or custom stylings. I'm not saying it's bad, butI would like to know why do we change it :-)

kamilmateusiak: Actually, regardless of reason Sambor had to change it, we need that ;D take a look a the thread in Agent App: https://github.com/livechat/my.livechatinc.com/pull/12849#discussion_r406655529
google translate doesn't like our buttons with so-called "loose text" when loading or icon is rendered ;x

SamborGornicz: The reason behind that change is that the "loose text" is incompatible with [that line](https://github.com/livechat/design-system/pull/179/files#diff-ce3c4a2a8c708ff0a84602f45be811a5R116).

quarties: cool beans 👌

PR #188, opened 2020-05-26
PP-2458 - text button

quarties:


packages/design-system/src/components/Button/Button.md
```diff
@@ -54,6 +54,20 @@ Use as a final confirmation for a destructive action such as deleting. These are
 </ComponentHtmlMarkup>
 ```
 
+### Text Button
+
```
dprzybylek, line 58: Idea: could you add text button as a separate subsection? Like a ButtonGroup, it will be more visible.

packages/design-system/src/scss/variables/colors.scss
```diff
@@ -5,6 +5,7 @@ $hover-background-color: #e1e9ec;
 $hover-basic-color: #f9fbfb;
 $hover-primary-color: #4379d6;
 $hover-red-color: #b9484a;
+$hover-font-primary-blue-color: #295bb2;
```
dprzybylek, line 8: @kamilmateusiak please review colours 🙏

packages/design-system/src/components/Button/Button.md
quarties: Maybe we should create sub-sections for each type of button? @kamilmateusiak wdyt?

packages/design-system/src/scss/variables/colors.scss
quarties: > image
> Need to ask - we are fine with this outline?

It was based on [Zeplin project](https://zpl.io/VYBq7xk):
image

BTW Justyna will do the design review after demo :-)

quarties:
> Need to ask - we are fine with this outline?

@dprzybylek it was based on [Zeplin project](https://zpl.io/VYBq7xk):
image

BTW Justyna will do the design review after demo :-)

packages/design-system/src/components/Button/Button.js
```diff
@@ -26,20 +26,19 @@ const Button = React.forwardRef((props, ref) => {
...
-
-  if (primary) {
-    buttonType = 'primary';
-  } else if (destructive) {
-    buttonType = 'destructive';
-  } else if (secondary) {
-    buttonType = 'secondary';
-  }
+  const buttonType =
+    (primary && 'primary') ||
```
kamilmateusiak, line 37: the previous way to resolve `buttonType` was less cool but simpler and cleaner

packages/design-system/src/components/Button/Button.js
```diff
@@ -26,20 +26,19 @@ const Button = React.forwardRef((props, ref) => {
...
+    (destructive && 'destructive') ||
+    (secondary && 'secondary') ||
+    (text && 'text') ||
```
kamilmateusiak, line 40: `text` property name might be not the best here. It is usually used to pass some content to the component.

packages/design-system/src/components/Button/style.scss
```diff
@@ -160,7 +187,16 @@
...
-    margin-right: 5px;
+
+    &-left {
+      margin-right: 5px;
+      order: -1;
+    }
+
+    &-right {
+      margin-left: 5px;
+      order: 1;
```
kamilmateusiak, line 198: you are smart :D I would do some stuff with flex-direction xD

packages/design-system/src/scss/variables/colors.scss
kamilmateusiak: there is `$hover-primary-color: #4379d6;` already, no need for a new and pretty similar color

packages/design-system/src/scss/variables/colors.scss
```diff
@@ -14,6 +15,7 @@ $font-primary-color: rgba(66, 77, 87, 1);
 $font-secondary-color: rgba(66, 77, 87, 0.7);
 $font-tertiary-color: rgba(66, 77, 87, 0.5);
 $font-placeholder-color: rgba(66, 77, 87, 0.4);
+$font-primary-blue-color: $primary-color;
```
kamilmateusiak, line 18: Let's not complicate it now. Soon we will have a new color tokens

packages/design-system/src/components/Button/style.scss
```diff
@@ -90,6 +90,28 @@
...
+      color: $hover-font-primary-blue-color;
+      background-color: transparent;
+    }
+
+    &.btn .btn__loader {
+      justify-content: flex-start;
+    }
+
+    &.btn--large, &.btn--compact {
+      padding-left: 3px;
```
kamilmateusiak, line 110: hmm, usually larger button has bigger paddings: 
lc-btn--compact - `padding: 4px 16px;
lc-btn - `padding: padding: 6px 16px;`
lc-btn--large - `padding: 9px 24px;`

Shouldn't it be the same?

packages/design-system/src/components/Button/Button.js
kamilmateusiak: carbon has interesting name for this type of a button: "ghost" :D https://www.carbondesignsystem.com/components/button/usage

packages/design-system/src/components/Button/style.scss
quarties: I tried and I gave up xD Then I realized that there's still `order` ^^

packages/design-system/src/components/Button/style.scss
quarties: I'll discuss it Jo.

packages/design-system/src/components/Button/Button.js
quarties: TBH I see it exactly opposite. But it's only me, maybe I just like the cool and weird JS syntax ¯\_(ツ)_/¯

packages/design-system/src/components/Button/Button.js
quarties: Maybe we should migrate buttons to `type="text"` etc. (but keep current types declaration with boolean props as deprecated way). WDYT?

packages/design-system/src/scss/variables/colors.scss
quarties: But it's not the same as in design... I'm not sure what is the drill here, should we keep fewer colors to easier unify them later or stick with design?

packages/design-system/src/scss/variables/colors.scss
quarties: fixed in d51a706

packages/design-system/src/components/Button/Button.js
quarties: fixed in d51a706

packages/design-system/src/components/Button/Button.md
quarties: @kamilmateusiak ☝️ 🙏

packages/design-system/src/scss/variables/colors.scss
kamilmateusiak: yeah, the color is slightly different so the chance there is an issue with design is high. Please ask jo about that. Maybe those colors could be merged

packages/design-system/src/components/Button/Button.js
kamilmateusiak: I'd rather die than make changes again in AA ;x 
If we can support current props and the new `type` prop in the same time it could be a good idea. I saw that there is a few cases in AA where it's not handy to have `primary`, `secondary` etc props. The `type` property might be better.

packages/design-system/src/components/Button/Button.md
kamilmateusiak: I am against subsections for each type of button. It's exactly the same component with different configuration

quarties: I'll try to figure out something here.

packages/design-system/src/components/Button/Button.js
quarties: Yeah, I had the same issues. I think we should use `type` or `kind` prop and keep `primary` etc. as deprecated, but still usable ofc.

packages/design-system/src/components/Button/Button.md
quarties: Let's keep it as it is. I put my suggestion to our jira task.

packages/design-system/src/components/Button/Button.js
quarties: Done in 99b10e2c47c91f12884be7178dde5774cedf65b5

packages/design-system/src/interfaces/buttons.d.ts
```diff
@@ -20,6 +21,8 @@ export interface IButtonProps
   renderIcon?: string | React.ReactNode;
   ref?: React.Ref<HTMLButtonElement> | React.Ref<React.Component<IButtonProps>>;
   secondary?: boolean;
+  text?: boolean;
+  iconPosition?: ButtonIconPosition;
```
kamilmateusiak, line 25: 1. no `kind` property here 
2. you can add note about deprecated properties here for hints in code editor (example in tooltip.d.ts)
3. add `export type ButtonKind = "primary" | ... | "text"` above

packages/design-system/src/components/Button/Button.js
```diff
@@ -4,9 +4,27 @@ import classNames from 'classnames/bind';
 import styles from './style.scss';
 import getMergedClassNames from '../../utils/getMergedClassNames';
 import { Loader } from '../Loader';
+import { noticeAboutDeprecation } from '../../helpers/notice-about-deprecation';
 
 const cx = classNames.bind(styles);
 const acceptedSizes = ['large', 'compact'];
+const acceptedKinds = ['primary', 'secondary', 'destructive', 'text'];
+
+const getDeprecatedKind = (primary, destructive, secondary) => {
```
kamilmateusiak, line 13: if you have the mana (xD) for it you could move it to `helpers.ts`

packages/design-system/src/components/Button/Button.js
```diff
@@ -26,19 +44,21 @@ const Button = React.forwardRef((props, ref) => {
...
 
-  if (primary) {
-    buttonType = 'primary';
-  } else if (destructive) {
-    buttonType = 'destructive';
-  } else if (secondary) {
-    buttonType = 'secondary';
+  if (deprecatedKind) {
+    noticeAboutDeprecation(
+      `deprecated prop '${deprecatedKind}' - please use 'kind' prop instead`
```
kamilmateusiak, line 60: shouldn't you add a component name to the message to make it easier for devek to find it and fix?

packages/design-system/src/components/Button/style.scss
```diff
@@ -90,6 +90,28 @@
     }
   }
 
+  &--text {
+    background-color: transparent;
```
kamilmateusiak, line 94: Loader for other buttons (even when the text is long) is centered, but for `text` button it's aligned to left:
Monosnap 2020-06-05 17-35-12

packages/design-system/src/scss/variables/colors.scss
kamilmateusiak: kind reminder: ask jo about this color :)

packages/design-system/src/components/Button/Button.js
quarties: I thought about it, but it's not reusable at all, is it?

packages/design-system/src/components/Button/Button.js
```diff
@@ -48,14 +68,26 @@ const Button = React.forwardRef((props, ref) => {
...
-      [`${baseClass}--primary`]: buttonType === 'primary',
-      [`${baseClass}--destructive`]: buttonType === 'destructive',
-      [`${baseClass}--secondary`]: buttonType === 'secondary',
+      [`${baseClass}--${buttonKind}`]: !!buttonKind,
       [`${baseClass}--${size}`]: acceptedSizes.some(s => s === size)
     }),
     className
   );
 
+  const positionedIcon = (
```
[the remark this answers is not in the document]
quarties: 6de75d239c41beca9ffd5f3ee6d5b65204383df7

packages/design-system/src/interfaces/buttons.d.ts
quarties: 2ebe102873729d6f5b0b13d7b05f8cec0d521e16

packages/design-system/src/scss/variables/colors.scss
quarties: bb8c3a99f12b1b5c63a52627f4570cc29add5bca

packages/design-system/src/components/Button/Button.js
quarties: 6de75d239c41beca9ffd5f3ee6d5b65204383df7

packages/design-system/src/components/Button/style.scss
quarties: I'll discuss it with Jo

packages/design-system/src/components/Button/Button.js
kamilmateusiak: yes, but the component file could be smaller and more readable. Nvm, ignore it. The file is not that big and this method is declared before component so it's all good :)

packages/design-system/src/components/Button/Button.js
```diff
@@ -26,19 +45,21 @@ const Button = React.forwardRef((props, ref) => {
...
 
-  if (primary) {
-    buttonType = 'primary';
-  } else if (destructive) {
-    buttonType = 'destructive';
-  } else if (secondary) {
-    buttonType = 'secondary';
+  if (deprecatedKind) {
+    noticeAboutDeprecation(
+      `deprecated prop '${deprecatedKind}' in Button component - please use 'kind' prop instead`
```
kamilmateusiak, line 61: I think we should inform the AA teams about this change before merging it.

quarties: @kamilmateusiak good idea - could you take care of that? I'm not a member of AA private channels and I guess it's easier to announce there changes like this 😅

packages/design-system/src/components/Button/style.scss
quarties: Discussed and fixed in cb726c9 - text buttons will behave like other buttons :-)

packages/design-system/src/components/Button/style.scss
quarties: Discussed and fixed in cb726c9 - text buttons will behave like other buttons :-)

PR #192, opened 2020-06-18
add promo export

dprzybylek:
🤦‍♀️


packages/design-system/src/components/index.js
```diff
@@ -73,4 +73,7 @@ export { Loader } from './Loader';
 export { Divider } from './Divider';
 
 export { ProgressBar, ProgressCircle } from './Progress';
+
+export { Promo } from './Promo';
```
kamilmateusiak, line 77: do the same with Banner :)

dprzybylek: already done 😅

PR #193, opened 2020-06-18
fix name for banner, add export

dprzybylek:


packages/design-system/src/interfaces/promo.d.ts
```diff
@@ -2,7 +2,7 @@ import { MouseEventHandler } from "react";
 
 export type PromoSize = "small" | "medium" | "large";
 
-export interface IProgressBarProps extends React.HTMLAttributes<HTMLDivElement> {
+export interface IPromoProps extends React.HTMLAttributes<HTMLDivElement> {
```
kamilmateusiak, line 5: please remove: 
`import { MouseEventHandler } from "react";`
and add:
`// <reference types="react" />` instead and use `React.MouseEventHandler<HTMLButtonElement>` in interfaces

importing react here was problematic in the past

kamilmateusiak: use buttons.d.ts as reference

dprzybylek: done

PR #195, opened 2020-06-19
promo design review

dprzybylek:
bigger header
align margins

### DIFF
--- a/packages/design-system/src/components/Promo/style.scss
+++ b/packages/design-system/src/components/Promo/style.scss
@@ -15,45 +15,53 @@ $base-class: 'promo';
   align-items: center;
   justify-content: space-between;
   min-width: 180px;
-  margin-bottom: 8px;
+
+  &__content {
+    display: flex;
+    justify-content: flex-start;
+    flex-direction: row;
+  }
+
+  &__wrapper {
+    display: flex;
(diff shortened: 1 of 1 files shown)


packages/design-system/src/components/Promo/style.scss
```diff
@@ -81,6 +81,8 @@ $base-class: 'promo';
     }
     
     .#{$base-class}__header {
+      font-size: 16px;
```
kamilmateusiak, line 84: if the style is the same for every banner size you can declare it under `baseClass`. That way it will not be duplicated and the size of styles file will be smalller.

dprzybylek: done

kamilmateusiak: please move the whole `.#{$base-class}__header` under base class

dprzybylek: I removed margin, but I wouldn't remove the whole class. If i did it, children components will have bold font by default, that's a bad idea.

kamilmateusiak: Please take a look at https://github.com/livechat/design-system/pull/196. It seems that I can not describe it so I did a PR to this one with required changes.

PR #197, opened 2020-06-25
promo design review v2

dprzybylek:
* more place for text
* 20px margin on the right of img

### DIFF
--- a/packages/design-system/package.json
+++ b/packages/design-system/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@livechat/design-system",
-  "version": "0.9.1-beta.4",
+  "version": "0.9.1-beta.5",
   "publishConfig": {
     "access": "public"
   },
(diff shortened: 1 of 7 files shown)


packages/design-system/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@livechat/design-system",
-  "version": "0.9.0",
+  "version": "0.9.1-beta.4",
```
kamilmateusiak, line 3: bump to `0.9.1-beta.5`

dprzybylek: sure sure

packages/design-system/src/components/Promo/style.scss
```diff
@@ -92,7 +92,7 @@ $base-class: 'promo';
     }
 
     .#{$base-class}__wrapper {
-      max-width: 520px;
+      max-width: 680px;
```
kamilmateusiak, line 95: Banner should be changed as well. The same max-width value was there and I expect the same issue with spacings

dprzybylek: done, jo approved

PR #200, opened 2020-07-10
DS-7 SearchBar component

MrVorgoth:
**Jira:**
https://livechatinc.atlassian.net/browse/DS-7

**Description:**
Search bar with some options like:
- `controlled/uncontrolled` option (either setting state directly in component or passing value from parent component)
- `compact mode` - either for mobile or smaller screen designs
- `loading/error`- to further improve component -> display `error` or `loading` state (they need to be passed from above)
- `debounceTime` - to stop abusing search bar and lets you query only by specific number of `ms`
- `onChange/onSubmit` options - to have two modes: by either allowing to search by `enter` (onSubmit) or just via typing (maybe with debounced option to enhance user experience)
- `onClear` - when we want user to clear input, but don't make a search
- `onKeyDown(event)` - we want to allow user to use event for example `event.preventDefault()`

It's the first component I'm doing in `Design system` so I hope you can take a thorough look in this PR. Thanks :)

EDIT: Add debounce and throttle from lodash as dependecy and use it in our Design System (after talking with @kamilmateusiak )

### DIFF
--- a/packages/design-system/src/components/SearchBar/SearchBar.js
+++ b/packages/design-system/src/components/SearchBar/SearchBar.js
@@ -63,6 +63,8 @@ class SearchBar extends React.PureComponent {
       searchTerm: ''
     });
 
+    // 1) Cancel debounced search
+    // 2) Use onClear instead of onSubmit | onChange function
     searchFunction('');
 
     if (this.inputRef.current) {
@@ -90,22 +92,10 @@ class SearchBar extends React.PureComponent {
   };
 
   render() {
-    const {
-      className,
-      placeholder,
-      value,
-      loading,
-      size,
-      error,
-      onChange,
-      onSubmit,
-      ...restProps
-    } = this.props;
-
+    const { className, placeholder, value, loading, size, error } = this.props;
     const { searchTerm, isInCompactMode } = this.state;
 
     const isCompactSize = size === 'compact';
-
     const shouldDisplayCloseButton = searchTerm && !loading && !isInCompactMode;
 
     return (
@@ -130,7 +120,7 @@ class SearchBar extends React.PureComponent {
             placeholder={placeholder}
             ref={this.inputRef}
             value={value || this.state.searchTerm}
-            onInput={this.handleChange}
+            onChange={this.handleChange}
             onKeyDown={this.handleKeyPress}
             className={cx(`lc-${baseClass}__input`, {
               [`lc-${baseClass}__input-compact`]: isInCompactMode
--- a/packages/design-system/src/components/SearchBar/SearchBar.js
+++ b/packages/design-system/src/components/SearchBar/SearchBar.js
@@ -72,23 +72,25 @@ class SearchBar extends React.PureComponent {
     }
   };
 
-  handleCloseIconKeyDown = event => {
+  handleClearButtonKeyDown = event => {
     if (event.keyCode === KeyCodes.enter) {
       this.handleClear();
     }
   };
 
   toggleCompactMode = () => {
-    this.setState(
-      prevState => ({
-        isInCompactMode: !prevState.isInCompactMode
-      }),
-      () => {
-        if (!this.state.isInCompactMode && this.inputRef.current) {
-          this.inputRef.current.focus();
+    if (this.props.size === 'compact') {
+      this.setState(
+        prevState => ({
+          isInCompactMode: !prevState.isInCompactMode
+        }),
+        () => {
+          if (!this.state.isInCompactMode && this.inputRef.current) {
+            this.inputRef.current.focus();
+          }
         }
-      }
-    );
+      );
+    }
   };
 
   render() {
@@ -96,19 +98,23 @@ class SearchBar extends React.PureComponent {
     const { searchTerm, isInCompactMode } = this.state;
 
     const isCompactSize = size === 'compact';
-    const shouldDisplayCloseButton = searchTerm && !loading && !isInCompactMode;
+    const shouldDisplayClearButton = searchTerm && !loading && !isInCompactMode;
+
+    const searchIconClassName = cx(`lc-${baseClass}__icon--search`, {
+      [`lc-${baseClass}__icon--search-compact`]: isCompactSize
+    });
+    const inputClassName = cx(`lc-${baseClass}__input`, {
+      [`lc-${baseClass}__input-compact`]: isInCompactMode
+    });
 
     return (
       <span className={className}>
         <div className={styles[`${baseClass}__container`]}>
           <SearchIcon
-            fill="#424d57"
             width="18px"
             height="18px"
-            className={cx(`lc-${baseClass}__icon--search`, {
-              [`lc-${baseClass}__icon--search-compact`]: isCompactSize
-            })}
-            onClick={isCompactSize ? this.toggleCompactMode : null}
+            className={searchIconClassName}
+            onClick={this.toggleCompactMode}
           />
           {loading && (
             <Loader
@@ -122,18 +128,15 @@ class SearchBar extends React.PureComponent {
             value={value || this.state.searchTerm}
             onChange={this.handleChange}
             onKeyDown={this.handleKeyPress}
-            className={cx(`lc-${baseClass}__input`, {
-              [`lc-${baseClass}__input-compact`]: isInCompactMode
-            })}
+            className={inputClassName}
           />
-          {shouldDisplayCloseButton && (
+          {shouldDisplayClearButton && (
             <CloseIcon
-              fill="#424d57"
               width="18px"
               height="18px"
               onClick={this.handleClear}
-              onKeyDown={this.handleCloseIconKeyDown}
-              className={styles[`${baseClass}__icon--close`]}
+              onKeyDown={this.handleClearButtonKeyDown}
+              className={styles[`${baseClass}__icon--clear`]}
               tabIndex="0"
             />
           )}
@@ -144,7 +147,7 @@ class SearchBar extends React.PureComponent {
   }
 }
 
-const basePropTypes = {
+SearchBar.propTypes = {
   className: PropTypes.string,
   placeholder: PropTypes.string,
   /**
@@ -165,8 +168,7 @@ const basePropTypes = {
   onSubmit: PropTypes.func
 };
 
-/* eslint-disable react/default-props-match-prop-types */
-const baseDefaultProps = {
+SearchBar.defaultProps = {
   placeholder: 'Search...',
   value: null,
   loading: false,
@@ -176,9 +178,5 @@ const baseDefaultProps = {
   onChange: noop,
   onSubmit: null
 };
-
-SearchBar.propTypes = basePropTypes;
-
-SearchBar.defaultProps = baseDefaultProps;
 
 export default SearchBar;
--- a/packages/design-system/src/components/SearchBar/SearchBar.js
+++ b/packages/design-system/src/components/SearchBar/SearchBar.js
@@ -24,6 +24,7 @@ class SearchBarComponent extends React.PureComponent {
     this.setState({
       searchTerm: ''
     });
+    // TODO: what if we had two searchbars in code? Selecting by ID won't be good
     document.getElementById('search-bar-input').focus();
   };
 
@@ -78,6 +79,7 @@ class SearchBarComponent extends React.PureComponent {
             {...restProps}
           />
         </div>
+        <span className={styles[`${baseClass}__error`]}>{error}</span>
       </span>
     );
   }
--- a/packages/design-system/src/components/SearchBar/SearchBar.js
+++ b/packages/design-system/src/components/SearchBar/SearchBar.js
@@ -82,18 +82,16 @@ export class SearchBar extends React.PureComponent {
   };
 
   toggleCompactMode = () => {
-    if (this.props.collapsable) {
-      this.setState(
-        prevState => ({
-          isInCompactMode: !prevState.isInCompactMode
-        }),
-        () => {
-          if (!this.state.isInCompactMode && this.inputRef.current) {
-            this.inputRef.current.focus();
-          }
+    this.setState(
+      prevState => ({
+        isInCompactMode: !!this.props.collapsable && !prevState.isInCompactMode
+      }),
+      () => {
+        if (!this.state.isInCompactMode && this.inputRef.current) {
+          this.inputRef.current.focus();
         }
-      );
-    }
+      }
+    );
   };
 
   render() {
--- a/packages/design-system/src/components/SearchBar/SearchBar.js
+++ b/packages/design-system/src/components/SearchBar/SearchBar.js
@@ -55,15 +55,15 @@ export class SearchBar extends React.PureComponent {
   };
 
   handleClear = () => {
-    const { onChange, onSubmit } = this.props;
-    const searchFunction = onSubmit || onChange;
+    const { onChange, onSubmit, onClear } = this.props;
+    const searchFunction = onClear || onSubmit || onChange;
 
     this.setState({
       searchTerm: ''
     });
 
-    // 1) Cancel debounced search
-    // 2) Use onClear instead of onSubmit | onChange function
+    // TODO: Cancel debounced search
+    // Add debounce and throttle from lodash as dependecy and use it in our Design System (after talking with @kamilmateusiak )
     searchFunction('');
 
     if (this.inputRef.current) {
@@ -171,7 +171,8 @@ SearchBar.propTypes = {
   debounceInMs: PropTypes.number,
   error: PropTypes.string,
   onChange: PropTypes.func,
-  onSubmit: PropTypes.func
+  onSubmit: PropTypes.func,
+  onClear: PropTypes.func
 };
 
 SearchBar.defaultProps = {
@@ -182,7 +183,8 @@ SearchBar.defaultProps = {
   debounceInMs: 0,
   error: null,
   onChange: noop,
-  onSubmit: null
+  onSubmit: null,
+  onClear: null
 };
 
 export default SearchBar;
--- a/packages/design-system/src/components/SearchBar/SearchBar.js
+++ b/packages/design-system/src/components/SearchBar/SearchBar.js
@@ -118,6 +118,8 @@ export class SearchBar extends React.PureComponent {
       onClear,
       onSubmit,
       onChange,
+      onKeyDown,
+      onCollapse,
       ...restProps
     } = this.props;
     const { searchTerm, isInCompactMode } = this.state;
--- a/packages/design-system/src/components/SearchBar/SearchBar.js
+++ b/packages/design-system/src/components/SearchBar/SearchBar.js
@@ -45,7 +45,11 @@ export class SearchBar extends React.PureComponent {
   };
 
   handleKeyPress = event => {
-    const { onSubmit } = this.props;
+    const { onSubmit, onKeyDown } = this.props;
+
+    if (onKeyDown) {
+      onKeyDown(event);
+    }
 
     if (onSubmit) {
       if (event.keyCode === KeyCodes.enter) {
@@ -172,7 +176,11 @@ SearchBar.propTypes = {
   error: PropTypes.string,
   onChange: PropTypes.func,
   onSubmit: PropTypes.func,
-  onClear: PropTypes.func
+  onClear: PropTypes.func,
+  /**
+   * Pass onKeyDown function if you want to use event props for example event.preventDefault()
+   */
+  onKeyDown: PropTypes.func
 };
 
 SearchBar.defaultProps = {
@@ -184,7 +192,8 @@ SearchBar.defaultProps = {
   error: null,
   onChange: noop,
   onSubmit: null,
-  onClear: null
+  onClear: null,
+  onKeyDown: null
 };
 
 export default SearchBar;
--- a/packages/design-system/src/components/SearchBar/SearchBar.js
+++ b/packages/design-system/src/components/SearchBar/SearchBar.js
@@ -17,9 +17,9 @@ class SearchBar extends React.PureComponent {
     super(props);
 
     this.inputRef = React.createRef();
-    this.onChangeDebounced = debounce(this.props.debounceInMs, value => {
-      this.props.onChange(value);
-    });
+    // this.debouncedOnChange = debounce(this.props.debounceInMs, value => {
+    //   this.props.onChange(value);
+    // });
 
     this.state = {
       searchTerm: '',
@@ -36,7 +36,7 @@ class SearchBar extends React.PureComponent {
     });
 
     if (debounceInMs) {
-      this.onChangeDebounced(value);
+      this.debouncedOnChange(value);
     } else {
       onChange(value);
     }
@@ -57,19 +57,28 @@ class SearchBar extends React.PureComponent {
       searchTerm: ''
     });
 
-    this.inputRef.current.focus();
+    if (this.inputRef.current) {
+      this.inputRef.current.focus();
+    }
   };
 
-  handleKeyDown = key => {
+  handleCloseIconKeyDown = key => {
     if (key.keyCode === KeyCodes.enter) {
       this.handleClear();
     }
   };
 
   toggleCompactMode = () => {
-    this.setState(state => ({
-      isInCompactMode: !state.isInCompactMode
-    }));
+    this.setState(
+      prevState => ({
+        isInCompactMode: !prevState.isInCompactMode
+      }),
+      () => {
+        if (!this.state.isInCompactMode && this.inputRef.current) {
+          this.inputRef.current.focus();
+        }
+      }
+    );
   };
 
   render() {
@@ -125,7 +134,7 @@ class SearchBar extends React.PureComponent {
               width="18px"
               height="18px"
               onClick={this.handleClear}
-              onKeyDown={this.handleKeyDown}
+              onKeyDown={this.handleCloseIconKeyDown}
               className={styles[`${baseClass}__icon--close`]}
               tabIndex="0"
             />
--- a/packages/design-system/src/components/SearchBar/SearchBar.js
+++ b/packages/design-system/src/components/SearchBar/SearchBar.js
@@ -11,7 +11,6 @@ import FieldError from '../FieldError';
 import { Input } from '../InputField';
 import { KeyCodes } from '../../constants/keyCodes';
 
-const acceptedSizes = ['basic', 'compact'];
 const baseClass = 'search-bar';
 const noop = () => {};
 
@@ -26,7 +25,7 @@ class SearchBar extends React.PureComponent {
 
     this.state = {
       searchTerm: '',
-      isInCompactMode: this.props.size === 'compact'
+      isInCompactMode: props.collapsable
     };
   }
 
@@ -79,7 +78,7 @@ class SearchBar extends React.PureComponent {
   };
 
   toggleCompactMode = () => {
-    if (this.props.size === 'compact') {
+    if (this.props.collapsable) {
       this.setState(
         prevState => ({
           isInCompactMode: !prevState.isInCompactMode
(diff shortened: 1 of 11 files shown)


packages/design-system/src/components/SearchBar/SearchBar.js
```diff
@@ -0,0 +1,168 @@
...
+    this.inputRef.current.focus();
+  };
+
+  handleKeyDown = key => {
+    if (key.keyCode === KeyCodes.enter) {
+      this.handleClear();
+    }
+  };
+
+  toggleCompactMode = () => {
```
kamilmateusiak, line 69: awesome, but let's make it easier for a user and do `this.inputRef.current.focus();` after `setState`

packages/design-system/src/components/SearchBar/SearchBar.js
```diff
@@ -0,0 +1,168 @@
...
+    this.setState(state => ({
+      isInCompactMode: !state.isInCompactMode
+    }));
+  };
```
kamilmateusiak, line 73: awesome, but let's make it easier for a user and do `this.inputRef.current.focus();` after `setState` if it was openned:
```suggestion
  toggleCompactMode = () => {
    this.setState(prevState => ({
      isInCompactMode: !state.isInCompactMode
    }), () => {
      if (!this.state.isInCompactMode && this.inputRef.current) {
        this.inputRef.current.focus();
      }
    });
  };
```

packages/design-system/src/components/SearchBar/SearchBar.js
```diff
@@ -0,0 +1,168 @@
...
+      }
+    }
+  };
+
+  handleClear = () => {
+    this.setState({
+      searchTerm: ''
+    });
+
+    this.inputRef.current.focus();
```
kamilmateusiak, line 60: please add check for `if (this.inputRef.current) { this.inputRef.current.focus() }`

packages/design-system/src/components/SearchBar/SearchBar.js
```diff
@@ -0,0 +1,168 @@
...
+          )}
+        </div>
+        <span className={styles[`${baseClass}__error`]}>{error}</span>
+      </span>
+    );
+  }
+}
+
+const basePropTypes = {
+  className: PropTypes.string,
```
kamilmateusiak, line 141: you need to add the props interface to forms.d.ts file to use it in Agent App or any other TS app

packages/design-system/src/components/SearchBar/SearchBar.js
```diff
@@ -0,0 +1,168 @@
...
+            type="input"
+            placeholder={placeholder}
+            ref={this.inputRef}
+            value={value || this.state.searchTerm}
+            onInput={this.handleChange}
+            onKeyDown={this.handleKeyPress}
+            className={cx(`lc-${baseClass}__input`, {
+              [`lc-${baseClass}__input-compact`]: isInCompactMode
+            })}
+            {...restProps}
```
kamilmateusiak, line 120: Monosnap 2020-08-11 11-23-27
you need to omit this property

packages/design-system/src/components/SearchBar/SearchBar.js
```diff
@@ -0,0 +1,168 @@
...
+            })}
+            onClick={compact ? this.toggleCompactMode : null}
+          />
+          {loading && (
+            <Loader
+              size="small"
+              className={styles[`${baseClass}__icon--loader`]}
+            />
+          )}
+          <input
```
kamilmateusiak, line 110: can't u use input from `Input.js`?

styleguide.config.js
```diff
@@ -136,6 +136,12 @@ module.exports = {
           components: [
             './packages/design-system/src/components/Loader/Loader.js'
           ]
+        },
+        {
+          name: 'SearchBar',
```
kamilmateusiak, line 141: it should be sub-item of "Forms" section

packages/design-system/src/components/SearchBar/SearchBar.md
```diff
@@ -0,0 +1,46 @@
...
+Most common use case for search bar is with onChange handler. It can be enhanced with debounce method to reduce many callbacks
+
+```js
+const onChange = value => {
+  console.log('Search term: ', value);
+}
+<SearchBar onChange={onChange} />
+```
+
+### Search bar with onSubmit handler (see console)
```
kamilmateusiak, line 26: no need to add this "(see console)", dev is able to see the source code of the example

packages/design-system/src/components/SearchBar/SearchBar.js
```diff
@@ -0,0 +1,168 @@
...
+
+    if (onSubmit) {
+      if (key.keyCode === KeyCodes.enter) {
+        onSubmit(this.state.searchTerm);
+      }
+    }
+  };
+
+  handleClear = () => {
+    this.setState({
```
kamilmateusiak, line 56: you need to inform the parent component about clear if the component is controlled (controlled===has value prop from the outside). Please add `onClear` property (you can't just use onChange :) )

packages/design-system/src/components/SearchBar/SearchBar.js
```diff
@@ -0,0 +1,168 @@
...
+    });
+
+    if (debounceInMs) {
+      this.onChangeDebounced(value);
+    } else {
+      onChange(value);
+    }
+  };
+
+  handleKeyPress = key => {
```
kamilmateusiak, line 45: what if `event.preventDefault();` will be needed?:)

packages/design-system/src/components/SearchBar/SearchBar.js
```diff
@@ -0,0 +1,168 @@
...
+
+const baseClass = 'search-bar';
+const noop = () => {};
+
+class SearchBar extends React.PureComponent {
+  constructor(props) {
+    super(props);
+
+    this.inputRef = React.createRef();
+    this.onChangeDebounced = debounce(this.props.debounceInMs, value => {
```
kamilmateusiak, line 20: move it out from constructor and rename it to `debouncedOnChange` 
also if you use debounce you need to make sure that debounce is canceled before the component will unmount

packages/design-system/src/components/SearchBar/SearchBar.js
```diff
@@ -0,0 +1,168 @@
...
+
+  handleClear = () => {
+    this.setState({
+      searchTerm: ''
+    });
+
+    this.inputRef.current.focus();
+  };
+
+  handleKeyDown = key => {
```
kamilmateusiak, line 63:
```suggestion
  handleCloseIconKeyDown = key => {
```

packages/design-system/src/components/SearchBar/SearchBar.js
```diff
@@ -0,0 +1,168 @@
...
+              width="18px"
+              height="18px"
+              onClick={this.handleClear}
+              onKeyDown={this.handleKeyDown}
+              className={styles[`${baseClass}__icon--close`]}
+              tabIndex="0"
+            />
+          )}
+        </div>
+        <span className={styles[`${baseClass}__error`]}>{error}</span>
```
kamilmateusiak, line 134: we have a `FieldError` component for that. You won't need to duplicate html and styles

packages/design-system/src/components/SearchBar/style.scss
```diff
@@ -0,0 +1,70 @@
...
+
+  &:focus, &:active {
+    outline: none;
+    border-color: $primary-color
+  }
+}
+
+.#{$base-class}__input-compact {
+  width: 0;
+  padding-right: 0;
```
kamilmateusiak, line 59: the animation is jumping because of this padding-right reset. Can we improve that?

packages/design-system/src/components/SearchBar/SearchBar.js
```diff
@@ -0,0 +1,168 @@
...
+  }
+}
+
+const basePropTypes = {
+  className: PropTypes.string,
+  placeholder: PropTypes.string,
+  value: PropTypes.string,
+  loading: PropTypes.bool,
+  compact: PropTypes.bool,
+  debounceInMs: PropTypes.number,
```
kamilmateusiak, line 146: add some comment to this prop, it will appear in "Props & methods" section of docs 
right now it's not clear what it is

packages/design-system/src/components/SearchBar/SearchBar.js
```diff
@@ -0,0 +1,168 @@
...
+    );
+  }
+}
+
+const basePropTypes = {
+  className: PropTypes.string,
+  placeholder: PropTypes.string,
+  value: PropTypes.string,
+  loading: PropTypes.bool,
+  compact: PropTypes.bool,
```
kamilmateusiak, line 145:
```suggestion
  collapsed: PropTypes.bool,
```
I see in the design that search bar can be compact which means that the input height is smaller etc. Can we add `size` property? (`size?: 'compact'`, something more in the future is possible)

packages/design-system/src/components/SearchBar/SearchBar.js
MrVorgoth: Why move it our from the constructor?

packages/design-system/src/components/SearchBar/SearchBar.js
MrVorgoth: Can you tell me more about it? Because right now I'm confused if I should use `e.preventDefault()` in handleKeyPress function or use it based on some param

packages/design-system/src/components/SearchBar/SearchBar.js
MrVorgoth: I added comments to 3 props:
- `debounceInMs`,
- `compact` (which, will be changed later for `size` as per comment below),
- `value`

packages/design-system/src/components/SearchBar/SearchBar.js
MrVorgoth: I added those props to `forms.d.ts` file, but I'm requesting your check :)

packages/design-system/src/components/SearchBar/SearchBar.js
MrVorgoth: Can you be more precise or explain why I can't use onChange / onSubmit with empty value? I just did something like that in this PR. Should i have another prop `onClear`, which will inform user above about clearing input? Right now I don't see it, because I can use onClear() inside SearchBar component - this will delegate to component above and what does it give me?

packages/design-system/src/components/SearchBar/style.scss
MrVorgoth: It's improved :) I added width and padding to animation

packages/design-system/src/components/SearchBar/SearchBar.js
kamilmateusiak: hmm, let's add optional property `onKeyDown`, if someone will need to call `preventDefault` method of event it will be possible,
here you will need to add the following code:
```
if (this.props.onKeyDown) {
  this.props.onKeyDown(event)
}
```

packages/design-system/src/components/SearchBar/SearchBar.js
```diff
@@ -0,0 +1,184 @@
...
+    });
+
+    if (debounceInMs) {
+      this.debouncedOnChange(value);
+    } else {
+      onChange(value);
+    }
+  };
+
+  handleKeyPress = key => {
```
kamilmateusiak, line 48:
```suggestion
  handleKeyPress = event => {
```

packages/design-system/src/components/SearchBar/SearchBar.js
kamilmateusiak: Just add `onClear` to the `searchFunction` definition: `const searchFunction = onClear || onSubmit || onChange;`
If the developer needs different behavior he will use this property. Let's say it won't be a good idea to fetch new data on clear button click - it seems to me like a common case when you need to optimize requests.

packages/design-system/src/components/SearchBar/SearchBar.js
```diff
@@ -0,0 +1,184 @@
...
+    // 1) Cancel debounced search
+    // 2) Use onClear instead of onSubmit | onChange function
+    searchFunction('');
+
+    if (this.inputRef.current) {
+      this.inputRef.current.focus();
+    }
+  };
+
+  handleCloseIconKeyDown = key => {
```
kamilmateusiak, line 75:
```suggestion
  handleCloseIconKeyDown = event => {
```

packages/design-system/src/components/SearchBar/SearchBar.js
```diff
@@ -0,0 +1,184 @@
...
+  placeholder: PropTypes.string,
+  /**
+   * Pass this prop if you want to control searchbar term from parent component
+   */
+  value: PropTypes.string,
+  loading: PropTypes.bool,
+  /**
+   * Pass size of `compact` if you want to display extendable searchbar (usually for smaller screens)
+   */
+  size: PropTypes.oneOf(acceptedSizes),
```
kamilmateusiak, line 158: rather than using `size` property for this collapsed input I would use `collapsable: boolean` 
I see on design smaller input (for tables etc.). The size of text and high of input should depend on this `size` property instead of collapse mode.

packages/design-system/src/components/SearchBar/SearchBar.md
```diff
@@ -0,0 +1,46 @@
...
+}
+<SearchBar size="compact" onChange={onChange} />
+```
+
+### Search bar with onChange handler
+
+Most common use case for search bar is with onChange handler. It can be enhanced with debounce method to reduce many callbacks
+
+```js
+const onChange = value => {
```
kamilmateusiak, line 20: can you make the loader visible in docs? maybe some timeout to mock a call to api

packages/design-system/src/components/SearchBar/style.scss
```diff
@@ -0,0 +1,61 @@
+@import '../../scss/variables/colors';
+
+$base-class: "search-bar";
+
+// TODO: should container have fixed width and height?
```
kamilmateusiak, line 5: it should have height. Width should be 100%. It's pretty common now that you need to add styles to input to use it in AA because it doesn't have `width: 100%`.

packages/design-system/src/interfaces/forms.d.ts
```diff
@@ -1,5 +1,9 @@
 // <reference types="react" />
 
+export type SearchBarSize = "basic" | "compact";
+type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
+type OmitSearchBar = Omit<ISearchBarProps, "debounceInMs">;
```
kamilmateusiak, line 5: Something bad happened here `OmitSearchBar` type is based on `ISearchBarProps` but `ISearchBarProps` extends `OmitSearchBar`. It's like a circular dependency. Also I don't know why you need to omit `debounceInMs`?

packages/design-system/src/interfaces/forms.d.ts
```diff
@@ -174,6 +178,17 @@ export interface IFieldGroupProps extends React.HTMLAttributes<HTMLElement> {
   stretch?: boolean;
 }
 
+export interface ISearchBarProps extends OmitSearchBar {
```
kamilmateusiak, line 181:
```suggestion
export interface ISearchBarProps extends React.HTMLAttributes<HTMLElement> {
```

packages/design-system/src/components/SearchBar/SearchBar.js
```diff
@@ -0,0 +1,184 @@
...
+
+  render() {
+    const { className, placeholder, value, loading, size, error } = this.props;
+    const { searchTerm, isInCompactMode } = this.state;
+
+    const isCompactSize = size === 'compact';
+    const shouldDisplayCloseButton = searchTerm && !loading && !isInCompactMode;
+
+    return (
+      <span className={className}>
```
kamilmateusiak, line 102: div in span is probably not a good idea. try using https://validator.w3.org/nu/#textarea. It will throw an error: `Error: Element div not allowed as child of element span in this context. (Suppressing further errors from this subtree.)`

styleguide.config.js
```diff
@@ -109,6 +109,7 @@ module.exports = {
         {
           name: 'Forms',
           components: [
+            './packages/design-system/src/components/SearchBar/SearchBar.js',
```
kamilmateusiak, line 112: There is also a file named `setup.js` and `packages/design-system/src/components/index.js`. You need to add this new component there to make it work in external apps.

kamilmateusiak (changes requested):
some stuff needs to be fixed

packages/design-system/src/components/SearchBar/SearchBar.js
```diff
@@ -0,0 +1,184 @@
...
+        <div className={styles[`${baseClass}__container`]}>
+          <SearchIcon
+            fill="#424d57"
+            width="18px"
+            height="18px"
+            className={cx(`lc-${baseClass}__icon--search`, {
+              [`lc-${baseClass}__icon--search-compact`]: isCompactSize
+            })}
```
quarties, line 110: Let's put it out from JSX code to a constant - it will be more clear and easier to find in the future.

packages/design-system/src/components/SearchBar/SearchBar.js
quarties, line 75: Again, let's use `ClearButton` as name here to keep consistency

packages/design-system/src/components/SearchBar/SearchBar.js
```diff
@@ -0,0 +1,184 @@
...
+    // 2) Use onClear instead of onSubmit | onChange function
+    searchFunction('');
+
+    if (this.inputRef.current) {
+      this.inputRef.current.focus();
+    }
+  };
+
+  handleCloseIconKeyDown = key => {
+    if (key.keyCode === KeyCodes.enter) {
```
quarties, line 76: @kamilmateusiak should this be enter or space? I would personally try to use space, as the element would be focused like a link.

packages/design-system/src/components/SearchBar/SearchBar.js
```diff
@@ -0,0 +1,184 @@
...
+        }
+      }
+    );
+  };
+
+  render() {
+    const { className, placeholder, value, loading, size, error } = this.props;
+    const { searchTerm, isInCompactMode } = this.state;
+
+    const isCompactSize = size === 'compact';
```
quarties, line 98: Isn't it the same as `isInCompactMode`?

packages/design-system/src/components/SearchBar/SearchBar.js
```diff
@@ -0,0 +1,184 @@
...
+      <span className={className}>
+        <div className={styles[`${baseClass}__container`]}>
+          <SearchIcon
+            fill="#424d57"
+            width="18px"
+            height="18px"
+            className={cx(`lc-${baseClass}__icon--search`, {
+              [`lc-${baseClass}__icon--search-compact`]: isCompactSize
+            })}
+            onClick={isCompactSize ? this.toggleCompactMode : null}
```
quarties, line 111: Maybe let's mobe this logic inside to `toggleCompactMode` method?

packages/design-system/src/components/SearchBar/SearchBar.js
quarties: SGTM ☝️

packages/design-system/src/components/SearchBar/SearchBar.js
```diff
@@ -0,0 +1,184 @@
...
+            />
+          )}
+        </div>
+        {error && <FieldError>{error}</FieldError>}
+      </span>
+    );
+  }
+}
+
+const basePropTypes = {
```
quarties, line 147: Why do you create a const if you can set it directly in `SearchBar`? It's not exported or something.

packages/design-system/src/components/SearchBar/style.scss
```diff
@@ -0,0 +1,61 @@
...
+  &:focus, &:active {
+    outline: none;
+    border-color: $primary-color
+  }
+}
+
+.#{$base-class}__input-compact {
+  width: 0;
+  padding-right: 0;
+}
```
quarties, line 61: trailing new line 🙏

packages/design-system/src/components/SearchBar/SearchBar.js
```diff
@@ -0,0 +1,184 @@
...
+    const { searchTerm, isInCompactMode } = this.state;
+
+    const isCompactSize = size === 'compact';
+    const shouldDisplayCloseButton = searchTerm && !loading && !isInCompactMode;
+
+    return (
+      <span className={className}>
+        <div className={styles[`${baseClass}__container`]}>
+          <SearchIcon
+            fill="#424d57"
```
quarties, line 105: I guess we have color SCSS variable - let's move the `fill` definition to CSS code to use a variable there.

packages/design-system/src/components/SearchBar/style.scss
```diff
@@ -0,0 +1,61 @@
...
+    &-compact {
+      cursor: pointer;
+    }
+  }
+
+  &--close {
+    @extend .#{$base-class}__icon;
+    right: 13px;
+    cursor: pointer;
+  }
```
quarties, line 33: It should be `--clear`

quarties: Also, accessibility states (especially focus) should be set for this.

packages/design-system/src/components/SearchBar/SearchBar.js
```diff
@@ -0,0 +1,184 @@
...
+      () => {
+        if (!this.state.isInCompactMode && this.inputRef.current) {
+          this.inputRef.current.focus();
+        }
+      }
+    );
+  };
+
+  render() {
+    const { className, placeholder, value, loading, size, error } = this.props;
```
quarties, line 95: Please use `...restProps` and pass it to the container - it's useful for tests and development.

quarties (changes requested):

packages/design-system/src/components/SearchBar/SearchBar.js
MrVorgoth: But but, it was like that earlier and you wanted a `size` property :D I'll change it back then

packages/design-system/src/components/SearchBar/style.scss
MrVorgoth: I've added `width: 100%`, but I've got a question. Should I leave `max-width: 290px` or nope? Because right now my component spans for 100% width and it doesn't look good. Or maybe it should stay for 100% width and lets user wrap it inside some component and style it on their own when importing this component from design system?

packages/design-system/src/components/SearchBar/SearchBar.js
MrVorgoth: You're correct. I've looked on many DS components and tried to do the same (to follow convention)

packages/design-system/src/components/SearchBar/SearchBar.js
```diff
@@ -0,0 +1,184 @@
...
+import CloseIcon from 'react-material-icon-svg/dist/CloseIcon';
+import { debounce } from '@livechat/data-utils';
+
+import styles from './style.scss';
+import { Loader } from '../Loader';
+import FieldError from '../FieldError';
+import { Input } from '../InputField';
+import { KeyCodes } from '../../constants/keyCodes';
+
+const acceptedSizes = ['basic', 'compact'];
```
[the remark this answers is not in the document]
MrVorgoth: Hey, I've deleted the `size` property and renamed it to `collapsable` :)

packages/design-system/src/components/SearchBar/style.scss
MrVorgoth: I've changed it to `--clear`. I didn't add any accessiblity states, because frankly i don't know how they should look like. Should I discuss it with designer? On focus I see that there is blue outline near clear button "X"

packages/design-system/src/components/SearchBar/style.scss
kamilmateusiak: In other libraries, there is no max-width for input components. Let's leave it like that.

packages/design-system/src/components/SearchBar/SearchBar.js
kamilmateusiak: usually, the `space` just scrolls the page so imho no need for that

packages/design-system/src/components/SearchBar/style.scss
kamilmateusiak: Use the same style as in the Banner component

packages/design-system/src/components/SearchBar/SearchBar.js
```diff
@@ -0,0 +1,199 @@
...
+  };
+
+  render() {
+    const {
+      className,
+      placeholder,
+      value,
+      loading,
+      collapsable,
+      error,
```
kamilmateusiak, line 106: in `restProps` you will have all remaining props. if you don't want to pass `debounceInMs` to the main `div` you should take it from it as you do for `collapsable` etc.

packages/design-system/src/components/SearchBar/SearchBar.js
```diff
@@ -0,0 +1,199 @@
...
+    if (event.keyCode === KeyCodes.enter) {
+      this.handleClear();
+    }
+  };
+
+  toggleCompactMode = () => {
+    if (this.props.collapsable) {
+      this.setState(
+        prevState => ({
+          isInCompactMode: !prevState.isInCompactMode
```
kamilmateusiak, line 88: this.props.collapsable && !prevState.isInCompactMode and remove above if

packages/design-system/src/components/SearchBar/SearchBar.js
```diff
@@ -0,0 +1,199 @@
...
+  handleClear = () => {
+    const { onChange, onSubmit, onClear } = this.props;
+    const searchFunction = onClear || onSubmit || onChange;
+
+    this.setState({
+      searchTerm: ''
+    });
+
+    // TODO: Cancel debounced search
+    // Add debounce and throttle from lodash as dependecy and use it in our Design System (after talking with @kamilmateusiak )
```
kamilmateusiak, line 70: import { debounce } from 'lodash/debounce'

packages/design-system/src/interfaces/forms.d.ts
```diff
@@ -174,6 +174,17 @@ export interface IFieldGroupProps extends React.HTMLAttributes<HTMLElement> {
   stretch?: boolean;
 }
 
+export interface ISearchBarProps {
```
kamilmateusiak, line 177:
```suggestion
type ISearchBarHtmlProps = Omit<Omit<React.HTMLAttributes<HTMLDivElement>, 'onChange'>, 'onSubmit'>

export interface ISearchBarProps extends ISearchBarHtmlProps {
```

quarties (approved):
✅ except Kamil's comments

packages/design-system/src/components/SearchBar/style.scss
```diff
@@ -0,0 +1,62 @@
...
+    padding: 7px 13px;
+    fill: $icon-primary-color;
+
+    &-compact {
+      cursor: pointer;
+    }
+  }
+
+  &--clear {
+    @extend .#{$base-class}__icon;
```
kamilmateusiak, line 30: no need for extend

packages/design-system/src/components/SearchBar/style.scss
```diff
@@ -0,0 +1,62 @@
+@import '../../scss/variables/colors';
+
```
kamilmateusiak, line 2: add @import "../../scss/mixins/button";

packages/design-system/src/components/SearchBar/style.scss
```diff
@@ -0,0 +1,62 @@
...
+      cursor: pointer;
+    }
+  }
+
+  &--clear {
+    @extend .#{$base-class}__icon;
+    right: 13px;
+    cursor: pointer;
+    fill: $icon-primary-color;
+  }
```
kamilmateusiak, line 34:
```suggestion
  @include icon-button-interactive-styles;
  }
```

this will reset browser specific styles and gives us control how the button looks

packages/design-system/src/components/SearchBar/SearchBar.js
```diff
@@ -0,0 +1,209 @@
...
+    }
+  };
+
+  toggleCompactMode = () => {
+    this.setState(
+      prevState => ({
+        isInCompactMode: !!this.props.collapsable && !prevState.isInCompactMode
+      }),
+      () => {
+        if (!this.state.isInCompactMode && this.inputRef.current) {
```
kamilmateusiak, line 98:
```suggestion
        if (this.state. isInCompactMode && this.props.onCollapse) {
          this.props.onCollapse()
        }

        if (!this.state.isInCompactMode && this.inputRef.current) {
```

packages/design-system/src/components/SearchBar/style.scss
MrVorgoth: Hmm without extend my styles are breaking 😕

kamilmateusiak: that is because you used only one class in jsx: className={styles[`${baseClass}__icon--clear`]} instead of className={cx(styles[`${baseClass}__icon`], styles[`${baseClass}__icon--clear`])}

packages/design-system/src/components/SearchBar/SearchBar.js
MrVorgoth: Added this prop in 9ff2ce5

kamilmateusiak (approved):

PR #202, opened 2020-08-10
TextField labelAdornment

walaszczykm:
TextField labelAdornment property allows displaying a React node next to the label as a way to decorate the whole input element with an additional element like icon or tooltip trigger and etc.

The main motivation for adding this prop is to allow extending already existing text explanation for input with some possible interactive decoration elements. The label adornment is treated generically it only restricts the dimensions and layout but is more like a "render spot" for the consumer content rather than the actual element.

### DIFF
--- a/packages/design-system/src/components/InputField/InputField.js
+++ b/packages/design-system/src/components/InputField/InputField.js
@@ -39,6 +39,7 @@ const InputField = React.forwardRef((props, ref) => {
 
 InputField.propTypes = {
   labelText: PropTypes.string,
+  /**  specifies additional decorative element rendered at the end of the label */
   labelAdornment: PropTypes.node,
   id: PropTypes.string.isRequired,
   onChange: PropTypes.func.isRequired,
--- a/packages/design-system/src/components/InputField/InputField.js
+++ b/packages/design-system/src/components/InputField/InputField.js
@@ -9,6 +9,7 @@ const InputField = React.forwardRef((props, ref) => {
     error,
     description,
     labelText,
+    labelAdornment,
     className,
     id,
     fieldClassName,
@@ -21,6 +22,7 @@ const InputField = React.forwardRef((props, ref) => {
       error={error}
       description={description}
       labelText={labelText}
+      labelAdornment={labelAdornment}
       className={className}
       htmlFor={id}
     >
@@ -37,6 +39,7 @@ const InputField = React.forwardRef((props, ref) => {
 
 InputField.propTypes = {
   labelText: PropTypes.string,
+  labelAdornment: PropTypes.node,
   id: PropTypes.string.isRequired,
   onChange: PropTypes.func.isRequired,
   className: PropTypes.string,
--- a/packages/design-system/src/components/TextField/TextField.js
+++ b/packages/design-system/src/components/TextField/TextField.js
@@ -15,6 +15,7 @@ const TextField = props => {
     error,
     description,
     labelText,
+    labelAdornment,
     className,
     htmlFor,
     children
@@ -34,10 +35,16 @@ const TextField = props => {
       {labelText && (
         <div
           className={cx({
+            [`${baseClass}__label`]: true,
             [`${baseClass}__label--inline`]: inline
           })}
         >
           <FieldLabel htmlFor={htmlFor}>{labelText}</FieldLabel>
+          {labelAdornment && (
+            <div className={cx({ [`${baseClass}__label-adornment`]: true })}>
+              {labelAdornment}
+            </div>
+          )}
         </div>
       )}
       <div>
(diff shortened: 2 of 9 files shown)


packages/design-system/src/components/TextField/TextField.js
```diff
@@ -34,10 +35,16 @@ const TextField = props => {
       {labelText && (
         <div
           className={cx({
+            [`${baseClass}__label`]: true,
             [`${baseClass}__label--inline`]: inline
           })}
         >
           <FieldLabel htmlFor={htmlFor}>{labelText}</FieldLabel>
+          {labelAdornment && (
+            <div className={cx({ [`${baseClass}__label-adornment`]: true })}>
```
Andarist, line 44: using `cx` is just needed for composing several conditional classes (adding an unconditional one with `: true` is just a convenience thing), this is equivalent to:
```
className={`${baseClass}__label-adornment`}
```

walaszczykm: without the `cx` there will not be added a prefix for the class name `lc-*`

packages/design-system/src/components/TextField/TextField.js
```diff
@@ -51,6 +58,7 @@ const TextField = props => {
 
 TextField.propTypes = {
   labelText: PropTypes.string,
+  labelAdornment: PropTypes.node,
```
kamilmateusiak, line 61: Add a comment about what it is - your comment will appear in the docs the "Props & methods" section of TextField properties. 
Try to add it to the example of TextField.

walaszczykm: @kamilmateusiak in case of the example I was feeling a little bit confused as in the docs for `TextxField` component there are no dedicated examples for particular props so it would look like something very special 😕 the only example is for using your own input and that is all. But ofc I would add an explanation as a comment for the prop itself as you suggested 🙂 👍

packages/design-system/src/components/InputField/InputField.js
```diff
@@ -37,6 +39,7 @@ const InputField = React.forwardRef((props, ref) => {
 
 InputField.propTypes = {
   labelText: PropTypes.string,
+  labelAdornment: PropTypes.node,
```
kamilmateusiak, line 42: add this prop to ITextFieldProps forms.d.ts

packages/design-system/src/components/TextField/TextField.js
kamilmateusiak: you don't need `cx`, you could use className={styles[`${baseClass}__label-adornment`]} - the class name needs to be "binded" to styles

kamilmateusiak:
Monosnap 2020-08-11 10-12-45
common problem with flex

walaszczykm:
> Monosnap 2020-08-11 10-12-45
> 
> common problem with flex, the icon shrinked :( why do you need `overflow: hidden` there?

@kamilmateusiak I didn't want to allow rendering the content over the desired place with restricted dimensions. Maybe we could just add `flex-shrink: 0` in combination with fixed width and height it should resolve the problem. Or maybe it would be better to remove the `overflow: hidden`?

kamilmateusiak:
> @kamilmateusiak I didn't want to allow rendering the content over the desired place with restricted dimensions. Maybe we could just add `flex-shrink: 0` in combination with fixed width and height it should resolve the problem. Or maybe it would be better to remove the `overflow: hidden`?

Let's add `flex-shrink: 0;`
Also please add `margin-left: 5px;` and `justify-content: flex-start` because if someone use element with size 20x20 there won't be any margin

packages/design-system/src/interfaces/forms.d.ts
```diff
@@ -102,6 +102,7 @@ export interface IInputProps
 
 export interface IInputFieldProps extends IInputProps {
   labelText?: string;
+  labelAdornment?: React.ReactNode;
```
kamilmateusiak, line 105: the same goes for IMultiSelectFieldProps, ITextAreaFieldProps, INumericInputFieldProps, ISelectFieldProps
lets create:
```
export interface IWithTextFieldProps {
  labelText?: string;
  id: string;
  inline?: boolean;
  description?: React.ReactNode;
  error?: string;
  labelAdornment?: React.ReactNode;
}
```
and extend mentioned interfaces with it

packages/design-system/src/components/TextField/TextField.js
```diff
@@ -58,6 +58,7 @@ const TextField = props => {
 
 TextField.propTypes = {
   labelText: PropTypes.string,
+  /**  specifies additional decorative element rendered at the end of the label */
```
kamilmateusiak, line 61: I know that it's duplicating code but could you add this comment also in NumericInputField.js, SelectField.js, MultiselectField.js, TextareaField.js? We will refactor it later.

packages/design-system/src/interfaces/forms.d.ts
kamilmateusiak: Could you do me a favor?:D We need to also change one property in buttonGroup.d.ts - `size` should be optional.

walaszczykm: sure no problem :D 👍

PR #209, opened 2020-10-26
AA-10215: Rewrite next/prev year logic

AlithAnar:
Jira: https://livechatinc.atlassian.net/browse/AA-10215

### DIFF
--- a/packages/design-system/src/components/DatePicker/DatePicker.js
+++ b/packages/design-system/src/components/DatePicker/DatePicker.js
@@ -58,6 +58,25 @@ class DatePicker extends React.PureComponent {
     );
   };
 
+  onSetRef = datePicker => {
+    this.datePicker = datePicker
+    if (this.props.innerRef) {
+      this.props.innerRef(datePicker)
+    }
+  }
+
+  handleNextYear = () => {
+    if (this.datePicker) {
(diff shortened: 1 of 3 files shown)


packages/design-system/src/components/DatePicker/DatePicker.js
```diff
@@ -58,6 +58,25 @@ class DatePicker extends React.PureComponent {
     );
   };
 
+  onSetRef = datePicker => {
+    this.datePicker = datePicker
+    if (this.props.innerRef) {
+      this.props.innerRef(datePicker)
```
kamilmateusiak, line 64: this is a breaking change;/ until now it was possible to pass React.Ref here 
```
type Ref<T> = { bivarianceHack(instance: T | null): void }["bivarianceHack"] | RefObject<T> | null;
```

kamilmateusiak:
Monosnap 2020-10-27 16-03-16
Take a look at a working example. It working without your changes but you need to move logic to the parent component and handle it yourself using state. 

I agree that this is not how it should be handled so I added my implementation in the PR: https://github.com/livechat/design-system/pull/210. That way we will not introduce any breaking changes and handle the issue :)

AlithAnar:
Closing in favor of #210

PR #214, opened 2020-11-27
DS-121 Improve search bar close icon

MrVorgoth:
Jira: https://livechatinc.atlassian.net/browse/DS-121

Hey, according to the above task the close icon in search bar is disappearing quicker from the DOM than some helper function/hook a developer wants to use and it is causing problems. I've prepared a fix that icon is always visible in the DOM (`display: none/block`) and even if user will turn this icon on through dev console then nothing will happen (blocked handlers on this icon).

Moreover I'm trying new concept of `data-element` attributes to help user find elements inside given component. WDYT about this approach?

PS: In the code there are some prettier code formatting changes

### DIFF
--- a/packages/design-system/src/components/SearchBar/SearchBar.js
+++ b/packages/design-system/src/components/SearchBar/SearchBar.js
@@ -194,7 +194,7 @@ export class SearchBar extends React.PureComponent {
     const closeIconClassName = cx(
       [`lc-${baseClass}__icon`, `lc-${baseClass}__clear-icon`],
       {
-        [`lc-${baseClass}__clear-icon--disable`]: !shouldDisplayClearButton
+        [`lc-${baseClass}__clear-icon--hidden`]: !shouldDisplayClearButton
       }
     );
 
@@ -225,7 +225,7 @@ export class SearchBar extends React.PureComponent {
             onBlur={this.handleBlur}
             onKeyDown={this.handleKeyPress}
             className={inputClassName}
-            data-element="search-input"
+            data-testId="search-input"
           />
           <CloseIcon
             width="18px"
@@ -236,7 +236,7 @@ export class SearchBar extends React.PureComponent {
             }
             className={closeIconClassName}
             tabIndex="0"
-            data-element="close-icon"
+            data-testId="close-button"
           />
         </div>
         {error && <FieldError>{error}</FieldError>}
@@ -281,7 +281,7 @@ SearchBar.propTypes = {
   onExpand: PropTypes.func,
   onFocus: PropTypes.func,
   onBlur: PropTypes.func,
-  forwardedRef: PropTypes.func
+  forwardedRef: React.RefObject
 };
 
 SearchBar.defaultProps = {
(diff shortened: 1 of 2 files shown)


packages/design-system/src/components/SearchBar/SearchBar.js
```diff
@@ -268,14 +280,15 @@ SearchBar.propTypes = {
...
+  onBlur: PropTypes.func,
+  forwardedRef: PropTypes.func
 };
 
 SearchBar.defaultProps = {
   placeholder: 'Search...',
   debounceTime: 0
 };
 
-export default React.forwardRef(function SearchBarWithRef(props, ref) {
```
kamilmateusiak, line 279: if you just use an anonymous arrow function here the component will be harder to find in React Dev Tools (https://reactjs.org/docs/forwarding-refs.html#displaying-a-custom-name-in-devtools) + possible error message will be less helpful

packages/design-system/src/components/SearchBar/SearchBar.js
```diff
@@ -268,14 +280,15 @@ SearchBar.propTypes = {
   onCollapse: PropTypes.func,
   onExpand: PropTypes.func,
   onFocus: PropTypes.func,
-  onBlur: PropTypes.func
+  onBlur: PropTypes.func,
+  forwardedRef: PropTypes.func
```
kamilmateusiak, line 284: What about `React.RefObject`?

packages/design-system/src/components/SearchBar/style.scss
```diff
@@ -50,6 +51,10 @@ $base-class: "search-bar";
       box-sizing: border-box;
       border: 1px solid #8db4ee;
     }
+
+    &--disable {
```
kamilmateusiak, line 55:
```suggestion
    &--hidden {
```
`disable` usually means something different

packages/design-system/src/components/SearchBar/SearchBar.js
```diff
@@ -212,20 +225,19 @@ export class SearchBar extends React.PureComponent {
...
+          <CloseIcon
+            width="18px"
+            height="18px"
+            onClick={shouldDisplayClearButton ? this.handleClear : null}
+            onKeyDown={
+              shouldDisplayClearButton ? this.handleClearButtonKeyDown : null
+            }
+            className={closeIconClassName}
+            tabIndex="0"
+            data-element="close-icon"
```
kamilmateusiak, line 239: I was thinking about that recently and read a couple of articles about e2e tests and selectors. 
`className` is guaranteed to not change in design-system lib but maybe we should think about this property just as a way to style things. `data-testId` could inform about what is this element and it could be used in tests. Also there is a smaller chance that it will change (than class name). It will be the same until the element is removed or replaced with something else.

To sum it up - let's add `data-` attribute but it would be better to use the convention proposed in AA + use `button` word instead of `icon` since it's clickable:
```suggestion
            data-testId="close-button"
```

kamilmateusiak (changes requested):
some minor things to be fixed

quarties (approved):
LGTM (except @kamilmateusiak comments) 💪

packages/design-system/src/components/SearchBar/SearchBar.js
MrVorgoth: I'm all in for that approach :)

kamilmateusiak (approved):

PR #224, opened 2021-02-15
Fix prop-types of UserGuideTooltip

jakubsikora:
UserGuideTooltip were using `Element` directly which was not working with SSR (next.js in my case).


kamilmateusiak (approved):
❤️

PR #231, opened 2021-03-14
AA-122 Add Icon component

MrVorgoth:
Adding Icon wrapper component with props:
- required icon (imported icon)
- required size (one of specific sizes) - interface is exported
- color (one of specific colors) - interface is also exported


package.json
```diff
@@ -24,6 +24,7 @@
     "polished": "^3.4.2"
   },
   "devDependencies": {
+    "@svgr/webpack": "^5.5.0",
```
kamilmateusiak, line 27: this package has a crazy amount of dependencies but it should be ok because it's installed only for generating docs

packages/design-system/src/components/Icon/Icon.js
```diff
@@ -0,0 +1,46 @@
...
+   */
+  icon: PropTypes.func.isRequired,
+  /**
+   * Defines width and height of the icon
+   */
+  iconSize: PropTypes.oneOf(['XSmall', 'Small', 'Medium', 'Large', 'XLarge'])
+    .isRequired,
+  // iconColor: PropTypes.oneOf(Object.keys(IconColorName)),
+  iconColor: PropTypes.oneOf([
+    'IconColorPrimaryDefault',
```
kamilmateusiak, line 33: doesn't make sense to me
Instead of controlling icon with simple boolean prop `disabled` some1 will need to pass `iconColor=IconColorPrimaryDisabled` 
It's better to just hardcode those tokens and use them if `disabled` or particular `iconType` is passed to komponent

packages/design-system/src/components/Icon/Icon.js
```diff
@@ -0,0 +1,46 @@
...
+  /**
+   * Icon passed in svg format
+   */
+  icon: PropTypes.func.isRequired,
+  /**
+   * Defines width and height of the icon
+   */
+  iconSize: PropTypes.oneOf(['XSmall', 'Small', 'Medium', 'Large', 'XLarge'])
+    .isRequired,
+  // iconColor: PropTypes.oneOf(Object.keys(IconColorName)),
```
kamilmateusiak, line 31: remove comment

packages/design-system/src/components/Icon/Icon.md
```diff
@@ -0,0 +1,60 @@
...
+
+### Icon Sizes
+
+One way to show the icon sizes
+```js
+  <>
+    <p>Extra small icon (16x16)</p>
+    <Icon icon={ChatsIcon} iconSize={IconSizeName.XSmall} />
+
+    <p>Smal icon (20x20)</p>
```
kamilmateusiak, line 14: typo

packages/design-system/src/components/Icon/Icon.md
```diff
@@ -0,0 +1,60 @@
...
+    <Icon icon={ChatsIcon} iconColor={IconColorName.IconColorLinkDefault} iconSize={IconSizeName.Small} />
+
+    <p>Mediun icon (24x24)</p>
```
kamilmateusiak, line 17: typo

packages/design-system/src/components/Icon/Icon.md
```diff
@@ -0,0 +1,60 @@
...
+    <Icon icon={ChatsIcon} iconColor={IconColorName.IconColorWarningDefault} iconSize={IconSizeName.Large} />
+  
+    <p>Extra large icon (48x48)</p>
+    <Icon icon={ChatsIcon} iconColor={IconColorName.IconColorSuccessDefault} iconSize={IconSizeName.XLarge} />
+  </>
+```
+
+### Icon Sizes
+
+And the other way to show icon sizes
```
kamilmateusiak, line 30: If Bartek will provide you some descriptions for those icon size (ie. when to use each size) then let's use the second way to show icons docs. If we don't plan to explain anything then the first one is sufficient

packages/design-system/src/components/Icon/index.js
```diff
@@ -0,0 +1,3 @@
+export { default as Icon } from './Icon';
+
+export { IconSizeName, IconColorName } from './constants';
```
kamilmateusiak, line 3:
```suggestion
export { IconSizeName as IconSize, IconColorName as IconColor } from './constants';
```
can we rename those on export outside?

packages/design-system/src/components/Icon/Icon.js
```diff
@@ -0,0 +1,46 @@
...
+
+Icon.propTypes = {
+  /**
+   * Icon passed in svg format
+   */
+  icon: PropTypes.func.isRequired,
+  /**
+   * Defines width and height of the icon
+   */
+  iconSize: PropTypes.oneOf(['XSmall', 'Small', 'Medium', 'Large', 'XLarge'])
```
kamilmateusiak, line 29: can we rename it to be `size`? It would be better to omit `icon` prefix for the component API accessible outside.

packages/design-system/src/components/Icon/Icon.js
```diff
@@ -0,0 +1,46 @@
...
+      {GeneratedIcon}
+    </span>
+  );
+};
+
+Icon.propTypes = {
+  /**
+   * Icon passed in svg format
+   */
+  icon: PropTypes.func.isRequired,
```
kamilmateusiak, line 25: `<Icon icon={required('///')} />` vs `<Icon source={required('///')} />`
wdyt?

packages/design-system/src/interfaces/forms.d.ts
```diff
@@ -181,6 +181,13 @@ export interface ISearchBarProps extends ISearchBarHtmlProps {
   onExpand?(): void;
 }
 
+export interface IIconProps {
+  icon: React.ReactNode;
```
kamilmateusiak, line 185: that is not the right type if you want to handle require()

packages/design-system/src/components/Icon/helpers.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+
+export function generateIcon(icon, iconSize, iconColor) {
```
kamilmateusiak, line 3: if it's that simple maybe we could directly use this `React.createElement` in the component instead of some mysterious generateIcon

packages/design-system/src/interfaces/forms.d.ts
```diff
@@ -181,6 +181,13 @@ export interface ISearchBarProps extends ISearchBarHtmlProps {
   onExpand?(): void;
 }
 
+export interface IIconProps {
```
kamilmateusiak, line 184: please extend some html interface to make use of this rest operator you use in the component

packages/design-system/src/components/Icon/helpers.js
MrVorgoth: Sure, we can. I wanted to extract it to helpers to have less code in Icon file, but indeed it's a short function

packages/design-system/src/components/Icon/Icon.js
MrVorgoth: yeah, I think `source` is the better name here than `icon` 👍

packages/design-system/src/interfaces/forms.d.ts
MrVorgoth: Can you tell me what's the proper type for handling `required` as I can't seem to find it :(

packages/design-system/src/components/Icon/index.js
kamilmateusiak: instead of IconColor export IconType

packages/design-system/src/components/Icon/Icon.js
```diff
@@ -0,0 +1,51 @@
...
+  return (
+    <span {...restProps} className={className}>
+      {GeneratedIcon}
+    </span>
+  );
+};
+
+Icon.propTypes = {
+  /**
+   * Icon passed in svg format
```
kamilmateusiak, line 28:
```suggestion
   * Icon passed in svg format, ie. `source={require('svg_icon')}
```

packages/design-system/src/components/Icon/constants.js
```diff
@@ -0,0 +1,57 @@
...
+};
+
+export const IconColorMapper = {
+  [IconTypeName.Primary]: '#424d57',
+  [IconTypeName.Inverted]: '#fff',
+  [IconTypeName.Link]: '#4384f5',
+  [IconTypeName.Success]: '#4bb678',
+  [IconTypeName.Warning]: '#efa842',
+  [IconTypeName.Error]: '#d64646'
+};
```
MrVorgoth, line 48: I was also thinking about passing disabled prop (to avoid duplication), but I think if we were to create some other props then different dictionaries seems like a better idea. WDYT?

```suggestion
export const IconColorMapper = disabled => ({
  [IconTypeName.Primary]: disabled ? '#686d72' : '#424d57',
  [IconTypeName.Inverted]: '#fff',
  [IconTypeName.Link]: '#4384f5',
  [IconTypeName.Success]: '#4bb678',
  [IconTypeName.Warning]: '#efa842',
  [IconTypeName.Error]: '#d64646'
});
```

packages/design-system/src/interfaces/forms.d.ts
```diff
@@ -183,6 +183,14 @@ export interface ISearchBarProps extends ISearchBarHtmlProps {
   onExpand?(): void;
 }
 
+export interface IIconProps extends React.HTMLAttributes<HTMLSpanElement> {
+  source: string;
+  size?: string;
+  iconType?: string;
```
kamilmateusiak, line 189: use correct type here:
`export type IconType = 'primary' | 'inverted' | 'link' | 'success' | 'warning' | 'error'`
The same goes for size, let's declare some type instead of using just string - it doesn't tell anything to developer

packages/design-system/src/components/index.js
```diff
@@ -79,3 +79,5 @@ export { Promo } from './Promo';
 export { UploadBar, FileUploadProgress } from './UploadBar';
 
 export { SearchBar } from './SearchBar';
+
+export { Icon, IconSize, IconType } from './Icon';
```
kamilmateusiak, line 83: You need to declare and export types for IconSize and IconType to make it accessible in TS projects

kamilmateusiak:
something is not working in Agent App - I receive an error: 
Dock 2021-04-23 11-27-10

If you want to try it out by yourself please install @livechat/design-system@0.12.0-beta.0 in AA

PR #232, opened 2021-03-25
PP-4192 - react 17 support

quarties:


packages/design-system/tests/setupTests.js
```diff
@@ -1,8 +1,9 @@
 import { configure } from 'enzyme';
-import Adapter from 'enzyme-adapter-react-16';
+import Adapter from '@wojtekmaj/enzyme-adapter-react-17';
```
quarties, line 2: I know it's not an official one, but it's still in progress: https://github.com/enzymejs/enzyme/pull/2430
However `@wojtekmaj` is one of the contributors to the official package, so I decided to go with his solution.

packages/design-system/package.json
```diff
@@ -97,8 +97,8 @@
     "typescript": "^2.9.1"
   },
   "peerDependencies": {
-    "react": "16",
-    "react-dom": "16"
+    "react": "16 || 17",
```
kamilmateusiak, line 100: can we enforce at least version [16.13.1](https://github.com/facebook/react/blob/master/CHANGELOG.md#16131-march-19-2020) here?:) I would like to have react version with hooks and some improvements

quarties (commented):
It looks like one 'Select' example is broken (_Select with controlled dropdown visibility_
): http://localhost:6060/#!/Select/7
When I click on the 'Show' button it's changing focus to the select field, but it's not opening at all (BTW similar example with MultiSelect works like a charm).

docs/customized-components/Logo.js
```diff
@@ -38,7 +38,7 @@ const styles = ({ color }) => ({
 
 export function LogoRenderer({ classes }) {
   return (
-    <a href="/design-system" className={classes.wrapper}>
```
quarties, line 41: I think this should not get changed

kamilmateusiak: We have a bug related to this link so I wanted to fix it by the way :)

PR #234, opened 2021-04-30
Introducing chars left counter

dprzybylek:
Chars left counter will be helpful to show how many chars are left in inputs with `maxlength` defined.
Value change should be handled by `onChange` method, in the same way as validation functions. `maxLength` is also passed to html input, to prevent exceeding the chars limit.
image


packages/design-system/src/components/InputField/InputField.md
```diff
@@ -27,20 +27,55 @@ onInputChange = (e) => {
...
+    setState({
+      value: e.target.value,
+      error: 'Validation message here',
+      chars: 0
+    });
+  } else {
+    setState({
+      value: e.target.value,
+      error: null,
+      chars: 5 - e.target.value.length
```
kamilmateusiak, line 63: I think the example is not working correctly. https://www.figma.com/file/KCYyNivnhN4wSDOM0Qjas4/KnowledgeBase-onboarding?node-id=845%3A6022 - here we have ie. 13/20 - that means that 13 characters were used. In your example it's the opposite - you are showing how many chars are left which doesn't feel intuitive imho

packages/design-system/src/components/TextField/style.scss
```diff
@@ -38,4 +45,10 @@ $base-class: 'text-field';
     align-items: center;
     justify-content: flex-start;
   }
+  &__label-charsCounter {
```
kamilmateusiak, line 48: Design System - LiveChat API Developers Docs and Guides 2021-05-04 12-05-17
I aksed Jo about this - both labelAdornment and new element should be positioned with `flex-end`

packages/design-system/src/components/InputField/InputField.js
```diff
@@ -13,6 +13,8 @@ const InputField = React.forwardRef((props, ref) => {
     className,
     id,
     fieldClassName,
+    maxLength,
+    charsLeft,
```
kamilmateusiak, line 17: I asked Justyna about the [design](https://www.figma.com/file/KCYyNivnhN4wSDOM0Qjas4/KnowledgeBase-onboarding?node-id=845%3A6022). It looks like on the right it should be possible to add also something different (ie. link) so those props looks not right. 

How about a different approach - let's start exporting hooks from the library! `useUsedCharactersCounter` hook seems awesome imho. You could use it in your app and pass the prepared text to the component with new props `labelRightNode`. How it sounds?

kamilmateusiak (changes requested):
It would be awesome to try a different approach here. I can't wait to see what you think about it :)

jdrzejb:
The current implementation forces you to use the counter with the input, the counter is not reusable in other components (like textarea), and you cannot use the top-right area to display anything else in the input.

dprzybylek:
Update: no hooks. After discuss we don't see any reason to add custom hook 🙈

packages/design-system/src/components/TextField/style.scss
```diff
@@ -36,6 +43,14 @@ $base-class: 'text-field';
     flex-shrink: 0;
     display: flex;
     align-items: center;
-    justify-content: flex-start;
+    justify-content: flex-end;
+  }
+  
+  &__label-leftNode {
```
kamilmateusiak, line 49:
```suggestion
  &__label-right-node {
```
xD

packages/design-system/src/components/TextField/style.scss
```diff
@@ -36,6 +43,14 @@ $base-class: 'text-field';
...
+    justify-content: flex-end;
```
kamilmateusiak, line 50: add `align-self: flex-end;`

packages/design-system/src/components/TextField/style.scss
```diff
@@ -36,6 +43,14 @@ $base-class: 'text-field';
     flex-shrink: 0;
     display: flex;
     align-items: center;
-    justify-content: flex-start;
+    justify-content: flex-end;
```
kamilmateusiak, line 46: add `align-self: flex-end;`

packages/design-system/src/components/InputField/InputField.js
```diff
@@ -47,7 +51,9 @@ InputField.propTypes = {
   fieldClassName: PropTypes.string,
   inline: PropTypes.bool,
   error: PropTypes.string,
-  description: PropTypes.node
+  description: PropTypes.node,
+  labelRightNode: PropTypes.node,
```
kamilmateusiak, line 55: could you please add those new props to forms.d.ts file with type declarations?

kamilmateusiak (changes requested):
the component api looks good now but it requires css and typescript fixes

PR #246, opened 2021-05-19
Allow to add only right node label

dprzybylek:
USer should be able to add only right node without a label

### DIFF
--- a/packages/design-system/src/components/InputField/InputField.md
+++ b/packages/design-system/src/components/InputField/InputField.md
@@ -114,6 +114,7 @@ onInputChange = (e) => {
   placeholder='Placeholder...'
   description='Text longer than 5 character will trigger error'
   onChange={onInputChange}
+  labelRightNode="foo"
 />
 ```
 
--- a/packages/design-system/src/components/InputField/style.scss
+++ b/packages/design-system/src/components/InputField/style.scss
@@ -11,6 +11,7 @@
   transition: border 0.25s ease-in-out;
   will-change: border-color;
   background-clip: padding-box;
+  width: 100%;
  
   &::placeholder { /* Chrome, Firefox, Opera, Safari 10.1+ */
     color: $font-primary-color;
--- a/packages/design-system/src/components/TextField/TextField.js
+++ b/packages/design-system/src/components/TextField/TextField.js
@@ -33,7 +33,7 @@ const TextField = props => {
 
   return (
     <div className={mergedClassNames}>
-      {labelText && (
+      {(labelText || labelRightNode) && (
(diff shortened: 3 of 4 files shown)


packages/design-system/src/components/TextField/TextField.js
```diff
@@ -33,7 +33,7 @@ const TextField = props => {
 
   return (
     <div className={mergedClassNames}>
-      {labelText && (
+      {(labelText || labelRightNode) && (
```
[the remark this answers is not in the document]
kamilmateusiak: Actually, by "Correct :D" I meant that you are correct, not that this behavior is correct. @dprzybylek can we cover that somehow?

packages/design-system/src/components/TextField/style.scss
```diff
@@ -34,6 +34,10 @@ $base-class: 'text-field';
     margin-right: 10px;
   }
 
+  &__label--inline--center {
+    margin-top: 20px;
```
kamilmateusiak, line 38: please no :) we need to avoid such magic values - it will come back to us sooner or later. I know that the structure of the component makes it hard to handle inline case but margin-top is definitely not the right thing to do here

packages/design-system/src/components/TextField/TextField.js
```diff
@@ -33,11 +33,12 @@ const TextField = props => {
...
         <div
           className={cx({
             [`${baseClass}__label`]: true,
-            [`${baseClass}__label--inline`]: inline
+            [`${baseClass}__label--inline`]: inline,
+            [`${baseClass}__label--inline--center`]: labelRightNode
```
kamilmateusiak, line 41: this class was applied also to `not-inline` component ;x

packages/design-system/src/components/InputField/InputField.md
```diff
@@ -114,6 +114,7 @@ onInputChange = (e) => {
   placeholder='Placeholder...'
   description='Text longer than 5 character will trigger error'
   onChange={onInputChange}
+  labelRightNode="foo"
```
kamilmateusiak, line 117: use different value

packages/design-system/src/components/InputField/style.scss
```diff
@@ -11,6 +11,7 @@
   transition: border 0.25s ease-in-out;
   will-change: border-color;
   background-clip: padding-box;
+  width: 100%;
```
kamilmateusiak, line 14: I feel that is a right thing to do because everyone pass width: 100% there anyway but we would need to use the same value for i.e textarea to make it consistent

PR #250, opened 2021-06-07
Handle inline label right node

kamilmateusiak:
As bonus - increase width to 100% of textarea and input elements. It's an breaking change but I believe it's needed

The same thing was developed in https://github.com/livechat/design-system/pull/246. Due to some bugs there I want to propose a different approach


dprzybylek (approved):
ship it 🚢

jdrzejb (approved):

mreszke (approved):

PR #257, opened 2021-08-06
DS-147 - New design tokens

marcinsawicki:
New design tokens interface. Updatet legacy theme with new interface and colors. Light theme currently is copy-paste from legacy to fit the interface. We will implement the correct colors after providing them by the designers.


przemuh (approved):

packages/themes/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@livechat/design-system-themes",
-  "version": "1.0.0",
+  "version": "1.1.0",
```
kamilmateusiak, line 3: I think that the version should be changed to `2.0.0-beta.0`
- `2.0.0` because this is a breaking change and if some one has this package installed with `^` before version it will be updated for him
- `beta` because the colors were not selected yet for `light` theme

kamilmateusiak (approved):
Nice contribution 💯 I am glad that you already decided what tokens should be used in theme

marcinsawicki:
Outdated

PR #262, opened 2021-09-10
DS-152 - Banner component

marcinsawicki:


packages/react-components/package.json
```diff
@@ -38,7 +38,8 @@
     "microbundle": "^0.13.3",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
-    "typescript": "^4.4.2"
+    "typescript": "^4.4.2",
+    "react-material-icon-svg": "1.7.0"
```
przemuh, line 42: You need also to update and commit `package-lock.json` ;)

packages/react-components/src/components/BannerIcon.tsx
```diff
@@ -0,0 +1,28 @@
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
```
przemuh, line 2: We don't need to use PropTypes anymore, we have TS :)

packages/react-components/src/components/Toast.tsx
```diff
@@ -0,0 +1,89 @@
...
+    baseClass,
+    {
+      [`${baseClass}--${variant}`]: variant,
+    },
+    className
+  );
+
+  const onActionClick = (action: IToastActionProps) => {
+    if (action && action.closeOnClick && onClose) {
+      action.handler;
```
przemuh, line 53: Did you forget about `()` ?

packages/react-components/src/components/Toast.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+  return (
+    <div className={mergedClassNames} id={id}>
+      <div className={`${baseClass}__icon`}>
+        <ToastIcon variant={variant} />
+      </div>
+      <div className={`${baseClass}__content`}>{children}</div>
+      {(action || removable) && (
+        <div className={`${baseClass}__actions`}>
+          {action && action.label && action.handler && (
```
przemuh, line 68: Do we need to check it if we've defined type for `IToastActionProps` ?

packages/react-components/src/components/Toast.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const mergedClassNames = cx(
+    baseClass,
+    {
+      [`${baseClass}--${variant}`]: variant,
+    },
+    className
+  );
+
+  const onActionClick = (action: IToastActionProps) => {
+    if (action && action.closeOnClick && onClose) {
```
przemuh, line 52: Wouldn't check for `action` be enough? I mean we defined `IToastActionProps` ...

packages/react-components/src/components/Toast.tsx
```diff
@@ -0,0 +1,89 @@
...
+    return action.handler;
+  };
+
+  return (
+    <div className={mergedClassNames} id={id}>
+      <div className={`${baseClass}__icon`}>
+        <ToastIcon variant={variant} />
+      </div>
+      <div className={`${baseClass}__content`}>{children}</div>
+      {(action || removable) && (
```
przemuh, line 66: Can I (as a user) use both? I mean - can I pass `action` and `removable`? Shouldn't it be exchangeable?

packages/react-components/src/stories/Banner.stories.tsx
```diff
@@ -0,0 +1,35 @@
...
+      </>
+    </BannerComponent>
+  </div>
+);
+
+Banner.args = {
+  type: BannerType.Info,
+  size: BannerSize.Small,
+  onClose: () => {
+    alert('onClose click');
```
przemuh, line 33: Instead of `alert` I would go with native storybook `actions` https://storybook.js.org/docs/react/essentials/actions. What do you think?

packages/react-components/src/stories/Promo.stories.tsx
```diff
@@ -0,0 +1,38 @@
...
+} from '../components/Promo';
+
+export default {
+  title: 'Components/Promo',
+  component: PromoComponent,
+} as ComponentMeta<typeof PromoComponent>;
+
+type IPromoArgs = IPromoProps;
+
+export const Promo = (args: IPromoArgs): React.ReactElement => (
```
przemuh, line 17: Instead of exporting one story I would add more of them to show different variants. It will be helpful for future visual testing.

We can go with one "default" story (for the playground). And variants with:

- small with link
- medium with button
- large with link
- large with button

grouping story with all variants (to display them in docs)
What do you think?

packages/react-components/src/stories/Promo.stories.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+Promo.args = {
+  buttonText: null,
+  header: 'This example headline has 40 characters',
+  img: 'https://via.placeholder.com/100',
+  light: false,
+  linkText: false,
+  size: PromoSize.Small,
+  onClose: () => {
+    alert('onClose click');
```
przemuh, line 36: The same here, I would go with `actions` instead of alerting.

packages/react-components/src/stories/Toast.stories.tsx
```diff
@@ -0,0 +1,36 @@
...
+export default {
+  title: 'Components/Toast',
+  component: ToastComponent,
+} as ComponentMeta<typeof ToastComponent>;
+
+interface IToastArgs extends IToastProps {
+  label: string;
+}
+
+export const Toast = (args: IToastArgs): React.ReactElement => (
```
przemuh, line 19: I would also add here more stories/variants to keep the visual-test coverage.

packages/react-components/src/stories/ToastWrapper.stories.tsx
```diff
@@ -0,0 +1,49 @@
...
+  ToastWrapper as ToastWrapperComponent,
+} from '../components/ToastWrapper';
+import { Variants } from '../components/Toast';
+
+export default {
+  title: 'Components/Toast Wrapper',
+  component: ToastWrapperComponent,
+} as ComponentMeta<typeof ToastWrapperComponent>;
+
+export const ToastWrapper = (args: IToastWrapperProps): React.ReactElement => (
```
przemuh, line 15: Now I am just thinking that maybe we should rename this component. Instead of a generic `wrapper` name maybe some sort of "positioner" would be a better name. 

I am also wondering if this story should be placed under `components`. as far as I remember we discussed that `patterns` would be a better place.

packages/react-components/src/components/ToastWrapper.tsx
```diff
@@ -0,0 +1,85 @@
...
+            }}
+            timeout={ANIMATION_TIME}
+          >
+            <Toast
+              id={id}
+              variant={variant}
+              onClose={onClose}
+              removable={removable}
+              action={action}
+              className="toast__single"
```
przemuh, line 76: What about `lc` prefix?

packages/react-components/src/components/ToastWrapper.tsx
```diff
@@ -0,0 +1,85 @@
...
+interface IToastsProps extends IToastProps {
+  content: React.ReactElement;
+}
+
+export interface IToastWrapperProps {
+  className: string;
+  toasts: IToastsProps[];
+  fixed: boolean;
+  block: boolean;
+  animationType: string;
```
przemuh, line 31: Shouldn't we narrow it down to the supported list of animations?

packages/styles/src/components/_banner.scss
```diff
@@ -0,0 +1,89 @@
...
+        max-width: 640px;
+      }
+    }
+  }
+
+  &--large {
+    .lc-banner__content {
+      display: flex;
+      justify-content: flex-start;
+      width: 100%;
```
przemuh, line 44: I would extract it one level higher...and add `column` type one for `--small` variant. It will reduce the amount of the css code.

przemuh (changes requested):
Nice job 👍  I've left some comments, please tell me what do you think about them. I am open for discussion.

packages/react-components/src/stories/ToastWrapper.stories.tsx
skrabkab-design: Yup, positioning should be described in Patterns. I think positioner is much more obvious than wrapper.

skrabkab-design:
Few comments from my side:

Basics:
- Font on both Banner & Alerts seems to be black, and not using token $content-default
- Promo light - please change border color to $border-subtle
- Bold text on banners should be 600 not 700,
- Can we please add support for SVG's in banner image? It was requested here https://livechatinc.atlassian.net/browse/DS-137
- Margins for close icon should be top 12px, and right 12px - please have a look on the [design](https://www.figma.com/file/2pFu80PXO5A2tfyrAGnx91/Product-Components-[DS-2.0]?node-id=3401%3A20130) 

Additional (nice to have)
- Can we please add support header to Alert? It would be nice to have the same options in Promo & Alert.
- Can we rename Banner to Alerts?
- Can we rename Banner to Promo banner?
- Can we please change background of block with Promo Light version of banner?
- What does [this](https://tppr.me/t0v82) star mean?

packages/react-components/src/components/Toast.tsx
marcinsawicki: Here, we need to verify if the optional option `closeOnClick` is defined because this will execute the specific effect of clicking the action button (will close or no)

packages/react-components/src/components/Toast.tsx
marcinsawicki: Ofc You can. Here we are checking if there are any of these props passed, to create the div container for buttons. Later we check for the specific prop to render the correct element.

.stylelintrc.json
```diff
@@ -1,3 +1,8 @@
 {
-  "extends": ["stylelint-config-standard", "stylelint-config-sass-guidelines"]
+  "extends": ["stylelint-config-standard", "stylelint-config-sass-guidelines"],
+  "rules": {
+    "max-nesting-depth": null,
+    "selector-class-pattern": null,
```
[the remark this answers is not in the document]
marcinsawicki: [done](https://github.com/livechat/design-system/pull/262/files#diff-528119d81e5a10738ae49f5e20d333ebc37cea63657eb9069281693e6dae0ad3R5-R7)

package.json
```diff
@@ -45,5 +45,8 @@
     "stylelint-config-sass-guidelines": "^8.0.0",
     "stylelint-config-standard": "^22.0.0",
     "typescript": "^4.4.2"
+  },
+  "dependencies": {
+    "react-transition-group": "^4.4.2"
```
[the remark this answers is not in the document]
marcinsawicki: [done](https://github.com/livechat/design-system/pull/262/files#diff-6adad24655c6ad668b62d38efb24e18aa0dbb3c97e4243fc83d43338bb96e6ecR58)

packages/react-components/src/components/Banner.tsx
```diff
@@ -0,0 +1,64 @@
...
+    className
+  );
+
+  return (
+    <div className={mergedClassNames}>
+      <div className={`${baseClass}__content`}>
+        <BannerIcon type={type} />
+        <div className={`${baseClass}__content-text`}>{children}</div>
+      </div>
+      {onClose && (
```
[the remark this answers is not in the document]
marcinsawicki: [unit tests ready](https://github.com/livechat/design-system/pull/262/files#diff-935d901f32d1c13e65516cfe7d8b89aaed842e09cba4b6f3cc959a5a14c51101R78-R85)

packages/react-components/src/components/BannerIcon.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+const baseClass = 'banner__icon';
+
+interface IBannerIconProps {
+  type: BannerType;
+}
+
+export const BannerIcon: React.FC<IBannerIconProps> = ({ type }) => {
+  switch (type) {
+    case 'info':
```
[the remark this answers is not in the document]
marcinsawicki: Will be done in the future: https://livechatinc.atlassian.net/browse/DS-178

packages/react-components/src/components/Toast.tsx
marcinsawicki: [done](https://github.com/livechat/design-system/pull/262/files#diff-a656d3f3036b52add490eb747e708ce34fbc6a1478c86369b71e883eea03dda8R51-R55)

packages/react-components/src/components/Toast.tsx
marcinsawicki: [done](https://github.com/livechat/design-system/pull/262/files#diff-a656d3f3036b52add490eb747e708ce34fbc6a1478c86369b71e883eea03dda8R66)

packages/react-components/src/stories/Promo.stories.tsx
marcinsawicki: Promo stories are updated

packages/react-components/src/stories/Toast.stories.tsx
marcinsawicki: Toast stories are updated

packages/react-components/src/components/ToastWrapper.tsx
marcinsawicki: [done](https://github.com/livechat/design-system/pull/262/files#diff-5d9a6fd6a4447e4a0840b4ec77197b909f3b3566570d97e034489b9f5b9516e3R76)

packages/styles/src/components/_banner.scss
marcinsawicki: Let's handle this during the refactoring process and focus on migration only.

packages/react-components/src/components/Alert.spec.tsx
```diff
@@ -0,0 +1,86 @@
+import * as React from 'react';
+import { render, fireEvent } from '../test-utils';
+import { Alert, AlertSize, AlertType } from './Alert';
+
+const customClass = 'my-css-class';
+const bannerContent = 'This example content';
+const mockedFunction = jest.fn();
+
+describe('<Banner> component', () => {
```
przemuh, line 9: I think we test here `Alert` instead of `Banner` ;) but correct me if I am wrong.

packages/react-components/src/components/Alert.spec.tsx
```diff
@@ -0,0 +1,86 @@
+import * as React from 'react';
+import { render, fireEvent } from '../test-utils';
+import { Alert, AlertSize, AlertType } from './Alert';
+
+const customClass = 'my-css-class';
+const bannerContent = 'This example content';
+const mockedFunction = jest.fn();
```
przemuh, line 7: 1. `mockedFunction` is used only in one test, so I don't see a reason to create the variable here
1. The same with `customClass`, it's used only in one test so please move declaration closer to test.
1. `bannerContent` and default render function could be extracted. This is a good practice in component testing. By extracting I mean, instead of calling `render(<Alert` everytime in tests you can create function `renderComponent` and pass default props like banner content.

Let me know what do you think.

packages/react-components/src/components/AlertIcon.tsx
```diff
@@ -0,0 +1,64 @@
...
+        />
+      );
+    default:
+      return (
+        <InformationOutlineIcon
+          fill="#4384f5"
+          width="20px"
+          height="20px"
+          className={baseClass}
+        />
```
przemuh, line 61: This is a copy paste of `case: info`. Can we extract it?

Another question: Do we even need the separate component for `AlertIcon`? Maybe a iconConfig would be enough?

packages/react-components/src/components/PromoBanner.spec.tsx
```diff
@@ -0,0 +1,131 @@
...
+  });
+
+  it('should display button', () => {
+    const { getByTestId } = render(
+      <PromoBanner header={promoHeader} buttonText={text}>
+        {promoContent}
+      </PromoBanner>
+    );
+
+    expect(getByTestId('button')).toBeTruthy();
```
przemuh, line 30: `button` is a very generic name for data-testid. Especially in DesignSystem library. Like I said in previous comment - I would rather change it to different query.

packages/react-components/src/components/PromoBanner.spec.tsx
```diff
@@ -0,0 +1,131 @@
...
+  });
+
+  it('should display link', () => {
+    const { getByTestId } = render(
+      <PromoBanner header={promoHeader} linkText={text}>
+        {promoContent}
+      </PromoBanner>
+    );
+
+    expect(getByTestId('link')).toBeTruthy();
```
przemuh, line 40: The same with link. I would check if `linkText` is rendered under button (which also makes me confused - why is it called a `linkText`?

packages/react-components/src/components/Toast.spec.tsx
```diff
@@ -0,0 +1,113 @@
...
+  });
+
+  it('should call action function', () => {
+    const { getByTestId } = render(
+      <Toast action={toastAction}>{toastContent}</Toast>
+    );
+
+    fireEvent.click(getByTestId('actionButton'));
+    expect(mockedFunction).toHaveBeenCalled();
+  });
```
przemuh, line 76: I would combine these two cases

packages/react-components/src/components/Toast.spec.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+    expect(container.firstChild).toHaveClass('lc-toast--notification');
+  });
+
+  it('should render with action button', () => {
+    const { getByTestId } = render(
+      <Toast action={toastAction}>{toastContent}</Toast>
+    );
+
+    expect(getByTestId('actionButton')).toBeTruthy();
```
przemuh, line 66: It would be better to use `.toBeVisible` from jest-dom. It's more explicit and checks more things.

packages/react-components/src/components/AlertIcon.tsx
marcinsawicki: As I mentioned before, I just want to move that component for now. We can refactor that code when the Icon component will be ready.

packages/react-components/src/stories/PromoBanner.stories.tsx
```diff
@@ -0,0 +1,210 @@
...
+  args: IPromoBannerArgs
+): React.ReactElement => (
+  <div>
+    <PromoBannerComponent {...args}>
+      <>
+        A description with a <b>maximum of 100 characters</b>. That usually
+        means only one or two sentences.
+      </>
+    </PromoBannerComponent>
+  </div>
```
przemuh, line 76: The same here. I think this is a perfect place to use Template pattern.

PR #267, opened 2021-09-24
DS-167 first version of the icons package

MichalPaszowski:
Material design icons.


packages/react-components/package.json
```diff
@@ -34,6 +34,7 @@
     "@babel/preset-env": "^7.15.6",
     "@babel/preset-react": "^7.14.5",
     "@babel/preset-typescript": "^7.15.0",
+    "@livechat/design-system-material-icons": "^1.0.0",
```
przemuh, line 37: I would move it into `dependencies` because DS components depend on this package.

Similar to how it was done in "old" DS. The old "DS" has dependency on `material-ui-icons` ;)

przemuh:
Good job! :clap:

Let's try to figure out some details here:

1. Why the package name is `material-icons`? Will we add another package for other icon types? I think we can rename package folder it to `icons` and package name to `@livechat/design-system-icons`.
2. Following the convention :point_up: I would rename `icons` folder to `src` and move material icons under `src/material`. This will be helpful in the future when we will add our custom lc icons.

What do you think?

packages/react-components/package.json
MichalPaszowski: fixed

MichalPaszowski:
> Good job! 👏
> 
> Let's try to figure out some details here:
> 
> 1. Why the package name is `material-icons`? Will we add another package for other icon types? I think we can rename package folder it to `icons` and package name to `@livechat/design-system-icons`.
> 2. Following the convention ☝️ I would rename `icons` folder to `src` and move material icons under `src/material`. This will be helpful in the future when we will add our custom lc icons.
> 
> What do you think?

I have made the changes that you have mentioned.

Where should be put icons that are lazy loaded? Shall I create a separate directory for them?

packages/icons/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.ts",
+  "scripts": {
+    "build": "svgr --typescript -d ./dist ./src ",
+    "dev": "npm run build"
+  },
+  "author": "",
+  "files": [
+    "src"
```
przemuh, line 12: Are you sure 🤔 ? What do we want to expose? `src` or `dist`? :)

MichalPaszowski: my bad

MichalPaszowski:
I have moved icons to src/material and during the build process, they are moved to dist.
As for the license used by google, it is Apache 2.0. Seems to be fairly permissive: https://github.com/google/material-design-icons/blob/master/LICENSE

packages/icons/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "@livechat/design-system-icons",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.ts",
+  "scripts": {
+    "build": "svgr --typescript -d ./dist ./src/material ",
```
przemuh, line 7: What do you think about putting the whole `src` folder as an input?

packages/icons/src/material/account_circle.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
norbertprocak, line 1: So what happened to the "no dimensions inside svg" convention? Is it not applicable to design system?

packages/icons/package.json
MichalPaszowski: that would create a "material" directory inside "dist", do we want that?

packages/react-components/src/stories/foundations/components/IconsShowcase.tsx
```diff
@@ -0,0 +1,19 @@
+import * as React from 'react';
+
+import * as Icons from '@livechat/design-system-icons/dist/material';
+import { Icon, IconSizeName } from '../../../components/Icon';
+import './iconsShowcase.css';
+
+export const IconsShowcase = () => {
+  const iconsGrid = Object.keys(Icons).map((item) => {
+    const iconSource = (Icons as any)[item];
```
MichalPaszowski, line 9: Can someone suggest how to avoid dreadful 'any'?

VadymBezpalko (approved):

PR #269, opened 2021-10-15
moving switch to storybook

dprzybylek:
Moved switch to the storybook, removed snapshot, adjusted styles. 
image


dprzybylek:
> focus state is different than what we have on Button component, and also a bit different from what we have in Figma https://tppr.me/tqphO Can we try to fix that? @przemuh can give a reference for a focus state

Done ✅ Also I have added a comment - todo for shadows tokens

> small improvement, when switch is in compact variant then there is a different padding on the white knob https://tppr.me/wv3x1 can we try to align that, so bottom, top and side paddings are the same?

Cannot reproduce, seems legit in the storybook 😢 Do you use `cmd + +` or storybook zoom? Pls help!

> slider component should have this token, because no matter the theme it will remain white https://tppr.me/VhFO7

Done ✅ 

> I'm not sure if all tokens are correct, because it's a bit hard for me to understand how this is build in code. In Figma switch has property Enabled, which can be True or False, and also has states which can be Enabled, Hover, Disabled, Focus, Locked. I think this makes more sense than calling checked switch Enabled or Disabled, because both of those variants have states. Please have a look in Figma and it will be much more obvious what I mean, and I don't assume we'll change everything here, but maybe we could improve something here. Here is a cool example from Orbit by kiwi https://tppr.me/99Tzo they call this Checked (true/false) we also use the same pattern for checkbox https://tppr.me/2x6Nt and radio button https://tppr.me/W5uov

Imho we should leave it as it is. There is no pattern for naming here - `default toggled/checked/selected` these are the examples from other design systems

packages/react-components/src/components/Switch.spec.tsx
```diff
@@ -0,0 +1,15 @@
+import { fireEvent } from '@testing-library/dom';
+import * as React from 'react';
+import { render } from '../test-utils';
+import { Switch } from './Switch';
+
+describe('Switch', () => {
+  const toggleFn = jest.fn();
+
+  it('should call on toggle when clicked', () => {
```
walaszczykm, line 9: It would be nice to also test the proper behavior of the `Switch` component in terms of changing its states when the user interacts with it:
- When it is **enabled** -> user clicks it -> expect to be **disabled** 
- When it is **disabled** -> user clicks it -> expect to be **enabled**

Additionally, it would also be beneficial to test the duality of the Switch component nature: 
- Controlled input (changes in property `on` are directly reflected in the component visual and logical state)
- Uncontrolled input (maintain its internal state on its own and only notifies about changes)

packages/react-components/src/components/Switch.tsx
```diff
@@ -0,0 +1,86 @@
...
+  disabled = false,
+  name = baseClass,
+  on,
+  onChange = noop,
+  size = SwitchSize.Basic,
+  innerRef,
+  ...props
+}) => {
+  const getEnabledValue = () => (on !== undefined ? on : defaultOn);
+  const [enabled, setEnabled] = React.useState(getEnabledValue());
```
walaszczykm, line 42: The original implementation of `Switch` component contains logic that allows this component to be both: controlled or uncontrolled input making the state being derived from props 👇 

https://github.com/livechat/design-system/blob/0a1035019e20cb84e38d5a03267def2fa9228ee3/packages/design-system/src/components/Switch/Switch.js#L18

Here the controlled input part is missing as there is no synchronization between `on` property and `enabled` state. 
It can be easily added using `useEffect` hook 👇 
```ts
React.useEffect(() => {
    if (on !== undefined) {
      setEnabled(on);
    }
  }, [on]);
```

dprzybylek: Added, and I have decided to remove `defaultOn` property. Using `on` and `onChange` will cause the same effect

package-lock.json
```diff
@@ -1,8 +1,9747 @@
 {
   "name": "@livechat/ds-workspace",
   "version": "1.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
```
walaszczykm, line 4: You have used a newer version of npm while installing dependencies which caused the switch from lock version 1 to 2. As there are no changes in dependencies in this PR it would be easiest to just omit the changes to this file completely.

dprzybylek: In fact, I shouldn't have any changes in package-lock. Nothing new added, I will checkout the `v1` version

PR #270, opened 2021-10-15
DS-153 Card component

walaszczykm:
Migrate `Card` componenet to the new setup.
Jira link: https://livechatinc.atlassian.net/browse/DS-153

- Move code and styles
- Add types
- Create new tests using testing-library
- Add storybook story

### DIFF
--- a/packages/styles/src/components/_card.scss
+++ b/packages/styles/src/components/_card.scss
@@ -1,6 +1,6 @@
 .lc-card {
   background-color: var(--surface-basic-default);
-  border: solid 1px var(--border-default);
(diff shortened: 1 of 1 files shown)


walaszczykm:
@skrabkab-design Thanks for the review. I have adjusted the Card componenet according to your comments. In terms of `font-size` and `line-height` of texts I have followed the Figma design board where it is specified that Card title should use **Heading/S** and Card content should use **Paragraph/Medium.**

skrabkab-design:
Looks awesome! I'm only missing typo tokens, but as I understood this will be added.

PR #272, opened 2021-10-18
DS-170: RadioButton

VadymBezpalko:
image


przemuh (approved):

packages/react-components/src/stories/RadioButton.stories.tsx
```diff
@@ -0,0 +1,29 @@
...
+    <RadioButtonComponent {...args}>{children}</RadioButtonComponent>
+  </div>
+);
+
+RadioButton.args = {
+  checked: false,
+  disabled: false,
+  description: 'Lorem ipsum',
+  children: 'Radio button label',
+};
```
walaszczykm, line 29: by casting this object literal to a props type you can gain additional type safety for actually setting those initial args values
```suggestion
} as IRadioButtonProps;
```

packages/react-components/src/components/RadioButton.tsx
```diff
@@ -0,0 +1,50 @@
+import * as React from 'react';
+import cx from 'classnames';
+import { FieldDescription } from './FieldDescription';
+
+export interface IRadioButtonProps
+  extends React.HTMLAttributes<HTMLInputElement> {
+  description?: string;
```
walaszczykm, line 7: in the current implementation, the `description` prop has the type `React.ReactNode` not `string` which means that other valid react nodes (like the nested structure of paragraphs, spans, etc.) are allowed to be passed.

https://github.com/livechat/design-system/blob/0a1035019e20cb84e38d5a03267def2fa9228ee3/packages/design-system/src/interfaces/forms.d.ts#L161

PR #273, opened 2021-10-19
DS-169 - Text field, input field and numeric input

marcinsawicki:
https://livechatinc.atlassian.net/browse/DS-169
- Migrate the `NumericInput`, `NumericInputField`, `InputField`, `TextField`
- Added Storybook stories
- Added unit tests


packages/styles/src/components/_numeric-input.scss
```diff
@@ -0,0 +1,149 @@
...
+  &--error {
+    input {
+      &,
+      &:hover,
+      &:focus {
+        border-color: var(--color-negative-default);
+      }
+    }
+  }
+  /* stylelint-enable no-descending-specificity */
```
marcinsawicki, line 148: I've disabled this rule here to make a commit and put the task to review. I have difficulties with solving the `no-descending-specificity` issue because when I put that part higher, the default styles of component will rewrite those from the modifier.

packages/react-components/src/components/NumericInput.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+  const mergedStyle = style ? { ...style } : void 0;
+
+  React.useEffect(() => {
+    window.addEventListener('keydown', onKeyDown);
+
+    return () => {
+      window.removeEventListener('keydown', onKeyDown);
+    };
+  }, [value]);
```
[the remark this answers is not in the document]
walaszczykm: You can use functional state update version `updateValue(prevValue => prevValue + 1)` which also you to grab the current state value (at the call time) as a function argument 🙂

packages/react-components/src/components/NumericInput.tsx
```diff
@@ -0,0 +1,162 @@
...
+      [`${baseClass}--no-controls`]: noControls,
+      [`${baseClass}--disabled`]: disabled,
+    },
+    className
+  );
+
+  const handleOnFocus = () =>
+    window.addEventListener('keydown', onKeyDown, true);
+
+  const handleonBlur = () => window.removeEventListener('keydown', onKeyDown);
```
norbertprocak, line 55:
```suggestion
  const handleOnBlur = () => window.removeEventListener('keydown', onKeyDown);
```

norbertprocak (approved):

PR #276, opened 2021-10-20
DS-157: Tab component

walaszczykm:
- Migrate `Tab` and `TabsWrapper` to a new setup.
- Add new tests for both components
- Create interactive storybook stories

### DIFF
--- a/packages/styles/src/components/_tab.scss
+++ b/packages/styles/src/components/_tab.scss
@@ -9,14 +9,13 @@
   margin: 0 8px 11px;
   outline: none;
   overflow: visible;
-  padding: 0 12px;
+  padding: 0 12px 1px;
   position: relative;
   transition: color 0.25s;
 
   &__description {
     color: var(--content-subtle);
     display: inline-block;
-    opacity: 0.6;
     padding-left: 3px;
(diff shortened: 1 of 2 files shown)


skrabkab-design:
Nice one! I've got few thoughts here:
- We're aiming to remove the visual difference between tab name and number of items it contains, because the counter is already separated from the text, by putting it in (). I see they both use the same tokens, but the number seems to be lighter, it's probably because of applied opacity (0.6) - we can easily get rid of that. Can we please align this to what we have on the designs? [here](https://www.figma.com/file/2pFu80PXO5A2tfyrAGnx91/Product-Components?node-id=2043%3A14704) 
image
- I can't find an info in Inspect what color the divider is using, but can you please confirm if it's `--border-subtle` ?https://tppr.me/7I11q
- Probably we're moving current implementation, but if we could align this component to what we have in Figma it would be awesome, it's a matter of moving text 1px higher and decreasing the size of rectangle in selected item. Here is an overview https://tppr.me/f7hTo - I don't think any of those changes would cause regression, and we could gain better alignment between Figma & Code.

marcinsawicki (approved):

PR #283, opened 2021-10-28
DS-186 ButtonGroup component

przemuh:
- Added a ButtonGroup component and style it as in the FIGMA designs.
- Fixed one unit test warning about Form -> input without `onChange`
- Added userEvent to test-utils
- Fixed button height by removing line-height

### DIFF
--- a/packages/react-components/src/components/Form.spec.tsx
+++ b/packages/react-components/src/components/Form.spec.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 
 import { render } from '../test-utils';
+import noop from '../utils/noop';
 import { Form } from './Form';
 
 describe('<Form /> component', () => {
@@ -26,7 +27,7 @@ describe('<Form /> component', () => {
   it("should render nested elements as 'children'", () => {
     const { getByRole } = render(
       <Form>
-        <input type="text" value="Hello" onChange={() => null} />
+        <input type="text" value="Hello" onChange={noop} />
       </Form>
     );
 
--- a/packages/react-components/src/components/Form.spec.tsx
+++ b/packages/react-components/src/components/Form.spec.tsx
@@ -26,7 +26,7 @@ describe('<Form /> component', () => {
   it("should render nested elements as 'children'", () => {
     const { getByRole } = render(
       <Form>
-        <input type="text" value="Hello" />
+        <input type="text" value="Hello" onChange={() => null} />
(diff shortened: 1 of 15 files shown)


PawelRebes (approved):

skrabkab-design:
Looks awesome! 

Few questions:
- I don't see button group component, will this be separate PR?
-  For loading buttons I don't see an option to manage size. How does it work? 
image
- I know it's small thing, but I don't want to give people wrong impression – in the UI we use only first capital letter, in the example we have label Button Text - while it should be Button text.
image
- Do we plan to add Link button, it will be similar to text button but without paddings – this will be really helpful. We've discussed this on DM on Slack, but I guess this got missed along the way. Can we please add this? When we add that we can mark Text button as deprecated.

przemuh:
> * I don't see button group component, will this be separate PR?

Screenshot 2021-11-12 at 08 51 02

> * For loading buttons I don't see an option to manage size. How does it work?

This PR is not about loading button story :) Stories will be updated in separate PR. I have an idea to provide one customizable story per logical variant + one story with all possible variants to track potential visual changes.

> * I know it's small thing, but I don't want to give people wrong impression – in the UI we use only first capital letter, in the example we have label Button Text - while it should be Button text.

Sure thing, I will update it in next PR.

> * Do we plan to add Link button, it will be similar to text button but without paddings – this will be really helpful. We've discussed this on DM on Slack, but I guess this got missed along the way. Can we please add this? When we add that we can mark Text button as deprecated.

Link Button will be introduced in separate PR.

packages/react-components/src/components/Form.spec.tsx
```diff
@@ -26,7 +26,7 @@ describe('<Form /> component', () => {
   it("should render nested elements as 'children'", () => {
     const { getByRole } = render(
       <Form>
-        <input type="text" value="Hello" />
+        <input type="text" value="Hello" onChange={() => null} />
```
VadymBezpalko, line 29: noop? :)

VadymBezpalko (approved):

skrabkab-design:
One thing that i've noticed – we don't show the separators between segments in Segmented button. @przemuh Can we try to adjust it, so that it looks more consistent with the design? https://www.figma.com/file/2pFu80PXO5A2tfyrAGnx91/Product-Components?node-id=1654%3A839

PR #286, opened 2021-11-12
Colors palette + adjustments in build styles

przemuh:
- Added enum with design token names
- Adjustments in build process (styles package now generates also JS with Tokens enum)
- Foundation stories are now sorted by priority
- Implemented foundation story about Design Tokens

image


VadymBezpalko (approved):

marcinsawicki (approved):

skrabkab-design:
This is awesome release! 

- I think we should rename this Page to Color tokens, as Design tokens represent larger piece - color, typo, shadow, spacing etc. 
- Can we separate the tokens in following order: backround, surface, content, border, accent? I feel this would bring much more clarity to the concept. We have this in Figma, and it helps to outline the structure https://www.figma.com/file/8Ye4yfokmY0mkogOM2rMzW/Legacy-Theme-Tokens?node-id=491%3A5923
- It would be awesome to separate those tokens with some headline & short description, as we have in Figma.
- It would be great if we could align the text & color a bit better 
image
 
- Can we use a bit smoother style for the rectangles? I think using --border-subtle & corner radius (4px) would be great.
- I think we could also fix the color in token name, we could assign --content-default to it

PR #290, opened 2021-11-16
DS-00-grouping-input (for AA purposes)

VadymBezpalko:
image


marcinsawicki (approved):

packages/react-components/src/components/TagInput/EditableTagContent.tsx
```diff
@@ -0,0 +1,86 @@
+import * as React from 'react';
+import { KeyCodes } from '../../constants/keyCodes';
+import cx from 'classnames';
+import { escape } from 'lodash';
+
+export interface IEditableTagContentProps {
+  value: string;
+  className: string;
```
przemuh, line 8: Shouldn't className be optional?

packages/react-components/src/constants/keyCodes.ts
```diff
@@ -1,3 +1,12 @@
 export const KeyCodes = {
   esc: 'Escape',
+  enter: 'Enter',
+  backspace: 'Backspace',
+  delete: 'Delete',
+  del: 'Del', // for IE 9-11, FF 36 and earlier
```
przemuh, line 6: I guess we don't support IE and old versions of FF.

przemuh (commented):
I don't see any tests for added components. Could you add at least a few of them to illustrate the basic usage and provide proof that it works correctly? :)

packages/react-components/src/components/TagInput/EditableTagContent.tsx
VadymBezpalko: yep, it should :D

packages/react-components/src/constants/keyCodes.ts
VadymBezpalko: ok, I'll remove it then 👌

packages/react-components/src/components/TagInput/EditableTagContent.tsx
```diff
@@ -0,0 +1,86 @@
...
+      focusInputRef();
+      return;
+    }
+
+    const value = getValue();
+    if (e.key === KeyCodes.backspace && value === '') {
+      setRemoved(true);
+      remove();
+      focusInputRef();
+      return;
```
mreszke, line 53: do we need this return here?

packages/react-components/src/components/TagInput/EditableTagContent.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const onBlur = () => {
+    const ref = getRef();
+    if (!removed && ref) {
+      if (ref.innerText === '') {
+        remove();
+        return;
+      }
+
+      change(ref.innerText);
+    }
```
mreszke, line 66: wdyt?
```suggestion
    if (removed || !ref) {
        return;
    }

    if (ref.innerText === '') {
        remove();
        return;
    }

    change(ref.innerText);
```

packages/react-components/src/components/TagInput/Tag.tsx
```diff
@@ -0,0 +1,62 @@
...
+  className = '',
+  children,
+  index,
+  remove,
+  validator,
+  inputRef,
+  update,
+  ...restProps
+}) => {
+  const valid = React.useMemo(() => {
```
mreszke, line 29:
```suggestion
  const isValid = React.useMemo(() => {
```

mreszke: anyway, do You think it's worth to memoize it?

VadymBezpalko: @mreszke I made it memoized because of the validator function, it will be passed from component props and can be basically anything, so I thought it would be better to memorize it 😄

mreszke: @przemuh wdyt?

VadymBezpalko:
> I don't see any tests for added components. Could you add at least a few of them to illustrate the basic usage and provide proof that it works correctly? :)

Sure, I've added some tests to the components 👌

mreszke (approved):

packages/react-components/src/components/TagInput/Tag.tsx
przemuh: Discussed offline :)

przemuh (approved):

PR #291, opened 2021-12-03
DS-172 - Picker component

marcinsawicki:
https://livechatinc.atlassian.net/browse/DS-172

**UPDATE:**
After discussion with @skrabkab-design and @przemuh we decided to not migrate the old `Select` component, and prepare the new and shiny `Picker` component with new, simpler interface and functionality.


packages/react-components/src/components/Select/Select.tsx
```diff
@@ -0,0 +1,371 @@
...
+    if (getIsOpen()) {
+      return onBodyOpen();
+    }
+
+    if (!getIsOpen()) {
+      onBodyClose();
+    }
+  }, [getIsOpen]);
+
+  const filterItem = (item: any): any => {
```
przemuh, line 131: Are you sure that this function returns `any`? My guess is `boolean`. And it would be good to set typing for `item` also.

packages/react-components/src/components/Select/Select.tsx
```diff
@@ -0,0 +1,371 @@
...
+    if (e.key === EventKeys.ArrowDown || e.key === EventKeys.ArrowUp) {
+      e.preventDefault();
+      showSelectBody();
+    }
+  };
+
+  const onSelectHeadFocus = () => {
+    setIsFocused(true);
+
+    if (!isOpen) {
```
przemuh, line 223: Why in some cases we check `isOpen` which is an internal state, and sometimes we check `getIsOpened` ?

packages/react-components/src/components/Select/SelectList.tsx
```diff
@@ -0,0 +1,202 @@
...
+  onFocusedItemChange,
+}) => {
+  const hoverCallbacks: IHoverCallbacksProps = {};
+  let timerId: ReturnType<typeof setTimeout> | null = null;
+
+  React.useEffect(() => {
+    if (isOpen) {
+      if (listRef.current) {
+        listRef.current.scrollTop = 0;
+      }
```
przemuh, line 48:
```suggestion
        listRef?.current.scrollTop = 0;
```

przemuh (changes requested):
I've left some comments. I think we should re-think the API of the component. IMHO it might be simpler.

Potential bug: After searching/filtering without results, pressing ENTER selects the first element.

https://user-images.githubusercontent.com/3865478/144674457-4b9448c6-3b51-492c-8d98-5097aa43f45f.mp4

Moreover, I think there is something wrong with colors:

In DarkMode elements are not visible:

image

In other themes the hover/selected item has low contrast ratio IMHO:

image

Please check it with @skrabkab-design

packages/react-components/src/components/Select/SelectList.tsx
marcinsawicki: I didn't change that because:
`The left-hand side of an assignment expression may not be an optional property access.ts(2779)`

skrabkab-design:
Hey 👋 few things that I've noticed:
- image
We should use different color for hovered items `--surface-basic-hover`
- image
Selected item (text & check icon) should be blue --color-action-default
- image
Placeholder text should be the same size as default text that appears in the select. 
- image the placeholder text should use typography token - `paragraph-medium`
- image disabled item should use appropriate tokens - border-disabled & content-disabled - now we're doing it wrong by reducing the opacity of content default & border default,

skrabkab-design:
image
I can see there is a way to add props description to Storybook. I feel in this component this is a must, as without explanation those props seem to be real nightmare. @marcinsawicki Can you please try to add them, and copy description from DS Git repository?


FYI: @przemuh

skrabkab-design:
This is how the component should look like in Dark Mode 
image and this is how it looks in Storybook 
image

This is because select component doesn't have background applied both – select & dropdown should use --surface-basic-default token.

marcinsawicki:
> Details

Sure. As we discussed, I'll handle it in the stories PR.

PawelRebes (approved):

skrabkab-design:
1. Top & bottom paddings are different - please keep 4px padding 
image
2. Can we make first letter a capital one? This helps readability, and as an example, we should show it in a proper way
image
3. We have small differences in text position - the difference appears when enabling / disabling search option in this component 
image
4. Can we please adjust the title font style, to match the newest DS version in Figma? 🙏 https://www.figma.com/file/2pFu80PXO5A2tfyrAGnx91/Product-Components?node-id=2329%3A20904

przemuh (commented):
I've left some comments.

In general I still think that this component is unnecessary complicated. We should start from simple list and delegate as much as we can upper way. It would be good to not force the itemModel shape. We could use something like `encode/decode` with default `entity` behavior.

The dropdown list and select in terms of lists looks very the same. I would treat the Select component like a Form-Component. And follow a basic strategy of implementing form inputs - it means `onChange` instead of `onSelect`, `options` instead of `items` and `value` instead `selected`. `searchable` boolean flag instead of the verb `search`. As I said in code-comments - we could provide simple possibility of filtering but we should delegate it to the end user for more complex functions.

I know that you put a lot of effort in this component, but please try to take a step back and think about the simplicity over complexity. We don't need to map 1:1 the API, we should improve it. For me improve means make it as simple as possible :)

And please rememeber that not every select needs to be a controlled component. Let's give the end user possibility to read the value from the ref or whatever place he wants.

Remember that only one element could be focused :) we don't need to keep the focusedItemKey to know which element is currently focused. `document.activeElement` will tell you this :)

If you have any questions/doubts don't hesitate to ping me on slack.

packages/styles/src/components/_picker-list.scss
```diff
@@ -0,0 +1,34 @@
...
+  font-size: 15px;
+  font-weight: 400;
+  height: 224px;
+  line-height: 22px;
+  overflow: scroll;
+  padding: 4px 0;
+  position: absolute;
+  width: 341px;
+
+  &_item {
```
przemuh, line 18: If we follow the BEM (and I think we do) then we should use double underscore `__`

packages/styles/src/components/_trigger.scss
```diff
@@ -0,0 +1,38 @@
+$base-class: 'lc-trigger';
```
przemuh, line 1: `Trigger` is too much generic name for me. That's why I've suggested `lc-picker-trigger`.

przemuh (changes requested):
I like the way we changed direction. It looks way better than the previous one. But there are still some things that need to be re-think. 

I've left some comments - please let me know when something is not clear.

Good job 👏

packages/react-components/src/stories/Picker.stories.tsx
```diff
@@ -0,0 +1,38 @@
...
+  title: 'Components/Picker',
+  component: PickerComponent,
+  parameters: {
+    componentSubtitle: `TBD`,
+  },
+} as ComponentMeta<typeof PickerComponent>;
+
+const StoryTemplate: Story<IPickerProps> = (args: IPickerProps) => {
+  return (
+    <div style={{ width: 320 }}>
```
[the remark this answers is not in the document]
marcinsawicki: Ok, my bad because it should be `height`. It's to make more space for dropdown in Storybook docs, to see the list.

packages/react-components/src/components/Picker/Picker.tsx
```diff
@@ -0,0 +1,53 @@
...
+    setIsListOpen(false);
+    setSelectedItem(null);
+  };
+
+  return (
+    <div className={baseClass}>
+      <Trigger
+        selectedItemText={selectedItem?.name}
+        size={size}
+        triggerText={triggerText}
```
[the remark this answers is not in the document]
marcinsawicki: Yea, I like that idea, but still need to pass that any item is selected to render the clear button.

packages/react-components/src/components/Picker/Trigger.tsx
```diff
@@ -0,0 +1,63 @@
...
+import cx from 'classnames';
+import {
+  ChevronDown,
+  Close,
+} from '@livechat/design-system-icons/dist/material';
+import { Icon, IconSizeName, IconTypeName } from '../Icon';
+
+const baseClass = 'lc-trigger';
+
+const DEFAULT_TRIGGER_TEXT = 'Choose an option';
```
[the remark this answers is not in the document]
marcinsawicki: Good point, I think we can force the user of that component to put the value for default trigger text.

packages/react-components/src/components/Picker/PickerList.tsx
```diff
@@ -0,0 +1,60 @@
...
+import cx from 'classnames';
+import { Check } from '@livechat/design-system-icons/dist/material';
+import { Icon, IconTypeName } from '../Icon';
+
+const baseClass = 'lc-picker-list';
+
+export interface IPickerListItem {
+  key: string;
+  name: string;
+}
```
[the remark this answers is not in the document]
marcinsawicki: We will handle it in the later iteration

skrabkab-design:
image
When component is disabled there should be no option to clear it - X icon should not be visible

image
There is some weird color around text, this should not be visible

When user starts to search, closes the picker and opens it again the search should be cleared – so that all options are visible https://www.loom.com/share/9b262eeaae3943cb8d1cb0d0ccae68ca

image
Please remove the border around list

How the consumer would be able to remove the clear icon from select? Is there any props that decides wether or not the X icon appears?

packages/react-components/src/components/Picker/PickerList.tsx
```diff
@@ -32,17 +32,16 @@ export const PickerList: React.FC<IPickerListProps> = ({
   onClose,
   onSelect,
 }) => {
-  const mergedClassNames = cx(
-    baseClass,
-    `${baseClass}--${size}`
-  );
+  /* eslint-disable */
+  const mergedClassNames = cx(baseClass, `${baseClass}--${size}`);
+  /* eslint-enable */
```
marcinsawicki, line 37: I've disabled it because I get:
`35:59  error  Invalid type "any" of template literal expression  @typescript-eslint/restrict-template-expressions`

I don't know why, because we have the same implementation in `Trigger.tsx`

packages/react-components/src/components/Picker/Picker.tsx
```diff
@@ -0,0 +1,134 @@
...
+  const triggerRef = React.useRef<HTMLDivElement>(null);
+
+  React.useEffect(() => {
+    if (!isListOpen) {
+      setSearchPhrase(null);
+      return document.removeEventListener('click', onDocumentClick);
+    }
+
+    return document.addEventListener('click', onDocumentClick);
+  }, [isListOpen]);
```
przemuh, line 40: I think this is a potential memory leak. Plus returning value from useEffect should be a "clean-up" function.

My proposal:

```
React.useEffect(() => {
   const onDocumentClick = (e) => e.target.contains(triggerRef.current) && isListOpen && setIsListOpen(false)
   
   document.addEventListener('click', onDocumentClick);
   
   return () => document.removeEventListener('click', onDocumentClick)
} [isListOpen])
```

Moreover, @MichalPaszowski had an idea to implement a custom hook `onDocumentClick`, because this pattern is used in a lot of places.

Please double check this potential memory leak.

packages/react-components/src/components/Picker/Picker.tsx
```diff
@@ -0,0 +1,134 @@
...
+  const handleOnTriggerClick = () => {
+    if (disabled) {
+      return;
+    }
+
+    if (isListOpen) {
+      return setIsListOpen(false);
+    }
+
+    return setIsListOpen(true);
```
przemuh, line 60: `setIsListOpen(prev => !prev)`

packages/react-components/src/components/Picker/PickerList.tsx
```diff
@@ -0,0 +1,206 @@
...
+  items,
+  selectedItem,
+  size = TriggerSize.Medium,
+  onClose,
+  onSelect,
+}) => {
+  /* eslint-disable */
+  const mergedClassNames = cx(
+    baseClass,
+    `${baseClass}--${size}`,
```
przemuh, line 38:
```suggestion
    `${baseClass}--${size as string}`,
```

packages/react-components/src/components/Picker/PickerList.tsx
```diff
@@ -0,0 +1,206 @@
...
+    items.length === 0 && `${baseClass}__no-results`
+  );
+  /* eslint-enable */
```
przemuh, line 41:
```suggestion
  const mergedClassNames = cx(
    baseClass,
    `${baseClass}--${size as string}`,
    items.length === 0 && `${baseClass}__no-results`
  );
```

packages/react-components/src/components/Picker/PickerList.tsx
```diff
@@ -0,0 +1,204 @@
...
+  items,
+  selectedItem,
+  size = TriggerSize.Medium,
+  onClose,
+  onSelect,
+}) => {
+  const mergedClassNames = cx(
+    baseClass,
+    `${baseClass}--${size as string}`,
+    items.length === 0 && `${baseClass}__no-results`
```
MichalPaszowski, line 38: {
     [`${baseClass}__no-results`]: items.length === 0 
    }

packages/react-components/src/components/Picker/PickerList.tsx
```diff
@@ -0,0 +1,198 @@
...
+              {item.name}
+            </li>
+          );
+        }
+
+        return (
+          <li
+            id={item.key}
+            key={item.key}
+            className={cx(
```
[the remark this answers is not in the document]
MichalPaszowski: and I would make use of
    {
     class: condition
    }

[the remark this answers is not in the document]
marcinsawicki: > I would move that class merging stuff up

Are You sure? This is a list item that is mapped from the array. I don't feel that will give us profit in the case of creating const for every list item to just merge the classes if I understand that suggestion correctly. Can You give some more details, about that suggested change?

packages/react-components/src/components/Picker/PickerList.module.scss
```diff
@@ -0,0 +1,72 @@
+$base-class: 'picker-list';
```
MichalPaszowski, line 1: Have we abandoned 'lc-something' convention?

MichalPaszowski (approved):

packages/react-components/src/components/Picker/Picker.tsx
```diff
@@ -0,0 +1,140 @@
...
+  const handleOnSelect = (item: IPickerListItem) => {
+    setIsListOpen(false);
+    setSelectedItem(item);
+    onSelect(item);
+  };
+
+  const handleOnClearClick = () => {
+    setIsListOpen(false);
+    setSelectedItem(null);
+    onSelect(null);
```
przemuh, line 73: What do you think about adding test cases for calling `onSelect` for different scenarios?

I mean:

```
it("should call onSelect with selected item")
it("should call onSelect with null after clearing the selection")
```

?

packages/react-components/src/components/Picker/PickerList.module.scss
przemuh: `lc-` prefix will be added in build step.

You can find the configuration for this here:

```
//vite.config.js

  css: {
    modules: {
      generateScopedName: 'lc-[name]__[local]___[hash:base64:5]',
    },
  },
```

packages/react-components/src/components/Picker/PickerList.spec.tsx
```diff
@@ -0,0 +1,94 @@
...
+  onClose: () => noop,
+  onSelect: () => noop,
+};
+
+const renderComponent = (props: IPickerListProps) => {
+  return render(<PickerList {...props} />);
+};
+
+describe('<PickerList> component', () => {
+  it('shouldnt render PickerList if not isOpen', () => {
```
przemuh, line 31:
```suggestion
  it('shouldn't render PickerList if not isOpen', () => {
```

or

```suggestion
  it('should not render PickerList if not isOpen', () => {
```

packages/react-components/src/components/Picker/PickerList.spec.tsx
```diff
@@ -0,0 +1,94 @@
...
+  it('should call onSelect when list item clicked', () => {
+    const mockedFunction = vi.fn();
+    const { getByText } = renderComponent({
+      ...defaultProps,
+      isOpen: true,
+      onSelect: mockedFunction,
+    });
+
+    fireEvent.click(getByText('Option three'));
+    expect(mockedFunction).toHaveBeenCalledWith({
```
przemuh, line 55: It is way better to express the expected meaning of the test. Here the goal is to check id `onSelect` has been called. So I would change the naming a little bit. WDYT?

```suggestion
    const onSelect = vi.fn();
    const { getByText } = renderComponent({
      ...defaultProps,
      isOpen: true,
      onSelect,
    });

    fireEvent.click(getByText('Option three'));
    expect(onSelect).toHaveBeenCalledWith({
```

packages/react-components/src/components/Picker/PickerList.spec.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+  it('should mark selected list item as selected', () => {
+    const { getByText } = renderComponent({
+      ...defaultProps,
+      isOpen: true,
+      selectedItem: { key: 'three', name: 'Option three' },
+    });
+
+    expect(getByText('Option three')).toHaveClass(
+      styles[`${itemClassName}--focused`]
```
przemuh, line 69: I have a problem with the className here. Why is it `focused`? Shouldn't it be `--selected` or something similar?

packages/react-components/src/components/Picker/PickerList.spec.tsx
```diff
@@ -0,0 +1,94 @@
...
+      isOpen: true,
+      items: [{ key: 'three', name: 'Option three', disabled: true }],
+    });
+
+    expect(getByText('Option three')).toHaveClass(
+      styles[`${itemClassName}--disabled`]
+    );
+  });
+
+  it('should display empty state if no filter result', () => {
```
przemuh, line 85: Do we have plans to extend the API that allows providing a custom "empty state" message?

packages/react-components/src/components/Picker/Trigger.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { Icon, IconSizeName, IconTypeName } from '../Icon';
+import styles from './Trigger.module.scss';
+
+const baseClass = 'picker-trigger';
+
+export const enum TriggerSize {
+  Compact = 'compact',
+  Medium = 'medium',
+  Large = 'large',
+}
```
przemuh, line 17: I would change it to:

```
type TriggerSize = 'compact' | 'medium' | 'large'
```

Using enums have some disadvantages...and with plain string union we simplify the API. Instead of importing enum we could write something like:

```
<Trigger size="compact" />
```

instead of

```
import { TriggerSize, Trigger } ...
<Triger size={TriggerSize.Compact} />
```

And using string union will fix the problem of casting:

```
styles[`${baseClass}--${size as string}`],
```

WDYT?

packages/react-components/src/components/Picker/PickerList.tsx
```diff
@@ -0,0 +1,207 @@
...
+    return () => document.removeEventListener('keydown', onKeyDown);
+  }, []);
+
+  const scrollItems = () => {
+    if (!listRef.current) {
+      return;
+    }
+
+    const focusedElement = listRef.current.querySelector(
+      styles[`.${itemClassName}--hovered`]
```
przemuh, line 63: Does it work? Because I have doubts :) 

Shouldn't it be something like:

```suggestion
      `.${styles[`{itemClassName}--hovered`]}`
```

packages/react-components/src/components/Picker/PickerList.module.scss
```diff
@@ -0,0 +1,72 @@
...
+    cursor: pointer;
+    display: flex;
+    height: 36px;
+    justify-content: space-between;
+    padding: 7px 12px;
+
+    &--hovered,
+    &--focused {
+      background-color: var(--surface-basic-hover);
+    }
```
przemuh, line 48: Why cannot we use `:hover` and `:focus` ?

packages/react-components/src/components/Picker/Trigger.spec.tsx
```diff
@@ -0,0 +1,114 @@
...
+      ...defaultProps,
+      onClick: mockedFunction,
+    });
+
+    fireEvent.click(getByText('Example text'));
+    expect(mockedFunction).toHaveBeenCalled();
+  });
+
+  it('should call onClearClick when clear button clicked', () => {
+    const mockedFunction = vi.fn();
```
przemuh, line 82: The same here. I mean - the naming convention.

przemuh:
Navigation by keys doesn't scroll the list and the focus management is broken

https://user-images.githubusercontent.com/3865478/162406325-3bf8ea44-7047-4666-8261-37340151a6a2.mov

przemuh:
Please checkout the w3 wai-aria examples: https://www.w3.org/TR/wai-aria-practices/examples/combobox/combobox-select-only.html

There is a list of supported keys and working example of the "picker"

packages/react-components/src/components/Picker/Picker.tsx
marcinsawicki: 👍 done

packages/react-components/src/components/Picker/Trigger.tsx
marcinsawicki: 👍

packages/react-components/src/components/Picker/Picker.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+    return options.filter((item: IPickerListItem) => {
+      const search = searchPhrase.toLowerCase();
+      const itemName = item.name.toLowerCase();
+
+      if (itemName.includes(search) && !item.groupHeader) {
+        return item;
+      }
+    });
+  };
```
walaszczykm, line 110: there is more than `searchPhrase` tha can change in this component during its lifecycle, which means `getOptions`Twold be called multiple times unnecessary. Considering that it is doing list filtering, it may be a performance issue for larger lists so it's a good candidate for being a derived and memorized value that changes only when `searchPhrase` has changed. 

Additionally, [`Array.prototype.filter`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter) takes a predicated as its argument. A predicate is a function that should always return a boolean value. Here you are returning a specific item element. It works because the item object is cast to boolean type and it is true but it is unnecessary and hard to read. In case of any predicate you can return condition value directly 🙂 

```suggestion
  const items = React.useMemo<IPickerListItem[]>(() => {
    if (!searchPhrase) {
      return options;
    }

    return options.filter((item) => {
      if(item.groupHeader) {
        return false
      }

      const search = searchPhrase.toLowerCase();
      const itemName = item.name.toLowerCase();
      return itemName.includes(search)
    });
  }, [searchPhrase])
```

packages/react-components/src/components/Picker/Trigger.tsx
```diff
@@ -0,0 +1,110 @@
...
+    styles[`${baseClass}--${size}`],
+    isDisabled && styles[`${baseClass}--disabled`],
+    isOpen && styles[`${baseClass}--focused`],
+    isError && styles[`${baseClass}--error`]
+  );
+
+  React.useEffect(() => {
+    const onKeyDown = (e: KeyboardEvent) => {
+      const el = document.getElementById('picker-trigger');
+      const isFocused = document.activeElement === el;
```
walaszczykm, line 51: Instead of quering an element from DOM by it's id, the handler can use a React's ref instead
```suggestion
      const isFocused = document.activeElement === pickerTriggerRef.current;
```

packages/react-components/src/components/Picker/Trigger.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+      if (isFocused && e.key !== KeyCodes.tab) {
+        return onClick();
```
walaszczykm, line 54:
```suggestion
        onClick();
```

packages/react-components/src/components/Picker/Trigger.tsx
```diff
@@ -0,0 +1,110 @@
...
+      }
+    };
+
+    document.addEventListener('keydown', onKeyDown);
+
+    return () => document.removeEventListener('keydown', onKeyDown);
+  }, []);
+
+  const handleTriggerClick = () => {
+    return onClick();
```
walaszczykm, line 64:
```suggestion
   onClick();
```

packages/react-components/src/components/Picker/PickerList.tsx
```diff
@@ -0,0 +1,219 @@
...
+      const relativeTop = itemTop + itemHeigth - ulTop;
+      const itemOfsetTop = focusedElement.offsetTop;
+
+      if (relativeTop > ulHeight) {
+        listRef.current.scrollTop = itemOfsetTop - ulHeight + itemHeigth;
+      } else if (itemTop < ulTop) {
+        listRef.current.scrollTop = itemOfsetTop - (itemOfsetTop % itemHeigth);
+      }
+    }
+  };
```
walaszczykm, line 92: This function and all of those calculations can be replaced using [`scrollIntoView`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) on the currently active element.
We can use the `ref` prop from "host" elements which can receive a function that on each render would get the reference to the very specfifc DOM node of this element. Combining this with `currentItemKey` state value we can decide on which element to call `scrollIntoView`. That way we will receive exactly the same behavior and user experience without doing all calculations by hand 🙂 

It will look something like this:
```jsx
<li
  ref={element => {
    if(currentItemKey === item.key) {
      element?.scrollIntoView({ block: 'nearest' })
    }
  }}
>
{item.name}
</li>
```

```suggestion
```

packages/react-components/src/components/Picker/PickerList.tsx
```diff
@@ -0,0 +1,219 @@
...
+      indexRef.current = getNextItemIndex();
+
+      setCurrentItemKey(items[indexRef.current].key);
+      scrollItems();
+    }
+
+    if (e.key === KeyCodes.enter && !items[indexRef.current].disabled) {
+      e.preventDefault();
+
+      return onSelect(items[indexRef.current]);
```
walaszczykm, line 167:
```suggestion
     onSelect(items[indexRef.current]);
```

packages/react-components/src/components/Picker/PickerList.tsx
```diff
@@ -0,0 +1,219 @@
...
+
+    lastIndexRef.current = indexRef.current;
+    return indexRef.current;
+  };
+
+  const onKeyDown = (e: KeyboardEvent) => {
+    if (e.key === KeyCodes.esc) {
+      e.preventDefault();
+
+      return onClose();
```
walaszczykm, line 143:
```suggestion
      onClose();
```

packages/react-components/src/components/Picker/PickerList.tsx
```diff
@@ -0,0 +1,219 @@
...
+
+    if (items.length > 0) {
+      setCurrentItemKey(items[indexRef.current].key);
+    }
+
+    if (isOpen) {
+      document.addEventListener('keydown', onKeyDown);
+    }
+
+    return () => document.removeEventListener('keydown', onKeyDown);
```
walaszczykm, line 64: If we register the event listener conditionally, then we should also return the cleanup function for its unregistration also conditionally 😉 

```suggestion
    if (isOpen) {
      document.addEventListener('keydown', onKeyDown);
      return () => document.removeEventListener('keydown', onKeyDown);
    }
```

walaszczykm (approved):

PR #293, opened 2021-12-06
AA-12778: Tag component

VadymBezpalko:


packages/react-components/src/components/TagInput/TagInput.tsx
```diff
@@ -99,29 +100,38 @@ export const TagInput: React.FC<ITagInputProps> = ({
...
+          onKeyDown={onInputKeyDown}
+          onPaste={onPaste}
+        />
+      </div>
+      {error && (
+        <span className="lc-field-error">
+          <Text as="span" size="sm">
+            {error}
+          </Text>
+        </span>
```
przemuh, line 133: Why not just `<Text className="lc-field-error" as="span" size="sm">{error}</Text>` ? :)

packages/react-components/src/stories/Tag.stories.tsx
```diff
@@ -0,0 +1,125 @@
...
+import { Tag as TagComponent, ITagProps } from '../components/Tag';
+
+export default {
+  title: 'Components/Tag',
+  component: TagComponent,
+} as ComponentMeta<typeof TagComponent>;
+
+export const Tag = ({ children, ...args }: ITagProps): React.ReactElement => {
+  return (
+    <div className="spacer" style={{ display: 'flex' }}>
```
przemuh, line 14: AFAIR `spacer` class is helpful when you have many components under the div like:

```
<div className="spacer">
   <Button />
    <Button />
     <Button />
</div>
```

Do we need it here? Remember that it is going to be a part of "See code" panel.

packages/react-components/src/stories/Tag.stories.tsx
```diff
@@ -0,0 +1,125 @@
...
+        </TagComponent>
+        <TagComponent {...args} kind="warning" outline>
+          {children}
+        </TagComponent>
+        <TagComponent {...args} kind="success" outline>
+          {children}
+        </TagComponent>
+        <TagComponent {...args} kind="error" outline>
+          {children}
+        </TagComponent>
```
przemuh, line 66: I would extract it to separate story `kindsWithOutline`

packages/react-components/src/stories/Tag.stories.tsx
```diff
@@ -0,0 +1,125 @@
...
+        <TagComponent {...args} kind="warning" icon={Smiles}>
+          {children}
+        </TagComponent>
+        <TagComponent {...args} kind="success" icon={Smiles}>
+          {children}
+        </TagComponent>
+        <TagComponent {...args} kind="error" icon={Smiles}>
+          {children}
+        </TagComponent>
+      </div>
```
przemuh, line 84: Separate story: `kindsWithIcon`

packages/react-components/src/stories/Tag.stories.tsx
```diff
@@ -0,0 +1,125 @@
...
+        <TagComponent {...args} kind="warning" avatar={avatar}>
+          {children}
+        </TagComponent>
+        <TagComponent {...args} kind="success" avatar={avatar}>
+          {children}
+        </TagComponent>
+        <TagComponent {...args} kind="error" avatar={avatar}>
+          {children}
+        </TagComponent>
+      </div>
```
przemuh, line 101: Story: `kindsWithAvatar`

packages/react-components/src/stories/Tag.stories.tsx
```diff
@@ -0,0 +1,125 @@
...
+        <TagComponent {...args} kind="warning" icon={Smiles} dismissible>
+          {children}
+        </TagComponent>
+        <TagComponent {...args} kind="success" icon={Smiles} dismissible>
+          {children}
+        </TagComponent>
+        <TagComponent {...args} kind="error" icon={Smiles} dismissible>
+          {children}
+        </TagComponent>
+      </div>
```
przemuh, line 118: `dismissable` is just an option that could be configured in the first story. I think we don't need to list all kinds only for showing `X` icon.

przemuh (approved):
I've left some very minor comments .Good job 👏

packages/react-components/src/components/TagInput/TagInput.tsx
VadymBezpalko: @przemuh Copy/pasted from TextField component 😄 
But after applying className directly on Text component I found strange case -> the size of Text is changing - `lc-field-error ` sets font-size (13px) and Text component is setting its own font-size (14px) which has higher priority than the parent class. Checked on [Figma](https://www.figma.com/file/2pFu80PXO5A2tfyrAGnx91/Product-Components?node-id=4046%3A33393) and it should have font-size 14px and typography tokens as well, so I fixed it in TextField component as well.

But in the end, I found out that we have some duplicates in the codebase(we already have FieldError and FieldDescription components which have typography tokens) so I fixed it a little bit 😄

skrabkab-design:
I've checked the branch! Looks great!

MichalPaszowski (approved):

PR #295, opened 2022-01-20
Link + Button as a Link & Link as a Button 🤪

przemuh:
**This PR introduces:**

- one new component `<Link>` which displays a simple `<a>` tag,
- new kind "link" for `Button` component, which displays as a link,
- new prop `href` which can be passed to the `Button` component and makes it an `<a>`

**Bonus:**

- pin node v16 with `volta`
- upgrade Storybook to the latest-stable version


VadymBezpalko (approved):
🚀

skrabkab-design:
1. Wouldn't it make more sense to move [Link](https://613a8e945a5665003a05113b-nfwsbsnbvn.chromatic.com/?path=/story/components-link--link) to typography styles? It's pretty hard to understand the difference between Link in Button vs. link as a separate story https://tppr.me/UPF1E
2. I don't see props for button link - I believe we should show href field here, because otherwise it's not obvious what is the function of this component. Kiwi uses this [similarly](https://tppr.me/xlXGB) 
3. Let's change the button type here to Primary. We shouldn't show pattern of using descructive button here :D 
image
3. In disabled state Link button still reacts on hover 
image

I've collected some thoughts regarding the navigation in [Figma](https://www.figma.com/file/2pFu80PXO5A2tfyrAGnx91/Product-Components?node-id=5235%3A39844) Please have a look and let me know what do you think. 

Some general thought after looking again on button component – do you think we need a separate story for Loading state? I think once we add Pseudo states plugin we could simplify the menu structure. Loading state is not crucial, and I think we could skip this in navigation and only show this through props, same as we show another state - Disabled.

packages/react-components/src/components/Link.tsx
```diff
@@ -0,0 +1,31 @@
...
+ * Simple component which renders an `<a>` element.
+ *
+ * Place the `<Link>` inside a `<Text>` component or pass a custom className to set the typography styles.
+ *
+ * Use `<Button kind="link">` to act as a `<button>` - <a target="_self" href="/story/components-button--kinds">see the story</a>.
+ *
+ * Use `<Button href="">` to display as a `<Button>` - <a target="_self" href="/story/components-button--link">see the story</a>.
+ */
+export const Link: React.FC<LinkProps> = ({
+  bold = false,
```
norbertprocak, line 21: Was there any discussion inside the team about style-like props? For example `bold` is a prop that could be replaced with a proper `className`. Do we have a policy to avoid styling by props? And if we want to keep `bold` prop because it's "commonly used", then why not `fontSize` or `color` ?

packages/react-components/src/stories/Button.stories.tsx
```diff
@@ -4,12 +4,12 @@ import { ComponentMeta } from '@storybook/react';
...
 
 import { Icon } from '../components/Icon';
-import { Button, Props as ButtonProps } from '../components/Button';
+import { Button, ButtonProps } from '../components/Button';
 
 const icons = Object.fromEntries(
   Object.entries(MaterialIcons).map(([key, source]) => [
     key,
-    <Icon source={source} />,
+    <Icon source={source as 'string'} />,
```
norbertprocak, line 12:
```suggestion
    <Icon source={source as string} />,
```
?

norbertprocak (approved):

packages/react-components/src/components/Link.tsx
przemuh: The question is - do you want to force user to remember "proper" className or provide the prop `bold` which in fact adds another className (hides implementation details). 

This is some "kind" of Link. On the other hand, I'm wondering if we shouldn't get rid of it at all and provide text-decorations from `<Text>` component.

I would love to know the opinion of other devs from Desgin System @marcinsawicki @MichalPaszowski

przemuh:
> 1. Wouldn't it make more sense to move Link to typography styles? It's pretty hard to understand the difference between Link in Button vs. link as a separate story.

From UX perspective I can agree that it's hard to understand the difference. But from Dev perspective I think it is more than clear. I think we should make it clear that styles for link could be used only for links :) If we would move link styles to typo tokens only then we give the possibility to use them everywhere. Here we make a clear statement - this is a link, looks like a link, and should work as a link.


> 2. I don't see props for button link - I believe we should show href field here, because otherwise it's not obvious what is the function of this component. Kiwi uses this [similarly](https://tppr.me/xlXGB)

`href` is a prop that comes from `React.AnchorHTMLAttributes<HTMLAnchorElement>`. We can provide some text information in Button docs that it takes Button & Anchor attributes, and if you want to make your button a link then pass the `href`. I would love to see other devs opinion @marcinsawicki @MichalPaszowski 

I've added `href` prop but only for the story. In the main DOC table `href` is not present because this prop is inherited from AnchorProps. Is it ok?

> 3. Let's change the button type here to Primary. We shouldn't show pattern of using descructive button here :D
>    image

✅  Done - I've change it to primary kind

> 4. In disabled state Link button still reacts on hover
>    image

✅  Done + fixes for ButtonLink when button is disabled (link was active, not it is not)

> I've collected some thoughts regarding the navigation in [Figma](https://www.figma.com/file/2pFu80PXO5A2tfyrAGnx91/Product-Components?node-id=5235%3A39844) Please have a look and let me know what do you think.

One comment here - I would leave the states in kinds or provide another story 'states' - it will help us in Visual Regression verification on Chromatic. Without states in stories we cannot verify if button displays correctly in different states.

> Some general thought after looking again on button component – do you think we need a separate story for Loading state? I think once we add Pseudo states plugin we could simplify the menu structure. Loading state is not crucial, and I think we could skip this in navigation and only show this through props, same as we show another state - Disabled.

The same as above. We can remove the dedicated 'Loading' story and present it in a table under "states" story - it will help us in Visuall Regression verification.

skrabkab-design:
@przemuh Thanks for all the improvements! 🙌 

Few more things from my side:

- I think it's worth to create a separate story for States, so they would not be confused with Kinds – this story would contain Disabled & Loading
- In the Loading story we have 5 variants, instead of 6 - looks like the variant for link button is missing 
image
- Button component with icon is larger than without the icon, to avoid that we should reduce the paddings for the variant with icon
image
- I'm thinking about this name – Link button - it's pretty easy to confuse Text & Link button. I've done some research and I really like the way Shopify names this type of button – Plain button, wich explains the structure, and it's pretty self explanatory name. What do you think about changing the approach here, so that the name of this branch could be less confusing 😆 
- I don't see how we can adjust the text style of link button - we would need to have and option to use both Bold & Regular text style here
- Regarding @norbertprocak [comment](https://github.com/livechat/design-system/pull/295#discussion_r794314463) i think it's good to have some more styling options here, we sometimes use Link button (Plain?) with --content-default color (here is an [example](https://tppr.me/Q2M3O) ) it's not really popular pattern, but sometimes it works better than blue one, so I believe it could be useful.
- I've made some [draft](https://www.figma.com/file/2pFu80PXO5A2tfyrAGnx91/Product-Components?node-id=5477%3A39839) on how I would see the structure of the documentation, it's not a guideline on how we should do it, just my perspective - let's talk about it and think what works best.

skrabkab-design:
One more think related to buttons, not specifically to this branch, but I would leave it here, as I believe it would be easier to handle all button cases at once. 

Please have a look on this page in Figma, i've double checked the button group component, and sizes are a bit different there https://www.figma.com/file/2pFu80PXO5A2tfyrAGnx91/Product-Components?node-id=5477%3A39820

skrabkab-design:
One more thing - please keep the order of button sizes from the smallest to largest - compact, medium (default), large - this should be kept consistent across all components.

packages/react-components/src/components/Link.tsx
MichalPaszowski: IMHO, decorators from the Text component will fit most of the use cases and make sense.

MichalPaszowski:
> `href` is a prop that comes from `React.AnchorHTMLAttributes<HTMLAnchorElement>`. We can provide some text information in Button docs that it takes Button & Anchor attributes, and if you want to make your button a link then pass the `href`. I would love to see other devs opinion @marcinsawicki @MichalPaszowski

I am OK with not listing this property explicitly.

MichalPaszowski (approved):

packages/react-components/src/components/Link.tsx
marcinsawicki: Agree with @MichalPaszowski

marcinsawicki (approved):

PR #298, opened 2022-02-11
Popover

MichalPaszowski:
A popover is a component that is supposed to replace Dropdown.
Components : Popover - Popover ⋅ Storybook 2022-03-01 08-32-09


packages/react-components/src/stories/Popover.stories.tsx
```diff
@@ -0,0 +1,117 @@
...
+            height: '308px',
+          }}
+        >
+          <div
+            style={{
+              display: 'flex',
+              justifyContent: 'center',
+              alignItems: 'center',
+              border: '1px dashed var(--border-default)',
+              backgroundColor: 'var(--surface-basic-default)',
```
MichalPaszowski, line 88: it seems that this style does not work

packages/react-components/src/components/Popover.tsx
```diff
@@ -0,0 +1,131 @@
...
+import * as PopperCore from '@popperjs/core';
+import cx from 'classnames';
+
+type PopoverModifiers = {
+  [key: string]: Partial<PopperCore.StrictModifiers>;
+};
+export interface IPopoverProps {
+  children?: React.ReactNode;
+  className?: string;
+  closeOnEscPress?: boolean;
```
przemuh, line 12: I would check with WCAG if we shouldn't always close on ESC. If someone wants to block it, then it could be done by `event.stopPropagation` or something similar.

packages/react-components/src/components/Popover.tsx
```diff
@@ -0,0 +1,131 @@
...
+    popperRef.current,
+    {
+      modifiers: props.modifiers ? buildPopperModifiers(props.modifiers) : [],
+      placement: placement,
+    }
+  );
+
+  const mergedClassNames = cx(
+    `${className ? className : ''} popover ${visible ? 'popover__visible' : ''}`
+  );
```
przemuh, line 109:
```suggestion
  const mergedClassNames = cx('popover', {
     [className]: className,
     "popover--visible": visible
     });
```

packages/styles/src/components/_popover.scss
```diff
@@ -0,0 +1,18 @@
...
+  box-shadow: 0 1px 10px 0 rgba(66, 77, 87, 0.3);
+  display: none;
+  max-width: 336px;
+  min-width: 168px;
+
+  &:focus {
+    outline: none;
+  }
+
+  &__visible {
```
przemuh, line 15: `visible` is a some kind of state, not a nested element, so please use the modifier `--`.

packages/styles/src/components/_popover.scss
```diff
@@ -0,0 +1,18 @@
+$base-class: 'popover';
```
marcinsawicki, line 1: Shouldn't be `lc-popover`?

marcinsawicki (approved):

packages/styles/src/components/_popover.scss
MichalPaszowski: it should, thanks for pointing that out

packages/styles/src/components/_popover.scss
MichalPaszowski: 👌🏼

packages/react-components/src/components/Popover.tsx
```diff
@@ -0,0 +1,131 @@
...
+        {...attributes.popper}
+      >
+        {children}
+        <div
+          data-popper-placement={
+            attributes.popper
+              ? attributes.popper['data-popper-placement']
+              : 'bottom'
+          }
+        />
```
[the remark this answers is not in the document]
MichalPaszowski: we don't

packages/react-components/src/components/Popover.tsx
MichalPaszowski: 👌🏼

packages/react-components/src/components/Popover.tsx
```diff
@@ -0,0 +1,131 @@
...
+        name: 'preventOverlow',
+        options: {
+          rootBoundary: 'viewport',
+          ...(preventOverflow || {}),
+        },
+      },
+    ];
+
+    return [...calculatedOffset, ...calculateFlip, ...calculatePreventOverflow];
+  };
```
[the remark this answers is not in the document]
MichalPaszowski: 👌🏼

packages/react-components/src/components/Popover.tsx
MichalPaszowski: WCAG agrees with you. Reference https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html
👌🏼

PR #299, opened 2022-03-04
Datepicker

kamilmateusiak:
- Make sure to use appropriate typography tokens (see other components for reference – like Card or Radio button) ✅ 
- Make sure to use color tokens - [Figma](https://www.figma.com/file/2pFu80PXO5A2tfyrAGnx91/Product-Components?node-id=4478%3A40684) ✅ 
- Make sure to use appropriate icons from the DS icon package ✅ 
- Introduce small visual tweaks [Figma](https://www.figma.com/file/2pFu80PXO5A2tfyrAGnx91/Product-Components?node-id=4478%3A40684) ✅ 
- Remove option to enter custom date manually in this type of date picker https://tppr.me/dww7X using inputs inside select looks pretty weird, we're also not using this pattern in AA, so looks like this exists only in DS ✅ 

Added only a few tests to single DatePicker for now. It would be great to have some feedback from you firstly. 

Single DatePicker:
Screenshot 2022-03-13 at 22 54 25

RangeDatePicker:
Screenshot 2022-03-13 at 22 54 34


skrabkab-design:
Hey @kamilmateusiak thank you so much for working on that component! You've done amazing work at the right moment, in this sprint we're doing the last push to ship the v1. Thank you so much for your contribution! 🙌

skrabkab-design (approved):

packages/react-components/src/stories/RangeDatePicker.stories.tsx
```diff
@@ -0,0 +1,129 @@
...
+            onChange={(e) => {
+              select.onItemSelect(e.target.value);
+            }}
+          >
+            {args.options.map((o) => (
+              <option key={o.id} value={o.id}>
+                {o.label}
+              </option>
+            ))}
+          </select>
```
kamilmateusiak, line 117: For now I used html select in story as the Picker component is in development. 
RangeDatePicker as in old DS is exposing API which can be consument by children - as for select it will be two props: `onItemSelect` and `selected`

packages/react-components/src/components/Datepicker/hooks.ts
```diff
@@ -0,0 +1,95 @@
...
+import { calculateDatePickerMonth, getInitialStateFromProps } from './helpers';
+import {
+  IRangeDatePickerProps,
+  IRangeDatePickerReducerAction,
+  IRangeDatePickerState,
+  RangeDatePickerAction,
+  RangeDatePickerReducer,
+} from './types';
+
+export const useRangeDatePickerState = (
```
kamilmateusiak, line 11: Because the component has some extended internal state I decided to use useReducer hook to handle that. The main reasons: 
- initial state is calculated from props
- choosing one date we need to control other dates as well (reset the state or as in `RangeDatePickerAction.SELECT_SECOND_DAY_AS_FROM` switch the dates

kamilmateusiak:
components were written in the Datepicker directory: 
Screenshot 2022-03-13 at 23 12 55

I wanted to have helper, sub-components and hooks outside - the DatePicker is quite complicated so IMHO it's better to separate it.

kamilmateusiak:
also I can't resolve conflicts because I don't have `write` access to repository. Please update or if you want we can move the discussion to the `internal` PR but I won't be able to make changes probably

Issue #284, opened 2021-10-29
Move Date picker to new repository [DS-165]


skrabkab-design:
We're aiming to move [this](https://livechat.github.io/design-system/#/Components/DatePicker) component to Storybook

To do:
- Make sure to use appropriate typography tokens (see other components for reference – like Card or Radio button),
- Make sure to use color tokens - [Figma](https://www.figma.com/file/2pFu80PXO5A2tfyrAGnx91/Product-Components?node-id=4478%3A40684)
- Make sure to use appropriate icons from the DS icon package,
- Introduce small visual tweaks [Figma](https://www.figma.com/file/2pFu80PXO5A2tfyrAGnx91/Product-Components?node-id=4478%3A40684)
- Remove option to enter custom date manually in this type of date picker https://tppr.me/dww7X using inputs inside select looks pretty weird, we're also not using this pattern in AA, so looks like this exists only in DS,

kamilmateusiak:
Hello. I am sorry for no info about work status for so long (I've had some personal stuff happening). Currently I have some free time and the component migration is almost done. Unfortunately I don't have access to the repository:
```
ERROR: Permission to livechat/design-system.git denied to kamilmateusiak.
fatal: Could not read from remote repository
```
If you didn't start the work on that yet and you are still interested in external contribution please add me as contributor to the repository.

sgraczyk:
Canceled in favor of #331

PR #300, opened 2022-03-09
#331 Datepicker component

przemuh:
https://github.com/livechat/design-system/pull/299/files


skrabkab-design:
From my perspective this component is perfect! There are few things that we should improve:

- There should be a corner radius 
image

- The icons that jump between years are not the right ones - we should use those icons from our DS 
image

- I don't see the option to mark today on the date picker. Is this option supported?
image

skrabkab-design:
Component looks great! Awesome work @kamilmateusiak! 
Few minor things and we're ready to go! I think we'll finish them on our side. @przemuh can you take over from here?

[](https://user-images.githubusercontent.com/71127167/157433235-b8c7ad49-bcb2-45d9-80c6-f059b5966b20.png)The icons that jump between years are not the right ones - we should use those icons from our DS
image 
This one still needs to be fixed

image
When selecting range, the corner radius should be 0px for the inner corners

image
When selecting range in Dark mode - there is some lighter border around selection. Can we do anything to remove that effect?

image
We need to add support for entering the dates manually

image
We need to add Picker component, to replace the placeholder

kamilmateusiak:
> We need to add support for entering the dates manually
In the issue (https://github.com/livechat/design-system/issues/284) requirements was stated that manual inputs change should be removed and because of that I removed that option. @skrabkab-design did you change your mind?

packages/react-components/src/components/Datepicker/helpers.ts
```diff
@@ -0,0 +1,144 @@
...
+        ...base,
+        [styles[`${baseClass}__day--start`]]: to,
+      };
+    }
+    return {
+      ...base,
+      [styles[`${baseClass}__day--start`]]: [from, to],
+      [styles[`${baseClass}__day--end`]]: [from, to],
+    };
+  }
```
mreszke, line 93: It would be possible to make these `if` statements flatter

packages/react-components/src/components/Datepicker/helpers.ts
```diff
@@ -0,0 +1,144 @@
...
+      forcePreviousMonth ||
+      (toMonth && isSameMonth(initialToDate, toMonth))
+    ) {
+      return subMonths(initialToDate, 1);
+    }
+
+    return initialToDate;
+  }
+
+  return subMonths(toMonth || new Date(), 1);
```
mreszke, line 52: same as above

packages/react-components/src/components/Datepicker/RangeDatePicker.tsx
```diff
@@ -0,0 +1,248 @@
...
+import { useRangeDatePickerState } from './hooks';
+import {
+  IRangeDatePickerChildrenPayload,
+  IRangeDatePickerOption,
+  IRangeDatePickerProps,
+  RangeDatePickerAction,
+} from './types';
+
+export const RangeDatePicker = (
+  props: IRangeDatePickerProps
```
mreszke, line 25: please spread props here
```suggestion
 { .... }: IRangeDatePickerProps
```

packages/react-components/src/components/Datepicker/DatePicker.spec.tsx
```diff
@@ -0,0 +1,39 @@
+import { within } from '@testing-library/react';
+import * as React from 'react';
+import { render, userEvent } from 'test-utils';
+import { vi } from 'vitest';
+import { DatePicker } from './DatePicker';
+import { IDatePickerProps } from './types';
+
+describe('<DatePicker> component', () => {
```
mreszke, line 8: I'm missing some date range picking tests

mreszke (commented):
There is a lot of logic inside `useEffects` on props change, and IMO it's not correct usage of effects. This logic should be fired when calling some `onChange`

packages/react-components/src/components/Datepicker/DatePicker.tsx
```diff
@@ -0,0 +1,150 @@
...
+    numberOfMonths,
+    navbarElement,
+    renderDay,
+    innerRef,
+    ...restProps
+  } = props;
+
+  const [month, setMonth] = React.useState(propsMonth || new Date());
+
+  React.useEffect(() => {
```
AlithAnar, line 38: do we need to supply `React.` namespace everywhere? you could import it directly

packages/react-components/src/components/Datepicker/DatePicker.tsx
```diff
@@ -0,0 +1,150 @@
...
+    if (propsMonth && propsMonth !== month) {
+      setMonth(propsMonth);
+    }
+  }, [propsMonth, month]);
```
AlithAnar, line 42: props.month

packages/react-components/src/components/Datepicker/DatePicker.tsx
```diff
@@ -0,0 +1,150 @@
...
+        props.onMonthChange(month);
+        return;
+      }
+
+      setMonth(month);
+    },
+    [propsMonth, props.onMonthChange]
+  );
+
+  const getDatePickerClassNames = () => ({
```
AlithAnar, line 66: could you extract this helper outside this component? it can be parametrized like:

```suggestion
  const getDatePickerClassNames = (range: Range, classNames: string[]) => ({
```

packages/react-components/src/components/Datepicker/RangeDatePicker.tsx
```diff
@@ -0,0 +1,248 @@
...
+      type: RangeDatePickerAction.CURRENT_MONTH_CHANGE,
+      payload: { date: currentMonth },
+    });
+  }, [props.toMonth, props.initialFromDate, props.initialToDate]);
+
+  // call onChange when valid dates selected (for manual options - with date picker)
+  React.useEffect(() => {
+    const { from, selectedItem, to } = state;
+
+    if (from && to) {
```
AlithAnar, line 66:
```TS
   if(!(from && to)) {
      return 
   }
```

packages/react-components/src/components/Datepicker/RangeDatePicker.tsx
```diff
@@ -0,0 +1,248 @@
...
+      const selectedOption = getSelectedOption(selectedItem, props.options);
+
+      if (selectedOption) {
```
AlithAnar, line 69: if(!selectedOption) {
   return
}

packages/react-components/src/components/Datepicker/RangeDatePicker.tsx
```diff
@@ -0,0 +1,248 @@
...
+    );
+    const selectedOption = React.useMemo(() => {
+      return getSelectedOption(selectedItem, props.options);
+    }, [props.options, selectedItem]);
+
+    const selectedDays = React.useMemo(() => {
+      return [from, { from, to: temporaryTo }];
+    }, [from, temporaryTo]);
+
+    const disabledDays = React.useMemo(() => {
```
AlithAnar, line 203: `React.` prefix makes a lot of noise here. Consider importing useMemo and other hooks directly from React

packages/react-components/src/components/Datepicker/helpers.ts
```diff
@@ -0,0 +1,144 @@
...
+  differenceInCalendarDays,
+} from 'date-fns';
+import { Modifiers } from 'react-day-picker';
+import {
+  IRangeDatePickerProps,
+  IRangeDatePickerState,
+  IRangeDatePickerOption,
+} from './types';
+
+const baseClass = 'date-picker';
```
AlithAnar, line 15:
```suggestion
const baseClass = 'lc-date-picker';
```

?

packages/react-components/src/components/Datepicker/helpers.ts
AlithAnar: ☝️

packages/react-components/src/components/Datepicker/RangeDatePicker.tsx
MichalPaszowski: I am aware, but it has become kinda standard among components in DS. If I am to change that we would have to all agree that we want to make that change across all components. Personally, I don't mind that prefix @sgraczyk, @marcinsawicki ?

packages/react-components/src/components/Datepicker/helpers.ts
MichalPaszowski: We have dropped this convention along the way.

packages/react-components/src/components/Datepicker/RangeDatePicker.tsx
MichalPaszowski: 👌🏼

packages/react-components/src/components/Datepicker/DatePicker.tsx
MichalPaszowski: 👌🏼

packages/react-components/src/components/Datepicker/RangeDatePicker.tsx
MichalPaszowski: 👌🏼

packages/react-components/src/components/Datepicker/DatePicker.spec.tsx
```diff
@@ -0,0 +1,39 @@
+import { within } from '@testing-library/react';
```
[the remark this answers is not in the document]
MichalPaszowski: 👌🏼

packages/react-components/src/components/Datepicker/DatePicker.tsx
MichalPaszowski: should I get rid of propsMonth alias altogether?

packages/react-components/src/components/Datepicker/RangeDatePicker.tsx
MichalPaszowski: 👌🏼

packages/react-components/src/components/Datepicker/DatePicker.spec.tsx
MichalPaszowski: 👌🏼

packages/react-components/src/components/Datepicker/helpers.ts
MichalPaszowski: this code has been "inherited" from the last version of DS. I have simplified the code here, and seems to be working, but maybe I have missed some use cases here.

packages/react-components/src/components/Datepicker/helpers.ts
MichalPaszowski: I don't know how I can simplify this. 
we have to return a month that will be marked as a "current" one. Cases I see here are as follows:
 - if nothing has been provided to the helper, return last month
 - if only toMonth was supplied return it as the current month
 - if initialFromDate and initialToDate are different return last month.
 - if initialFromDate is not available and toMonth is not available return initialToDate
 Can someone help me out with this?

MichalPaszowski:
> There is a lot of logic inside `useEffects` on props change, and IMO it's not correct usage of effects. This logic should be fired when calling some `onChange`

I've had a screen-sharing session with Marek, and we have concluded that changing this would require rewriting a big chunk of code. It will stay at it is now until the component is rewritten to React DayPicker v8

packages/react-components/src/components/Datepicker/helpers.ts
mreszke:
```suggestion
  if (!initialToDate) {
     return subMonths(toMonth || new Date(), 1);
  }
  
  const forcePreviousMonth =  initialFromDate && !isSameMonth(initialFromDate, initialToDate);

  if (forcePreviousMonth || (toMonth && isSameMonth(initialToDate, toMonth))) {
    return subMonths(initialToDate, 1);
  }

  return initialToDate;
```

packages/react-components/src/components/Datepicker/helpers.ts
```diff
@@ -0,0 +1,174 @@
...
+    return {
+      ...base,
+      [styles[`${baseClass}__day--end`]]: to,
+    };
+  } else if (diff < 0) {
+    return {
+      ...base,
+      [styles[`${baseClass}__day--start`]]: to,
+    };
+  }
```
mreszke, line 81: `else if` is unnecessary here
```suggestion
  if (diff > 0) {
    return {
      ...base,
      [styles[`${baseClass}__day--end`]]: to,
    };
  }
  
  if (diff < 0) {
    return {
      ...base,
      [styles[`${baseClass}__day--start`]]: to,
    };
  }
```

packages/react-components/src/components/Datepicker/helpers.ts
```diff
@@ -0,0 +1,174 @@
...
+  const isRangeSelected = from && to;
+  return Boolean(!from || isRangeSelected);
+};
+
+export const getInitialStateFromProps = (
+  props: IRangeDatePickerProps
+): Partial<IRangeDatePickerState> => {
+  const state: Partial<IRangeDatePickerState> = {};
+
+  if (props.initialSelectedItemKey) {
```
mreszke, line 107:
```suggestion
  if (!props.initialSelectedItemKey) {
     return state
  }
  
  ...
```

mreszke (approved):

packages/react-components/src/components/Datepicker/DatePicker.tsx
AlithAnar: I think yes. There is no need to create a new variable if we `props.variable` has almost the same length.

packages/react-components/src/components/Datepicker/RangeDatePicker.tsx
AlithAnar: Ok. Just bear in mind you could benefit from some optimizations using the new JSX transform in React 17 in the future https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#whats-different-in-the-new-transform 

> Depending on your setup, its compiled output may slightly improve the bundle size.

packages/react-components/src/components/Datepicker/DatePicker.tsx
sgraczyk: We can drop importing `* as React` if whenever possible

packages/react-components/src/components/Datepicker/helpers.ts
MichalPaszowski: 👌🏼

packages/react-components/src/components/Datepicker/helpers.ts
MichalPaszowski: 👌🏼

packages/react-components/src/components/Datepicker/helpers.ts
MichalPaszowski: 👌🏼

AlithAnar (approved):

packages/react-components/package.json
```diff
@@ -70,11 +70,13 @@
     "@livechat/data-utils": "^0.2.16",
     "@livechat/design-system-icons": "^1.0.0-alpha.13",
     "@popperjs/core": "^2.11.2",
+    "classnames": "^2.3.1",
```
sgraczyk, line 73: Aren't we using `clsx`?

packages/react-components/src/components/DatePicker/DatePicker.spec.tsx
```diff
@@ -0,0 +1,39 @@
+import { within } from '@testing-library/react';
+import * as React from 'react';
+import { render, userEvent } from 'test-utils';
+import { vi } from 'vitest';
+import { DatePicker } from './DatePicker';
+import { IDatePickerProps } from './types';
+
+describe('<DatePicker> component', () => {
```
sgraczyk, line 8: Can we think of a more advanced set of tests? Alternatively, we could cover it in a separate PR, however, first, we need to make sure manual tests are passed.

packages/react-components/src/components/DatePicker/hooks.ts
```diff
@@ -0,0 +1,95 @@
+import * as React from 'react';
```
sgraczyk, line 1: Import only selected elements

packages/react-components/src/components/DatePicker/RangeDatePicker.tsx
```diff
@@ -0,0 +1,264 @@
+import * as React from 'react';
```
sgraczyk, line 1: Import only what's needed

packages/react-components/src/components/DatePicker/DatePickerNavbar.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+    if (numberOfMonths === 2 && isSameMonth(newMonth, toMonth)) {
+      return onMonthChange(subMonths(newMonth, 1));
+    }
+    return onMonthChange(newMonth);
+  };
+
+  return (
+    <div className={className}>
+      <div style={{ display: 'flex' }}>
```
sgraczyk, line 79: Why not a class-based implementation?

sgraczyk:
Make sure the package-lock changes are expected

sgraczyk (approved):

packages/react-components/src/components/DatePicker/DatePickerRangeCalendarsWrapper.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react';
+import cx from 'classnames';
```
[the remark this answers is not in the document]
MichalPaszowski: 👌🏼

packages/react-components/package.json
MichalPaszowski: 👌🏼

packages/react-components/src/components/DatePicker/DatePicker.spec.tsx
MichalPaszowski: will create separate task for that

packages/react-components/src/components/DatePicker/RangeDatePicker.tsx
MichalPaszowski: 👌🏼

packages/react-components/src/components/Datepicker/RangeDatePicker.tsx
MichalPaszowski: we have created a task for that https://github.com/livechat/design-system/issues/379

TomaszRatajczyk (approved):

PR #303, opened 2022-03-14
DS-162 tooltip

MichalPaszowski:
Tooltip component that is supposed to replace an older one that is using Popper.
Components : Tooltip - Tooltip ⋅ Storybook 2022-03-14 09-02-19
Components : Tooltip - Tooltip Info ⋅ Storybook 2022-03-14 09-02-47


packages/react-components/package.json
```diff
@@ -59,9 +59,12 @@
     "typescript": "^4.4.2"
   },
   "dependencies": {
+    "@emotion/css": "^11.7.1",
+    "@floating-ui/react-dom": "^0.6.0",
     "@livechat/design-system-icons": "^1.0.0-alpha.2",
     "@livechat/design-system-styles": "^1.0.0-alpha.2",
     "@popperjs/core": "^2.11.2",
+    "@types/react-transition-group": "^4.4.4",
```
przemuh, line 67: `types` shouldn't be a part of deps. Please move it to devDeps ;)

MichalPaszowski: 👌🏼

PR #304, opened 2022-03-14
DS-192 - Alert container full width

marcinsawicki:
https://livechatinc.atlassian.net/browse/DS-192


skrabkab-design:
image
Please make sure to keep icon alignment consistent with the designs in Figma

image
I think this margin is not needed,

image

We should apply the same approach to Promo banner component

Can we add another story that would show the version with buttons? It's really popular use case and we don't have this in the documentation.

image
Can we simplify the names of those stories?

MichalPaszowski (approved):

przemuh (approved):

PR #306, opened 2022-03-15
DS-164 - Search

marcinsawicki:
https://livechatinc.atlassian.net/browse/DS-164


packages/react-components/src/components/Search/Search.stories.tsx
```diff
@@ -0,0 +1,87 @@
+import * as React from 'react';
+import { ComponentMeta, Story } from '@storybook/react';
+
+import { Search as SearchComonent, ISearchProps, SearchSize } from './Search';
```
MichalPaszowski, line 4: SearchComonent all the way down

packages/react-components/src/components/Search/Search.module.scss
```diff
@@ -0,0 +1,90 @@
+$base-class: 'search';
```
MichalPaszowski, line 1: Have we abandoned 'lc-something' convention?

przemuh: https://github.com/livechat/design-system/pull/291#discussion_r845897391

przemuh: However, I would re-think the naming of the class and the component itself. `Search` is too generic for me. I would suggest `SearchInput`. 

What do you think guys?

packages/react-components/src/components/Search/Search.spec.tsx
```diff
@@ -0,0 +1,143 @@
...
+};
+
+describe('<Search> component', () => {
+  it('should render with default size', () => {
+    const { getByRole } = renderComponent({
+      ...defaultProps,
+    });
+
+    expect(getByRole('textbox')).toHaveClass(
+      styles[`${inputBaseClass}--compact`]
```
przemuh, line 25: I feel the inconsistency here. Why this component is rendered as a compact by default when other components in DS have a medium default size?

packages/react-components/src/components/Search/Search.spec.tsx
```diff
@@ -0,0 +1,143 @@
...
+      ...defaultProps,
+      size: SearchSize.Large,
+    });
+
+    expect(getByRole('textbox')).toHaveClass(
+      styles[`${inputBaseClass}--large`]
+    );
+  });
+
+  it('should render as disabled', () => {
```
przemuh, line 51: I would check the `disabled` attribute on input also.

packages/react-components/src/components/Search/Search.spec.tsx
```diff
@@ -0,0 +1,143 @@
...
+    );
+  });
+
+  it('should render with given custom placeholder', () => {
+    const { queryByPlaceholderText } = renderComponent({
+      ...defaultProps,
+      placeholder: 'Custom placeholder',
+    });
+
+    expect(queryByPlaceholderText(/Custom placeholder/i)).toBeTruthy();
```
przemuh, line 95:
```suggestion
    expect(queryByPlaceholderText(/Custom placeholder/i)).toBeVisible();
```

packages/react-components/src/components/Search/Search.stories.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+const StoryTemplate: Story<ISearchProps> = (args: ISearchProps) => (
+  <div style={{ width: 300 }}>
+    <SearchComponent {...args} />
+  </div>
+);
+
+export const Search = StoryTemplate.bind({});
+Search.args = {
+  onChange: (v) => console.log(v),
```
przemuh, line 22: Please don't use console.logs or alerts. The better alternative is to use `actions` https://storybook.js.org/docs/react/essentials/actions

przemuh:
Shouldn't it be collapsed when value is empty and focus is gone?

https://user-images.githubusercontent.com/3865478/162409834-ab70d526-e8cc-46af-9c92-ee576e47f2aa.mov

packages/react-components/src/components/Search/Search.tsx
```diff
@@ -0,0 +1,129 @@
...
+  const handleOnChange = (e: React.FormEvent<HTMLInputElement>) => {
+    const value = e.currentTarget.value;
+
+    setSearchValue(value);
+
+    if (isControlledSubmit) {
+      return;
+    }
+
+    return onChange(value);
```
przemuh, line 64: I don't buy it :) Could you elaborate more on this "feature"?

marcinsawicki:
> Shouldn't it be collapsed when value is empty and focus is gone?
> 
>  Screen.Recording.2022-04-08.at.11.39.20.mov

Hard to say. Old implementation does not close after losing focus.

packages/react-components/src/components/Search/Search.tsx
marcinsawicki: `isControlledSubmit` is a flag to control input behavior, disabling firing onChange automatically.

MichalPaszowski (approved):

packages/react-components/src/components/Search/Search.module.scss
AlithAnar: ☝️

packages/react-components/src/components/Search/Search.spec.tsx
AlithAnar: ☝️

packages/react-components/src/components/Search/Search.stories.tsx
```diff
@@ -0,0 +1,79 @@
...
+const StoryTemplate: Story<ISearchProps> = (args: ISearchProps) => (
+  <div style={{ width: 300 }}>
+    <SearchComponent {...args} />
+  </div>
+);
+
+export const Search = StoryTemplate.bind({});
+Search.args = {};
+
+export const states = (args: ISearchProps): React.ReactElement => (
```
AlithAnar, line 24: if this is a component, shouldn't it starts with upper-case?

packages/react-components/src/components/Search/Search.stories.tsx
```diff
@@ -0,0 +1,79 @@
...
+    <div style={containerStyles}>
+      <div style={textStyles}>Collapsable</div>
+      <SearchComponent {...args} isCollapsable />
+    </div>
+  </div>
+);
+
+states.args = {};
+
+export const sizes = (args: ISearchProps): React.ReactElement => (
```
AlithAnar, line 47: as above

packages/react-components/src/components/Search/Search.tsx
```diff
@@ -0,0 +1,124 @@
...
+    (isDisabled || isLoading) && styles[`${inputBaseClass}--disabled`],
+    isCollapsable && styles[`${inputBaseClass}--collapsable`],
+    !isCollapsed && styles[`${inputBaseClass}--collapsable--open`]
+  );
+
+  const handleOnChange = (e: React.FormEvent<HTMLInputElement>) => {
+    const value = e.currentTarget.value;
+
+    setSearchValue(value);
+    return !isControlledSubmit && onChange(value);
```
walaszczykm, line 59:
```suggestion
    if (!isControlledSubmit) {
      onChange(value);
    }
```

packages/react-components/src/components/Search/Search.tsx
```diff
@@ -0,0 +1,124 @@
...
+  };
+
+  const handleOnClear = () => {
+    setSearchValue('');
+    return onChange('');
```
walaszczykm, line 64:
```suggestion
    onChange('');
```

packages/react-components/src/components/Search/Search.tsx
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    inputRef.current?.focus();
+    return setIsCollapsed(false);
+  };
+
+  const handleOnKeyDown = (e: React.KeyboardEvent<HTMLInputElement>) => {
+    if (isControlledSubmit && e.key === KeyCodes.enter) {
+      e.preventDefault();
+      return onChange(searchValue);
```
walaszczykm, line 79:
```suggestion
      onChange(searchValue);
```

packages/react-components/src/components/Search/Search.tsx
```diff
@@ -0,0 +1,124 @@
...
+  };
+
+  const handleOnClick = () => {
+    if (!isCollapsable) {
+      return;
+    }
+
+    inputRef.current?.focus();
+    return setIsCollapsed(false);
+  };
```
walaszczykm, line 74: The collapsible search can only be expanded once, and then it never collapses again. IMO if we give an option to use collapsible input, it should collapse again when it is not focused and has no input value.

Additionally, the current implementation expands the search input on click, so it is not supporting keyboard navigation. Relying on focus/blur events, you can handle both: pointer and keyboard navigation 🙂 


https://user-images.githubusercontent.com/9700566/169520291-c422f395-9459-4b9e-8dbe-74142c915685.mov


```suggestion
  const handleOnClick = () => {
    inputRef.current?.focus();
  };

  const handleOnFocus = () => {
    if (isCollapsable) {
      setIsCollapsed(false);
    }
  };

  const handleOnBlur = () => {
    if (isCollapsable && !searchValue) {
      setIsCollapsed(true);
    }
  };
```

packages/react-components/src/components/Search/Search.tsx
```diff
@@ -0,0 +1,124 @@
...
+const inputBaseClass = `${baseClass}__input`;
+
+export const enum SearchSize {
+  Compact = 'compact',
+  Medium = 'medium',
+  Large = 'large',
+}
+
+export interface ISearchProps {
+  isControlledSubmit?: boolean;
```
walaszczykm, line 23: I don't understand the purpose of this flag. I know what controlled, and uncontrolled inputs are but in this case, using these flags changes this component from calling `onChange` on actual input change to on entering key down. Such behavior is really weird and unintuitive from the consumer's point of view. If we really want to give such a possibility to additionally handle "submitting," then this should have its own callback `onSubmit` and not reusing `onChange`. On the other hand, having an input that can only be submitted from the keyboard without any action button to do the same thing is also weird and not clear for the end-user.

IMO I would remove this prop and logic related to it. In order to have controlled / uncontrolled versions of this component, we just need to respect potential changes in `value` prop during the component lifetime and override the internal `searchValue` by it. It can be achieved using simple `useEffect`:

```ts
React.useEffect(() => {
    if (typeof value === 'string') {
      setSearchValue(value);
    }
  }, [value]);
```

packages/react-components/src/components/Search/Search.tsx
```diff
@@ -0,0 +1,135 @@
...
+  isCollapsable?: boolean;
+  isDisabled?: boolean;
+  isLoading?: boolean;
+  placeholder?: string;
+  size?: SearchSize;
+  value?: string;
+  onChange: (value: string) => void;
+}
+
+export const Search: React.FC<ISearchProps> = ({
```
AlithAnar, line 32:
```suggestion
export const SearchInput: React.FC<ISearchInputProps> = ({
```

AFAIR Design System project don't use "I" prefix for interfaces but I may be not familiar with the current convention.

packages/react-components/src/components/Search/Search.tsx
```diff
@@ -0,0 +1,135 @@
...
+        type="text"
+        value={searchValue}
+        placeholder={placeholder}
+        onChange={handleOnChange}
+        onBlur={handleOnBlur}
+        onFocus={handleOnFocus}
+        onKeyDown={handleOnKeyDown}
+        disabled={isDisabled || isLoading}
+      />
+      {!!searchValue && !isDisabled && !isLoading && (
```
AlithAnar, line 116: consider extracting this to a separate const

```TS
const isCloseIconVisible = !!searchValue && !isDisabled && !isLoading
```

AlithAnar (approved):

packages/react-components/src/components/Search/Search.tsx
walaszczykm: @AlithAnar looking at other examples it looks like on `v1` this prefix is used

AlithAnar: 👍

packages/react-components/src/components/Search/Search.spec.tsx
```diff
@@ -0,0 +1,128 @@
...
+  });
+
+  it('should clear input value if clear icon clicked', () => {
+    const { getByTestId, getByRole } = renderComponent({
+      ...defaultProps,
+      value: 'test',
+    });
+
+    fireEvent.click(getByTestId(`${baseClass}-clear-icon`));
+    expect(getByRole('textbox')).toHaveTextContent('');
```
walaszczykm, line 126: This assertion is always true as the input element does not have any text content so its default value is an empty string. Instead, you should use `toHaveValue` in order to check `value` property of the input element. 

There is a good rule "never trust tests you haven't seen failed" so whenever you write a test, make sure it is possible to fail on expected issue 🙂 

```suggestion
    expect(getByRole('textbox')).toHaveValue('');
```

walaszczykm (approved):
Looks great! 👏 
I have left one last comment related to tests, just please adjust it before merging the PR 🙂

AlithAnar (approved):
approving but please implement @walaszczykm suggestions

PR #311, opened 2022-04-08
DS-162 vite version of tooltip

MichalPaszowski:
Tooltip component that is supposed to replace an older one that is using Popper.
Components : Tooltip - Tooltip ⋅ Storybook 2022-03-14 09-02-19
Components : Tooltip - Tooltip Info ⋅ Storybook 2022-03-14 09-02-47


packages/react-components/src/components/Tooltip/Info.tsx
```diff
@@ -0,0 +1,30 @@
+import * as React from 'react';
+import { Icon, IconTypeName } from '../Icon';
+import { Close } from '@livechat/design-system-icons/react/material';
+import { getIconType } from './helpers';
+import styles from './Tooltip.module.scss';
+
+const baseClass = 'lc-tooltip';
```
marcinsawicki, line 7: `tooltip`

packages/react-components/src/components/Tooltip/Info.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+export const Info: React.FC<{
+  header: string;
+  text: string;
+  closeWithX?: boolean;
+  theme?: string;
+  handleCloseOnClick?: () => void;
```
marcinsawicki, line 14: `handleOnClick` ?

marcinsawicki: or `onCloseClick`?

packages/react-components/src/components/Tooltip/Interactive.tsx
```diff
@@ -0,0 +1,78 @@
+import * as React from 'react';
+import { Icon, IconTypeName } from '../Icon';
+import { Close } from '@livechat/design-system-icons/react/material';
+import { Button } from '../Button';
+import { getIconType } from './helpers';
+import styles from './Tooltip.module.scss';
+
+const baseClass = 'lc-tooltip';
```
marcinsawicki, line 8: Same. We don't use `lc-` prefix anymore

packages/react-components/src/components/Tooltip/Info.tsx
```diff
@@ -0,0 +1,30 @@
...
+}> = ({ header, text, closeWithX, theme, handleCloseOnClick }) => {
+  return (
+    <div style={{ position: 'relative' }}>
+      {closeWithX && (
+        <div className={styles[`${baseClass}-x`]} onClick={handleCloseOnClick}>
```
marcinsawicki, line 19: IMO this class should be named `${baseClass}-close`. WDYT?

packages/react-components/src/components/Tooltip/Interactive.tsx
```diff
@@ -0,0 +1,78 @@
...
+  handleClickSecondary,
+}) => {
+  const decoration = theme === 'invert' ? 'underline' : 'none';
+  return (
+    <div style={{ width: '270px' }}>
+      <div
+        style={{ position: 'relative', height: '25px', marginBottom: '10px' }}
+      >
+        {closeWithX && (
+          <div className={styles[`${baseClass}-x`]}>
```
marcinsawicki, line 39: Same

packages/react-components/src/components/Tooltip/Tooltip.module.scss
```diff
@@ -0,0 +1,129 @@
+$base-class: 'lc-tooltip';
```
marcinsawicki, line 1: `tooltip`

marcinsawicki (commented):
Left some comments.
Mainly it concerns removal `lc-` prefix from components, also consider renaming the methods to the ones I have proposed.

packages/react-components/src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,272 @@
...
+    if (isManaged) return;
+    if (visible) {
+      handleClose();
+    } else {
+      handleOpen();
+    }
+  };
+
+  const handleCloseOnClick = () => {
+    handleClose();
```
walaszczykm, line 155: we shouldnt allow closing the tooltip if it is managed
```suggestion
    if (isManaged) return;
    handleClose();
```

packages/react-components/src/components/Tooltip/Simple.tsx
```diff
@@ -0,0 +1,5 @@
+import * as React from 'react';
+
+export const Simple: React.FC<{ text: string }> = ({ text }) => {
```
walaszczykm, line 3: what is the purpose of this component? 🤔  why should I use it instead of `div` element?

packages/react-components/src/components/Tooltip/UserGuide.tsx
```diff
@@ -0,0 +1,89 @@
+import * as React from 'react';
+import { ModalPortal, ModalPortalProps } from '../Modal/';
+import { ITooltipProps, Tooltip } from './Tooltip';
+import SpotlightOverlay from './SpotlightOverlay';
+import cx from 'clsx';
+import styles from './Tooltip.module.scss';
+import { ClientRectObject } from '@floating-ui/core';
+import { throttle } from '@livechat/data-utils';
```
walaszczykm, line 8: not used
```suggestion
```

packages/react-components/src/components/Tooltip/UserGuide.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+const spotlightPadding = 200;
```
walaszczykm, line 10: not used
```suggestion
```

packages/react-components/src/components/Tooltip/UserGuide.tsx
```diff
@@ -0,0 +1,89 @@
...
+    const element = document.querySelector(parentElementName);
+    setParentElement(element);
+  }, [parentElementName]);
+
+  React.useEffect(() => {
+    parentElement && setRect(parentElement.getBoundingClientRect());
+  }, [parentElement]);
+
+  return parentElement ? (
+    <ModalPortal
```
walaszczykm, line 59: what is the purpose of using `ModalPortal` component here? 🤔 
I gett the same working experience when changing this to just `div` element
```suggestion
    <div>
```

packages/react-components/src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,270 @@
...
+  }, []);
+
+  React.useEffect(() => {
+    if (!refs.reference.current || !refs.floating.current) {
+      return;
+    }
+
+    // Only call this when the floating element is rendered
+    return autoUpdate(refs.reference.current, refs.floating.current, update);
+  }, [refs.reference, refs.floating, update, updatedPlacement]);
```
walaszczykm, line 110: This effect needs to rely on `visible` state value even if it is not directly used here because it synchronizes the visibility with the tooltip position. The "refs" are React Refs, which means their values are not reactive; hence the change is not triggering rerender and replying to the effect. This causes the bug when the tooltip is not opened initially as it is not properly tracking the element on the screen.
```suggestion
  }, [refs.reference, refs.floating, update, updatedPlacement, visible]);
```

Without the change:
<video src="https://user-images.githubusercontent.com/9700566/169988307-4674d39e-2a26-47c0-a334-544f49aa2e4c.mov" />

After syncing with `visible` state:
<video src="https://user-images.githubusercontent.com/9700566/169988426-0149eac0-7ba6-4605-80c4-47260e81964f.mov" />

packages/react-components/src/components/Tooltip/Info.tsx
MichalPaszowski: 👌🏼

packages/react-components/src/components/Tooltip/Info.tsx
MichalPaszowski: handleCloseClick 👌🏼

packages/react-components/src/components/Tooltip/Interactive.tsx
MichalPaszowski: 👌🏼

packages/react-components/src/components/Tooltip/Interactive.tsx
MichalPaszowski: 👌🏼

packages/react-components/src/components/Tooltip/Info.tsx
MichalPaszowski: 👌🏼

packages/react-components/src/components/Tooltip/Tooltip.module.scss
MichalPaszowski: 👌🏼

packages/react-components/src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,272 @@
...
+  };
+
+  const top = arrowOffsetY && arrowY ? arrowY + arrowOffsetY : arrowY;
+  const left = arrowOffsetX && arrowX ? arrowX + arrowOffsetX : arrowX;
+
+  const mergedClassNames = cx(styles[baseClass], {
+    [className ]: className,
+    [styles[`${baseClass}--invert`]]: theme === 'invert',
+    [styles[`${baseClass}--important`]]: theme === 'important',
+  });
```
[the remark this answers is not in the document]
MichalPaszowski: 👌🏼

packages/react-components/src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,272 @@
...
+export const Tooltip: React.FC<ITooltipProps> = (props) => {
+  const {
+    triggerRenderer,
+    referenceElement,
+    children,
+    className,
+    theme,
+    placement,
+    isVisible = false,
+    isManaged = false,
```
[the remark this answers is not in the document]
MichalPaszowski: YOu are correct - I check if isVisible is undefined and if it is not - I set 'isManaged' to true

packages/react-components/src/components/Tooltip/Tooltip.tsx
walaszczykm: Ok, after retesting the changes, I see that this one was not the best advice and needs to be reversed, apologies 🙇‍♂️ 

Mainly because in the case of the managed components we still need information about internal close button clicks. Instead of blocking the on close handler, we should only prevent changing the `visible` state directly.

So it would be best to wrap `setVisibility` calls with `if(!isManaged)` in `handleOpen` and `handleClose` 🙂

MichalPaszowski: I think we should allow this - the user can decide if he wants to show the 'x' icon or listen for the ESC key that allows customers to break whatever flow he is in.

packages/react-components/src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,422 @@
...
+    </span>
+    <span className={'tooltip-icon-wrap'}>
+      <Icon source={Attachment}></Icon>
+    </span>
+  </div>
+);
+
+Tooltip.args = {
+  placement: 'bottom',
+  isVisible: true,
```
walaszczykm, line 64: IMO for a basic tooltip example, it would be great to showcase its default behavior with a hover state. That way we shouldn't provide the `isVisisble` value.

```suggestion
```

packages/react-components/src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,422 @@
...
+      alignItems: 'center',
+    }}
+  >
+    <TooltipInfoExample {...args}></TooltipInfoExample>
+  </div>
+);
+
+TooltipInfo.args = {
+  placement: 'bottom',
+  isVisible: true,
```
walaszczykm, line 91: On the TooltipInfo story it would be great to use the button which is already there and showcase the simple managed tooltip case so button click => tooltip is shown, close button or esc => tooltip is closed
```suggestion
```

packages/react-components/src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,422 @@
...
+            />
+          ) : null}
+        </TooltipUserGuideComponent>
+      </div>
+    </div>
+  );
+};
+
+const TooltipInfoExample: React.FC<ITooltipProps> = (props) => {
+  return (
```
walaszczykm, line 363:
```suggestion
const TooltipInfoExample: React.FC<ITooltipProps> = (props) => {
  return (
```
and here continuing the case of managed tooltip we can add some local state
```suggestion
const TooltipInfoExample: React.FC<ITooltipProps> = (props) => {
  const [isVisible, setIsVisible] = React.useState(false);
  return (
```

packages/react-components/src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,422 @@
...
+        width: '500px',
+        height: '600px',
+        display: 'flex',
+        alignItems: 'center',
+        justifyContent: 'center',
+      }}
+    >
+      <TooltipComponent
+        {...props}
+        triggerRenderer={() => (
```
walaszczykm, line 375: here we can provide `isVisisble` from local state as alsouse `onClose` handler to change the state 
```suggestion
        {...props}
        isVisible={isVisible}
        onClose={() => {
          setIsVisible(false);
        }}
        triggerRenderer={() => (
```

packages/react-components/src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,422 @@
...
+          <div>
+            <Button
+              icon={<Icon source={DropDown}></Icon>}
```
walaszczykm, line 378: and finally use the Button to open the Tooltip 🙂 
```suggestion
            <Button
              onClick={() => setIsVisible(true)}
              icon={<Icon source={DropDown}></Icon>}
```

packages/react-components/src/components/Tooltip/Tooltip.tsx
walaszczykm: @MichalPaszowski exactly, you are totally right 🙂

packages/react-components/src/components/Tooltip/Simple.tsx
MichalPaszowski: The idea behind this is that we provide a closed set of components that satisfy the needs of AA. We wanted to avoid situations when the content of the tooltip does not follow the style guidelines provided by DS.

packages/react-components/src/components/Tooltip/UserGuide.tsx
MichalPaszowski: 👌🏼

packages/react-components/src/components/Tooltip/Simple.tsx
walaszczykm: cool, sounds reasonable 👍

packages/react-components/src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,422 @@
...
+          {state.reference === 'reference2' ? (
+            <UserGuideStep
+              header="Header - concise and clear"
+              image={{
+                src: beutifulImage,
+                alt: 'image',
+              }}
+              text="Tooltip content is used to explain the details of elements or features."
+              handleClickPrimary={() => dispatch({ type: 'reference3' })}
+              handleCloseClick={() => dispatch({ type: 'isVisible' })}
```
MichalPaszowski, line 332: better solution?

packages/react-components/src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,270 @@
...
+        top: y ?? '',
+        left: x ?? '',
+      }}
+      className={mergedClassNames}
+      onMouseEnter={handleMouseEnter}
+      onMouseLeave={handleMouseLeave}
+    >
+      {React.Children.map(children, (child) => {
+        if (React.isValidElement(child)) {
+          return React.cloneElement(child, { handleCloseClick, theme });
```
MichalPaszowski, line 179: it overwrites what we pass as props when using component

packages/react-components/src/components/Tooltip/Tooltip.stories.tsx
MichalPaszowski: https://github.com/livechat/design-system/pull/311#discussion_r880319126

packages/react-components/src/components/Tooltip/Tooltip.stories.tsx
MichalPaszowski: https://github.com/livechat/design-system/pull/311#discussion_r880319126

packages/react-components/src/components/Tooltip/Tooltip.tsx
MichalPaszowski: thanks!

packages/react-components/src/components/Tooltip/UserGuide.tsx
MichalPaszowski: Gonna be honest - I have copied it from the original design system. 
It gets even funnier because I experimented with ModalPortal in AA and switched it to div in a few places - the app works the same.

marcinsawicki (approved):

packages/react-components/src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,274 @@
...
+    void sleep(hoverOutDelayTimeout).then(() => {
+      if (!isHovered.current) {
+        setVisibility(false);
+      }
+    });
+  };
+
+  const handleOpen = () => {
+    if (onOpen) onOpen();
+    setVisibility(true);
```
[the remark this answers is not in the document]
MichalPaszowski: 👌🏼

walaszczykm (approved):
🚀 🚀 🚀

PR #312, opened 2022-04-08
DS-202 - Responsive version of promo banner

marcinsawicki:
https://livechatinc.atlassian.net/browse/DS-202


packages/react-components/src/components/PromoBanner/PromoBanner.tsx
```diff
@@ -22,7 +23,7 @@ export interface PromoBannerProps {
   img?: string;
   light?: boolean;
   linkText?: string;
-  size?: PromoBannerSize;
+  // size?: PromoBannerSize;
```
przemuh, line 26: ?

packages/react-components/src/components/PromoBanner/PromoBanner.tsx
```diff
@@ -36,23 +37,43 @@ export const PromoBanner: React.FC<PromoBannerProps> = ({
   img,
   light = false,
   linkText,
-  size = PromoBannerSize.Small,
+  // size = PromoBannerSize.Small,
```
przemuh, line 40: ?

MichalPaszowski (approved):

packages/react-components/src/components/PromoBanner/PromoBanner.tsx
```diff
@@ -32,23 +30,42 @@ export const PromoBanner: React.FC<PromoBannerProps> = ({
...
     },
     className
   );
 
-  const shouldRenderLargeFooter = (buttonText || linkText) && size === 'large';
-  const shouldRenderSmallOrMediumFooter =
-    (buttonText || linkText) && size !== 'large';
+  React.useEffect(() => {
+    const handleResize = debounce(() => {
+      if (containerRef.current && containerRef.current.offsetWidth <= 400) {
```
AlithAnar, line 53: please add those 400 and 800 thresholds to the const above or maybe to the upper level if those are reusable across other components

packages/react-components/src/components/PromoBanner/PromoBanner.tsx
```diff
@@ -32,23 +30,42 @@ export const PromoBanner: React.FC<PromoBannerProps> = ({
   img,
   light = false,
   linkText,
-  size = 'small',
   onButtonClick,
   onClose,
   onLinkClick,
 }) => {
+  const containerRef = React.useRef<HTMLDivElement>(null);
+  const [isSmallContainer, setIsSmallContainer] = React.useState(false);
```
AlithAnar, line 38: why not using the enum with container size (small / large) instead of 2 separate states?

packages/react-components/src/components/PromoBanner/PromoBanner.tsx
```diff
@@ -32,23 +30,42 @@ export const PromoBanner: React.FC<PromoBannerProps> = ({
...
+        return setIsSmallContainer(true);
+      }
+
+      if (containerRef.current && containerRef.current.offsetWidth >= 800) {
+        return setIsLargeContainer(true);
+      }
+
+      setIsLargeContainer(false);
+      return setIsSmallContainer(false);
+    }, 500);
```
AlithAnar, line 63: 500 => const

packages/react-components/src/components/PromoBanner/PromoBanner.tsx
marcinsawicki: Because we have three different states of how this banner is displayed, based on the width, I think we need to have those states.

walaszczykm: The number of possible states in which the banner can be displayed does not directly imply the number of state pieces you need. As long as they are mutually exclusive, they can be stored in a single state value as only one of the possible variants can be set at the time. 

Additionally, the current implementation uses two boolean values, which can store 2^2 = 4 possible values, while here, we can 3. That means there is one corrupted state theoretically possible to reach, which later can cause bugs: It has both:`isSmallContainer: true` and `isLargeContainer: true`.

Using a single state with an explicit set of possible values reduces such risk of going into an impossible state completely.
You can represent it like this:
```suggestion
  const [containerSize, setContainerSize] = React.useState<'small' | 'medium' | 'large'>('medium')
```

then later in places where you have used `isSmallContainer` replace it with `containerSize === 'small'` and similarly `isLargeContainer` with `containerSize === 'large'`.

packages/react-components/src/components/PromoBanner/PromoBanner.tsx
```diff
@@ -32,23 +34,51 @@ export const PromoBanner: React.FC<PromoBannerProps> = ({
...
+        containerRef.current &&
+        containerRef.current.offsetWidth >= LARGE_CONTAINER_WIDTH_TRESHOLD
+      ) {
+        return setContainerSize('large');
+      }
+
+      return setContainerSize('medium');
+    }, RESIZE_DEBOUNCE_TRESHOLD);
+
+    window.addEventListener('resize', handleResize);
```
walaszczykm, line 75: While finally testing your changes, I have spotted a bug related to the initial size of the component when the window is initially small or large. Here is a proposal for fixing this behavior.

In addition to registering the `handleResize` as a window resize event handler, it should be immediately called after the component is mounted to adjust it to the initial window size. Currently, the component is stuck with `'medium'` size until the first window resizes happens, but it can be loaded into large or small sizes at the beginning. 

To do this properly, you don't want to call the debounced function not to introduce unnecessary delay on the init. I would propose extracting the function denounced here to its own variable and calling it directly. It can be done like this:

```ts
    const handleResize = () => {
       /* ... */
    }

    const debouncedHandleResize = debounce(handleResize, RESIZE_DEBOUNCE_TRESHOLD);
    window.addEventListener('resize', debouncedHandleResize);

    handleResize();

    return () => {
      debouncedHandleResize.cancel();
      window.removeEventListener('resize', debouncedHandleResize);
    };
```

marcinsawicki: 👍

AlithAnar (approved):

walaszczykm (approved):

PR #316, opened 2022-04-21
Forms

przemuh:
A little cleanup in form components

I've removed variants of InputField, NumericInputFIeld etc. From now on we should use a composition layer from FormField.


packages/react-components/src/components/Input/Input.stories.tsx
```diff
@@ -0,0 +1,24 @@
...
+Input.args = {
+  size: 'medium',
+  placeholder: 'Placeholder text',
+  error: false,
+  disabled: false,
+} as InputProps;
+
+export const WithIcon = () => <div>TODO</div>;
+export const WithArrow = () => <div>TODO</div>;
+export const WithIconAndArrow = () => <div>TODO</div>;
```
sgraczyk, line 24: Requires adding examples before merging.

MichalPaszowski: Could not be completed with that component - we should merge it as it is and add a new component that will have these features.

PR #321, opened 2022-05-17
InAppMessage component

marcinsawicki:
Addresses [#334](https://github.com/livechat/design-system/issues/334)


marcinsawicki:
What about extending the default Storybook props description? Something like this:
components : InAppMessage - Default In App ⋅ Storybook 🔊 2022-05-18 16-28-07

packages/react-components/src/components/InAppMessages/InAppMessage.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+  const handleCloseInAppMessage = () => onClose();
+
+  const handleOnOverlayClick = (e: React.MouseEvent<HTMLDivElement>) => {
+    if (e.target === e.currentTarget) {
+      handleCloseInAppMessage();
+    }
+  };
+
+  const handleOnCloseButtonClick = () => handleCloseInAppMessage();
```
MichalPaszowski, line 69: I am not sure if that's needed - I would just use onClose in line 83

packages/react-components/src/components/InAppMessages/InAppMessageImage.spec.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react';
+import { render } from 'test-utils';
+import { InAppMessageImage } from './InAppMessageImage';
+
+const defaultProps = {
```
MichalPaszowski, line 5: I think we should keep that gif locally

MichalPaszowski (approved):

walaszczykm:
There is an issue with InApp layout on narrow screens

https://user-images.githubusercontent.com/9700566/169543229-1fe12403-d400-4460-b06f-a09d0bf1a525.mov

packages/react-components/src/components/InAppMessages/InAppMessage.tsx
```diff
@@ -0,0 +1,104 @@
...
+  header,
+  footer,
+  onClose,
+}) => {
+  React.useEffect(() => {
+    const onDocumentClick = (e: MouseEvent) => {
+      if (e.target === e.currentTarget) {
+        handleCloseInAppMessage();
+      }
+    };
```
walaszczykm, line 42: Why need a document click handler if there is already a backdrop/overly click handler? 🤔

packages/react-components/src/components/InAppMessages/InAppMessage.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+import styles from './InAppMessage.module.scss';
+
+const baseClass = 'in-app-message';
+
+export interface InAppMessageProps {
+  className?: string;
+  image?: InAppMessageImageProps;
+  header?: InAppMessageHeaderProps;
+  footer?: InAppMessageFooterProps;
```
walaszczykm, line 23: IMO in this case, the composition is better than the configuration. Instead of providing all of those nested props objects, why don't you just provide an instance of the image/header/footer component in the form of ReactElement? That way consumer will provide the required props for those subcomponents on his own. 

Additionally, React composition allows reducing the number of unnecessary re-renders because you are not creating new object references on each render.  The sent thing also is that typically UI library built around composition is easier to use and extend in more advanced cases.

So here the types will look like this:
```suggestion
  image?: React.ReactElement;
  header?: React.ReactElement;
  footer?: React.ReactElement;
```

And later on, render:
```ts
  return (
    <div
      data-testid={`${baseClass}-overlay`}
      className={styles[`${baseClass}__overlay`]}
      onMouseDown={handleOnOverlayClick}
    >
      <div className={styles[`${baseClass}__container`]}>
        {header}
        <div className={mergedClassNames}>
          <div className={styles[`${baseClass}__content`]}>
            {image}
            <div
              className={cx(styles[`${baseClass}__content__body`], {
                [styles[`${baseClass}__content__body__with-footer`]]: !!footer,
              })}
            >
              {children}
            </div>
          </div>
          {footer}
        </div>
      </div>
    </div>
  );
```

WDYT? 🤔

walaszczykm:
there is something wrong going on with buttons, their layout is broken on narrower screens

https://user-images.githubusercontent.com/9700566/171021760-67f3aac9-e961-449c-abaf-e87354c2aa86.mov

walaszczykm:
The image completely disappears on narrower screens

https://user-images.githubusercontent.com/9700566/171022002-84792d32-50eb-41c7-99f6-2bc65b886cab.mov

packages/react-components/src/components/InAppMessages/InAppMessage.stories.tsx
```diff
@@ -0,0 +1,250 @@
...
+              avatar={{
+                src: 'https://avatars2.githubusercontent.com/u/29309941?s=88&v=4',
+                alt: 'Agent',
+              }}
+              text={
+                <React.Fragment>
+                  <strong>Michał</strong> from LiveChat
+                </React.Fragment>
+              }
+              onCloseButtonClick={() => setIsInAppVisible(null)}
```
walaszczykm, line 215:
```suggestion
              onCloseButtonClick={() => setIsInAppVisible(false)}
```

marcinsawicki:
> The image completely disappears on narrower screens
> 
>  Screen.Recording.2022-05-30.at.17.19.02.mov

We have query in CSS which will hide the image for the sake of displaying the content. I think it's a good approach but probably will need a designer investigation.

walaszczykm:
> > The image completely disappears on narrower screens
> > Screen.Recording.2022-05-30.at.17.19.02.mov
> 
> We have query in CSS which will hide the image for the sake of displaying the content. I think it's a good approach but probably will need a designer investigation.

oOk but this should depend on available space and content size. In the attached example it is obvious there is plenty of space for an image even if it would be shrunk a little

marcinsawicki:
Yes, You are right. I will take this into account maybe with the support from designers perspective.

marcinsawicki:
The work on this component is temporarily blocked, probably we will drop this component, as we have `Dropdown` with the possibility to add own content, which will cover the cases from `InApp` component

mreszke:
How it's going to work with system in-app messages? AFAIK usually in-apps looks like this:
```
creation_timestamp: 1656916617.637
id: "0"
message: "<div><img src=\"https://cdn.livechat-files.com/api/file/lc/cdn-uploads/inapp/payment_issue.svg\" style=\"height: 112px; width: 184px; vertical-align: middle;\"></div><div style=\"font-family: 'Source Sans Pro',sans-serif; font-size: 15px; color: #3a3c42; margin: 24px 16px 32px;\"><h2 style=\"margin-bottom: 16px; font-size: 24px; color: #4e5665;\">Your payment was declined</h2><div style=\"margin-bottom: 8px;\">Please update your credit card details to keep chatting without interruption.</div><style>#in-app-notification .close {color: #9c999d !important;} #in-app-notification .popup {padding: 32px 24px !important; width: 480px; max-width: calc(100% - 20px); max-height: calc(100% - 20px); text-align: center;border-radius:4px;  box-shadow: 0 20px 60px 0 rgba(0, 0, 0, 0.3);} #in-app-notification .controls span{text-transform: none !important;font-size: 15px !important;font-weight: 600 !important;letter-spacing: 0;background: #4787F2 !important;line-height: 30px;} #in-app-notification .cancel {display: none !important;}</style><span style=\"display:none;\" data-expiration=\"{\" exp\":{\"max\":\"7\",\"delay\":\"1\"}}\"=\"\"></span>[Update payment method].(blue flat)"
type: "important"
url: "https://livechat-test.recurly.com/account/invoices/100590?ht=QFnrvxDJYBuqwTkkbNdSKSYrGZmNt77B"
```
So we don't have a possibility to split it to `header`, `buttons` etc

sgraczyk:
> 

What modification in `Design System` you'd need in order to build `InApp` without building it directly in `DS`? Sorry for asking but at this moment I'm focused on different areas of `DS`.

PR #323, opened 2022-06-03
DS-206 - missing svg icons

marcinsawicki:
https://livechatinc.atlassian.net/browse/DS-206


packages/icons/svg/material/camera.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
AlithAnar, line 1: do we need explicit dimensions inside svgs? in AA we are removing them from svgs in favor of upper-level size management

marcinsawicki: I see that every single SVG file has those dimensions, which do not cause any problems. Do You think that we should change that?

AlithAnar: nope, I was just curious :>

MichalPaszowski: I think these are removed when converted to React components.

MichalPaszowski (approved):

PR #329, opened 2022-06-10
DS-205 Replace Popper with FloatingUI in Dropdown component

MichalPaszowski:
Popper is dead - long live Floating UI.
No design changes.

### DIFF
--- a/packages/react-components/src/components/Popover/Popover.spec.tsx
+++ b/packages/react-components/src/components/Popover/Popover.spec.tsx
@@ -1,6 +1,27 @@
 import * as React from 'react';
 import { render, fireEvent } from 'test-utils';
 import { Popover } from './Popover';
+import { vi } from 'vitest';
+
+vi.mock('@floating-ui/react-dom', () => {
+  return {
+    useFloating: vi.fn(() => {
+      return {
+        x: 0,
+        y: 0,
+        placement: 'bottom',
+        refs: {
+          floating: 0,
+          reference: 0,
+        },
+      };
+    }),
+    autoUpdate: vi.fn(),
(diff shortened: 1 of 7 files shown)


packages/react-components/src/components/Popover/Popover.stories.css
```diff
@@ -21,3 +21,10 @@
   line-height: 20px;
   margin-left: 5px;
 }
+
+.lc-testArea {
```
sgraczyk, line 25: What is `testArea`?

marcinsawicki: And why `lc-` prefix?

marcinsawicki (approved):

packages/react-components/src/components/Popover/Popover.stories.css
MichalPaszowski: I know it is a stupid name but could not come up with something better. Basically, it's a div where the component is tested :P

MichalPaszowski: fixed

packages/react-components/src/components/Popover/Popover.spec.tsx
```diff
@@ -1,6 +1,27 @@
...
+      };
+    }),
+    autoUpdate: vi.fn(),
+    flip: vi.fn(),
+    offset: vi.fn(),
+    x: 0,
+  };
+});
 
 const PopperContent = () => {
```
sgraczyk, line 26: Still using `Popper` wording. Try to avoid library naming at all.

sgraczyk (approved):

PR #336, opened 2022-07-01
Improvements for Picker component

marcinsawicki:
Addresses: https://github.com/livechat/design-system/issues/337


packages/react-components/src/components/Picker/Trigger.tsx
```diff
@@ -55,9 +57,15 @@ export const Trigger: React.FC<ITriggerProps> = ({
...
+      document.addEventListener('keydown', onKeyDown);
+    }
 
-    return () => document.removeEventListener('keydown', onKeyDown);
+    return () => {
+      if (!disabledSearch) {
+        document.removeEventListener('keydown', onKeyDown);
+      }
+    };
   }, []);
```
MichalPaszowski, line 69: I think 'disabledSearch' should be added to the dependency table. Linter didn't complain about that?

MichalPaszowski (changes requested):

packages/react-components/src/components/Picker/PickerList.module.scss
```diff
@@ -14,6 +14,7 @@ $base-class: 'picker-list';
   padding: 4px 0;
   position: absolute;
   width: 341px;
+  z-index: 10000;
```
sgraczyk, line 17: Is there any rule on how we use `z-indexes`? Is there any stacking context?

packages/react-components/src/components/Picker/Trigger.tsx
```diff
@@ -14,6 +14,7 @@ const baseClass = 'picker-trigger';
 export type TriggerSize = 'compact' | 'medium' | 'large';
 
 export interface ITriggerProps {
+  disabledSearch: boolean;
```
sgraczyk, line 17: I'd consider alternative naming as in case of `Picker`

sgraczyk: Plus there is a difference -> `disableSearch` vs `disabledSearch`

packages/react-components/src/components/Picker/Picker.tsx
```diff
@@ -21,7 +21,7 @@ export interface IPickerProps {
   placeholder?: string;
   isRequired?: boolean;
   noSearchResultText?: string;
-  disableSearch?: boolean;
+  searchDisabled?: boolean;
```
marcinsawicki, line 24: I've decided to name it `searchDisabled` as this is an interface visible to consumers.
WDYT?

packages/react-components/src/components/Picker/Trigger.tsx
```diff
@@ -14,7 +14,7 @@ const baseClass = 'picker-trigger';
 export type TriggerSize = 'compact' | 'medium' | 'large';
 
 export interface ITriggerProps {
-  disabledSearch: boolean;
+  isSearchDisabled: boolean;
```
marcinsawicki, line 17: Here we have `isSearchDisabled` as this is an internal part of `Picker` component, where our naming convention perfectly fits, the consumer doesn't see it.

packages/react-components/src/components/Picker/Picker.tsx
sgraczyk: Fine for now

packages/react-components/src/components/Picker/Trigger.tsx
sgraczyk: Fine for now, we'll rethink this later

packages/react-components/src/utils/StackingContextLevel.scss
```diff
@@ -0,0 +1,5 @@
+/**
+ * Definition of z-index values to be used across application.
+ */
+
+$stacking-context-level-dropdown: 10000;
```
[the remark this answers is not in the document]
sgraczyk: It is ok now, however, I was wondering if 1000 would be enough.

sgraczyk (approved):

MichalPaszowski (approved):

PR #353, opened 2022-07-22
#344 Handle Switch disabled state visually

### LINKED ISSUE #344: [Switch] Lock icon and cursor not-allowed for disabled state
## Description
`Switch` component should have a `lock` icon and
```css
cursor: not-allowed;
```
in case of a `disabled` state.

## Design
[Figma - Locked Switch](https://www.figma.com/file/2pFu80PXO5A2tfyrAGnx91/Product-Components?node-id=6904%3A32302)

image

sgraczyk:
Resolves #344 

## Description
- Lock icon added for the disabled state,
- Applied proper color tokens for `hover`, `enabled`, and `disabled` cases,
- Changed naming convention as `enabled/disabled` was easy to be confused with `checked (on)/unchecked (off)`,
- Added multiple stories representing the changes,
- I've used `--color-action-default-rgb` for `box-shadow`

## Storybook
- [Disabled On](https://613a8e945a5665003a05113b-muqgbybjhw.chromatic.com/?path=/story/components-switch--disabled-on)
- [Disabled Off](https://613a8e945a5665003a05113b-muqgbybjhw.chromatic.com/?path=/story/components-switch--disabled-off)

### DIFF
--- a/packages/react-components/src/components/Switch/Switch.stories.tsx
+++ b/packages/react-components/src/components/Switch/Switch.stories.tsx
@@ -1,23 +1,35 @@
 import * as React from 'react';
-import { ComponentMeta } from '@storybook/react';
+import { ComponentMeta, Story } from '@storybook/react';
 
-import { Switch as SwitchComponent, SwitchProps } from './Switch';
-import noop from '../../utils/noop';
+import { Switch, SwitchProps } from './Switch';
 
 export default {
   title: 'Components/Switch',
-  component: SwitchComponent,
-} as ComponentMeta<typeof SwitchComponent>;
+  component: Switch,
+} as ComponentMeta<typeof Switch>;
 
-export const Switch = (args: SwitchProps): React.ReactElement => {
-  return (
-    <div>
-      <SwitchComponent {...args} />
-    </div>
-  );
+const Template: Story<SwitchProps> = (args: SwitchProps) => (
+  <Switch {...args} />
+);
+
+const basicWithoutStateHandler: Partial<SwitchProps> = {
+  onChange: undefined,
 };
 
-Switch.args = {
-  size: 'basic',
-  onChange: noop,
-};
+export const BasicSize = Template.bind({});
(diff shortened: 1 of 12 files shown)


mreszke (approved):

packages/react-components/src/components/Switch/Switch.stories.tsx
```diff
@@ -1,23 +1,35 @@
...
+EnabledOn.args = { ...basicWithoutStateHandler, on: true };
+
+export const EnabledOff = Template.bind({});
+EnabledOff.args = { ...basicWithoutStateHandler, on: false };
+
+export const DisabledOn = Template.bind({});
+DisabledOn.args = { ...basicWithoutStateHandler, on: true, disabled: true };
+
+export const DisabledOff = Template.bind({});
+DisabledOff.args = { ...basicWithoutStateHandler, on: false, disabled: true };
```
marcinsawicki, line 35: I think we can minimize the number of different stories here. We can divide it into "kinds" and "states", just like we have in `Button` component stories.

marcinsawicki (approved):

marcinsawicki:
I little comment from me.

packages/react-components/src/components/Switch/Switch.stories.tsx
sgraczyk: Addressed in https://github.com/livechat/design-system/pull/353/files/b1ba318006993158ba0bbe83499b5172ff99d8fd..7b2f4ecdd7189efd1a793d7d4fc2bbda4339a3a0

sgraczyk: @marcinsawicki Please check if this is what you've expected https://github.com/livechat/design-system/pull/353/commits/e443d1288ea431e9ba211fd8e717612c0ad8c826

marcinsawicki: It's perfect 👍

PR #356, opened 2022-07-25
#341 Reorganize Badge component to reflect Figma designs

### LINKED ISSUE #341: [Badge] Apply styles according to designs
## Description
Current implementation has styles that are misaligned with Figma designs.

Styles should be (to be confirmed within URL below):
```css
  font-size: 12px;
  font-style: normal;
  font-weight: 600;
  line-height: 16px;
  letter-spacing: 0.2px;
  text-align: center;
```

## Design
https://www.figma.com/file/2pFu80PXO5A2tfyrAGnx91/Product-Components?node-id=6904%3A34178

sgraczyk:
Resolves #341

## Description
The original issue assumed only the need to apply some styles, however, the component requires some bigger refactor in order to handle various variants. Changes applied:

- Added `kind` prop instead of `secondary` boolean
- Added `size` to handle 3 cases
- Added `type` to automatically handle `alert` and `dot` variants. I've ignored `new-message` and `extended` as for now I believe other forms of content should be handled externally.

## Storybook
- [Default](https://613a8e945a5665003a05113b-nhdgxclxti.chromatic.com/?path=/story/components-badge--default)
- [Sizes](https://613a8e945a5665003a05113b-nhdgxclxti.chromatic.com/?path=/story/components-badge--sizes)
- [Kinds](https://613a8e945a5665003a05113b-nhdgxclxti.chromatic.com/?path=/story/components-badge--kinds)
- [Types](https://613a8e945a5665003a05113b-nhdgxclxti.chromatic.com/?path=/story/components-badge--types)

### DIFF
--- a/packages/react-components/src/components/Badge/Badge.tsx
+++ b/packages/react-components/src/components/Badge/Badge.tsx
@@ -3,22 +3,33 @@ import * as React from 'react';
 import styles from './Badge.module.scss';
 import cx from 'clsx';
 
+const baseClass = 'badge';
+
 export interface BadgeProps extends React.HTMLAttributes<HTMLDivElement> {
-  secondary?: boolean;
+  kind?: 'primary' | 'secondary' | 'tertiary';
+  size?: 'large' | 'medium' | 'compact';
+  type?: 'content' | 'alert' | 'dot';
 }
 
 export const Badge: React.FC<BadgeProps> = ({
   children,
   className,
-  secondary = false,
+  kind = 'primary',
+  size = 'medium',
+  type = 'content',
 }) => {
-  return (
-    <span
-      className={cx(styles.badge, className, {
-        [styles['badge--secondary']]: secondary,
-      })}
-    >
-      {children}
-    </span>
+  const mergedClassNames = cx(
+    className,
+    styles[baseClass],
+    styles[`${baseClass}--${kind}`],
+    styles[`${baseClass}--${size}`]
   );
+
+  const content = {
+    ['content']: children,
+    ['alert']: '!',
+    ['dot']: <span className={styles[`${baseClass}__dot`]} />,
+  }[type];
+
+  return <span className={mergedClassNames}>{content}</span>;
 };
--- a/packages/react-components/src/components/Badge/Badge.module.scss
+++ b/packages/react-components/src/components/Badge/Badge.module.scss
@@ -1,23 +1,100 @@
-.badge {
-  background-color: var(--color-negative-default);
-  border-radius: 18px;
+$base-class: 'badge';
+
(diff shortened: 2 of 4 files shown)


packages/react-components/src/components/Badge/Badge.tsx
```diff
@@ -3,22 +3,33 @@ import * as React from 'react';
 import styles from './Badge.module.scss';
 import cx from 'clsx';
 
+const baseClass = 'badge';
```
AlithAnar, line 6:
```suggestion
const baseClass = 'lc_badge';
```
?
I don't know if we should still apply `lc_` prefix, but I saw a comment in the other PR that mentioned that.

AlithAnar (approved):

packages/react-components/src/components/Badge/Badge.tsx
marcinsawicki: No. We don't use the `lc-` prefix in class names.

marcinsawicki: If we found any in other components, we should fix that.

packages/react-components/src/components/Badge/Badge.tsx
```diff
@@ -3,22 +3,33 @@ import * as React from 'react';
...
-    </span>
+  const mergedClassNames = cx(
+    className,
+    styles[baseClass],
+    styles[`${baseClass}--${kind}`],
+    styles[`${baseClass}--${size}`]
   );
+
+  const content = {
+    ['content']: children,
```
marcinsawicki, line 29: I know that we want to give the possibility to put our own content here, but I wonder if we should not make sure that too much content does not stretch this badge.
badge-size

marcinsawicki (approved):
Looks fine for me, except for one comment related to stretching the badge size in a case where someone passes too much content to the component.

packages/react-components/src/components/Badge/Badge.tsx
sgraczyk: @AlithAnar AFAIR we're avoiding prefixes in order to avoid conflicts with the legacy version of the Design System. I remember also there should be a possibility to apply a custom prefix globally, but I'm not sure if it was implemented.

packages/react-components/src/components/Badge/Badge.tsx
sgraczyk: Figma designs provide such examples 
image
so I believe `Badge` should be `stretchable`. The question is if we want to support dedicated handling for `Number` and automatically trim above 99 with `99+`.

mreszke (approved):

PR #357, opened 2022-07-26
#338 Changes in Picker component related to enabling the multiselect option

### LINKED ISSUE #338: [Picker] Enable multi-select
## Description
`Picker` component should be configurable to add the possibility to select multiple elements from the list.

## Design
https://www.figma.com/file/2pFu80PXO5A2tfyrAGnx91/Product-Components?node-id=6122%3A61558

marcinsawicki:
Resolves #338 

## Description
After releasing the `Picker` component as a new approach after `Select`, we agreed that this component should be configurable to enable the multiselect feature.
That requires some changes in the component logic, to unify the "singleselect" and "multiselect" mode.

- Added `multiselect` prop to enable this option
- Added possibility to pass own class name
- Changed the styles of components, enabling the 100% width, to make it easier to fit with own design

## Storybook
- [Default](https://613a8e945a5665003a05113b-sodbtivsyz.chromatic.com/?path=/story/components-picker--picker)


packages/react-components/src/components/Icon/Icon.module.scss
```diff
@@ -2,6 +2,10 @@
   align-items: center;
   display: flex;
 
+  svg {
+    pointer-events: none;
+  }
```
sgraczyk, line 7: Can you elaborate on this one?

packages/react-components/src/components/Picker/Picker.module.scss
```diff
@@ -33,4 +34,9 @@ $base-class: 'picker';
       margin-right: 5px;
     }
   }
+
+  &__tag {
+    margin-bottom: 4px;
+    margin-right: 4px;
```
sgraczyk, line 40: How about this? Or we want to avoid declaring 0 margin for remaining directions?
```suggestion
    margin: 0 4px 4px 0;
```

packages/react-components/src/components/Picker/Picker.spec.tsx
```diff
@@ -7,24 +7,31 @@ import { IPickerProps, Picker } from './Picker';
 // eslint-disable-next-line @typescript-eslint/no-empty-function
```
sgraczyk, line 7: I would expect some test that would make sure that a `single-select` is able to pick only one value (eg. by clicking two different options and making sure only the last one is triggered in `onSelect`. Another idea is making sure that `unselect` is working. 

In general I believe we could come up with more testing cases.

packages/react-components/src/components/Picker/Picker.stories.tsx
```diff
@@ -111,12 +110,49 @@ PickerWithDisabledSearch.args = {
...
+    { key: 'groupA', name: 'Group A title header', groupHeader: true },
+    { key: 'one', name: 'Option one' },
+    { key: 'two', name: 'Option two' },
+    { key: 'three', name: 'Option three' },
+    { key: 'groupB', name: 'Group B title header', groupHeader: true },
+    { key: 'four', name: 'Option four' },
+    { key: 'five', name: 'Option five' },
+    { key: 'six', name: 'Option six', disabled: true },
+    { key: 'seven', name: 'Option seven', disabled: true },
+  ],
```
sgraczyk, line 128: Have you considered moving to a common variable, or do we think it won't work well in Storybook docs?

packages/react-components/src/components/Picker/Picker.tsx
```diff
@@ -6,44 +6,52 @@ import { Trigger, TriggerSize } from './Trigger';
...
   label?: string;
   error?: string;
   options: IPickerListItem[];
-  selectedOption?: IPickerListItem;
+  selectedOptions?: IPickerListItem[];
   size?: TriggerSize;
   placeholder?: string;
   isRequired?: boolean;
   noSearchResultText?: string;
+  multiselect?: boolean;
```
sgraczyk, line 26: I'd avoid the `boolean` type in favor of `type` like `mode/type` with values of `single | multi`.
```suggestion
  multiselect?: boolean;
```

packages/react-components/src/components/Picker/Picker.tsx
```diff
@@ -6,44 +6,52 @@ import { Trigger, TriggerSize } from './Trigger';
...
 const baseClass = 'picker';
 
 export interface IPickerProps {
+  className?: string;
   disabled?: boolean;
   label?: string;
   error?: string;
   options: IPickerListItem[];
-  selectedOption?: IPickerListItem;
+  selectedOptions?: IPickerListItem[];
```
sgraczyk, line 21: maybe `selected` is enough for the name of the `prop`?

packages/react-components/src/components/Picker/Picker.tsx
```diff
@@ -70,32 +78,96 @@ export const Picker: React.FC<IPickerProps> = ({
...
+    setSelectedItems(newSelectedItems);
+  };
+
+  const handleOnSelectAll = () => {
     setIsListOpen(false);
-    setSelectedItem(item);
-    onSelect(item);
+
+    const itemsToSelect = items.filter(
+      (item) => !item.disabled && !item.groupHeader && item.key !== 'select-all'
```
sgraczyk, line 120: Consider placing `select-all` in a variable

sgraczyk: Additionally, I'd think about a helper function `isItemSelectable` that would move this condition outside filter

packages/react-components/src/components/Picker/Picker.tsx
```diff
@@ -70,32 +78,96 @@ export const Picker: React.FC<IPickerProps> = ({
...
-    onSelect(null);
+    setSelectedItems(null);
   };
 
   const handleOnFilter = (text: string) => setSearchPhrase(text);
 
+  const handleItemRemove = (item: IPickerListItem) => {
+    const newSelectedItems = selectedItems
+      ? selectedItems.filter((selectedItem) => selectedItem !== item)
+      : null;
```
sgraczyk, line 136: Why `null` and not `empty array`?

packages/react-components/src/components/Picker/Picker.tsx
```diff
@@ -70,32 +78,96 @@ export const Picker: React.FC<IPickerProps> = ({
...
+              dismissible
+              onRemove={() => handleItemRemove(item)}
+            >
+              {item.name}
+            </Tag>
+          );
+        })}
+      </div>
+    );
+  };
```
sgraczyk, line 166: It's more of a component than an internal helper function. Consider extracting.

packages/react-components/src/components/Picker/Picker.tsx
```diff
@@ -129,23 +201,28 @@ export const Picker: React.FC<IPickerProps> = ({
...
           onClearClick={handleOnClearClick}
           onFilter={handleOnFilter}
         >
-          {selectedItem ? selectedItem.name : placeholder}
+          {selectedItems ? getSelectedItems(selectedItems) : placeholder}
         </Trigger>
         <PickerList
-          selectedItem={selectedItem}
+          selectedItemsKeys={
+            selectedItems ? getSelectedItemsKeys(selectedItems) : null
```
sgraczyk, line 216: And again, is operating on `null` necessary. `undefined` or empty array would probably be enough and would bring fewer issues in the code.

packages/react-components/src/components/Picker/PickerList.tsx
```diff
@@ -166,6 +187,29 @@ export const PickerList: React.FC<IPickerListProps> = ({
...
+              aria-current={currentItemKey === item.key}
+              id={item.key}
+              key={item.key}
+              className={cx(
+                styles[itemClassName],
+                styles[`${itemClassName}--select-all`]
+              )}
+              onClick={handleOnSelectAllClick}
+            >
+              Select all
```
sgraczyk, line 208: Could be a const

sgraczyk: Additionally, I believe developers should have control over this text. For example if used in non-English environment

packages/react-components/src/components/Picker/Trigger.tsx
```diff
@@ -20,8 +20,9 @@ export interface ITriggerProps {
   isItemSelected: boolean;
   isOpen: boolean;
   isRequired?: boolean;
+  isMultiSelect?: boolean;
   size?: TriggerSize;
-  onClick: () => void;
+  onClick: (e: React.MouseEvent | KeyboardEvent) => void;
```
sgraczyk, line 25: I'm not sure if it's a proper approach to handle both mouse and keyboard events using the `click` name. Either consider `onTrigger` or handle two separate events, as it's misleading.

packages/react-components/src/components/Tag/Tag.tsx
```diff
@@ -118,6 +118,7 @@ export const Tag: React.FC<TagProps> = ({
           className={styles[`${baseClass}__remove`]}
         >
           <Icon
+            data-dismiss-icon
```
sgraczyk, line 121: Why it's needed?

packages/react-components/src/components/Picker/Picker.spec.tsx
```diff
@@ -67,18 +76,56 @@ describe('<Picker> component', () => {
...
+      {
+        key: 'three',
+        name: 'Option three',
+      },
+    ]);
     userEvent.click(getByTestId('picker-trigger__clear-icon'));
     expect(onSelect).toHaveBeenCalledWith(null);
   });
 
+  it('should call onSelect with all correct elements in multiselect mode if "Select all" option is chosen', () => {
```
mreszke, line 89: I'm missing the test which allow to check the same thing but picking options one by one

packages/react-components/src/components/Picker/Picker.tsx
```diff
@@ -70,32 +78,96 @@ export const Picker: React.FC<IPickerProps> = ({
     }
   }, [isListOpen]);
 
-  const handleOnTriggerClick = () => {
-    if (disabled) {
+  React.useEffect(() => onSelect(selectedItems), [selectedItems]);
```
mreszke, line 81: can't we call `onSelect` in event handlers? And do we really need the internal state of the selected items?

packages/react-components/src/components/Tag/Tag.tsx
marcinsawicki: It's for preventing the list open in the case where a user wants to remove the selected item in multipicker mode.
[here](https://github.com/livechat/design-system/pull/357/files#diff-f098634dfe943b4c32066c5aeb6c944f3ffb3afd62308e62a43be6d6bcfd33b5R86)

packages/react-components/src/components/Icon/Icon.module.scss
marcinsawicki: To prevent the list open when the user is clicking on the selected item tag icon (multipicker mode), we need to check the event target. This prop will prevent the event handler to see svg as the target object.

packages/react-components/src/components/Picker/Picker.tsx
marcinsawicki: It's because imo is better to check `!selectedItems` then `selectedItems.length !== 0`.

packages/react-components/src/components/Picker/Picker.tsx
marcinsawicki: As above

packages/react-components/src/components/Picker/Picker.tsx
marcinsawicki: After some analysis, no. I've removed the internal state for `selectedItems` and from now, the component is fully controlled by external props passed inside.

kstvsko (changes requested):
1. When you chose something and the picker is opened you should see the chosen values in the picker field, not the placeholder.
image

2. When you chose something the icon that clears the picker should look differently.
image
3. On the default state there is something weird near the picker icon. 
image
4. When you select something and then deselect it and stop focusing on the picker the icon to clear the picker stays visible and there is no placeholder text.
image

TomaszRatajczyk (approved):

packages/react-components/src/components/Picker/Picker.tsx
```diff
@@ -70,32 +77,72 @@ export const Picker: React.FC<IPickerProps> = ({
     }
   }, [isListOpen]);
 
-  const handleOnTriggerClick = () => {
-    if (disabled) {
+  const handleOnTrigger = (e: React.MouseEvent | KeyboardEvent) => {
```
sgraczyk, line 80:
```suggestion
  const handleTrigger = (e: React.MouseEvent | KeyboardEvent) => {
```

packages/react-components/src/components/Picker/Picker.tsx
```diff
@@ -70,32 +77,72 @@ export const Picker: React.FC<IPickerProps> = ({
...
 
-    return setIsListOpen((prev) => !prev);
+    setIsListOpen((prev) => !prev);
   };
 
   const handleOnClose = () => {
     setIsListOpen(false);
   };
 
   const handleOnSelect = (item: IPickerListItem) => {
```
sgraczyk, line 94:
```suggestion
  const handleSelect = (item: IPickerListItem) => {
```

packages/react-components/src/components/Picker/Picker.tsx
```diff
@@ -70,32 +77,72 @@ export const Picker: React.FC<IPickerProps> = ({
...
-    setSelectedItem(item);
-    onSelect(item);
+
+    const itemsToSelect = items.filter(isItemSelectable);
+
+    onSelect(itemsToSelect);
   };
 
-  const handleOnClearClick = () => {
+  const handleOnClear = () => {
```
sgraczyk, line 124:
```suggestion
  const handleClear = () => {
```

packages/react-components/src/components/Picker/Picker.tsx
```diff
@@ -129,23 +176,33 @@ export const Picker: React.FC<IPickerProps> = ({
...
+          isMultiSelect={type === 'multi'}
           size={size}
-          onClick={handleOnTriggerClick}
-          onClearClick={handleOnClearClick}
+          onTrigger={handleOnTrigger}
+          onClear={handleOnClear}
           onFilter={handleOnFilter}
         >
-          {selectedItem ? selectedItem.name : placeholder}
+          {selected ? (
```
sgraczyk, line 187: It will trigger also for empty array, is this correct here?

packages/react-components/src/components/Picker/Trigger.spec.tsx
```diff
@@ -84,7 +84,7 @@ describe('<Trigger> component', () => {
     const { getByTestId } = renderComponent({
       ...defaultProps,
       isItemSelected: true,
-      onClearClick,
+      onClear: onClearClick,
```
sgraczyk, line 87: Is this expected?

packages/react-components/src/components/Picker/Trigger.spec.tsx
```diff
@@ -72,7 +72,7 @@ describe('<Trigger> component', () => {
     const onClick = vi.fn();
     const { getByText } = renderComponent({
       ...defaultProps,
-      onClick,
+      onTrigger: onClick,
```
sgraczyk, line 75: Is this expected?

sgraczyk (approved):

packages/react-components/src/components/Picker/TriggerBody.tsx
```diff
@@ -0,0 +1,42 @@
...
+  type: PickerType;
+  onItemRemove: (item: IPickerListItem) => void;
+}
+
+export const TriggerBody: React.FC<ITriggerBodyProps> = ({
+  items,
+  type,
+  onItemRemove,
+}) => {
+  if (type === 'single') {
```
mreszke, line 22: isn't better to have 2 separate components, one for `single picker` and second one for `multiple picker`?

packages/react-components/src/components/Picker/Picker.tsx
marcinsawicki: Good catch. After we removed the initial state of selected items, we could expect that an empty array can be passed. I will add an additional check here.

packages/react-components/src/components/Picker/Picker.tsx
```diff
@@ -180,29 +180,27 @@ export const Picker: React.FC<IPickerProps> = ({
...
+            isSearchDisabled={searchDisabled}
+            placeholder={placeholder}
+            items={selected}
+            type={type}
+            onItemRemove={handleItemRemove}
+            onFilter={handleOnFilter}
+          />
         </Trigger>
         <PickerList
           selectedItemsKeys={selected ? getSelectedItemsKeys(selected) : null}
```
sgraczyk, line 197: I think we can move this whole condition inside a memoized array `selectedItemKeys`. You can use `useMemo` here.

packages/react-components/src/components/Picker/Picker.tsx
```diff
@@ -70,32 +77,72 @@ export const Picker: React.FC<IPickerProps> = ({
...
+
+    const newSelectedItems = currentSelectedItems.includes(item)
+      ? currentSelectedItems.filter((selectedItem) => selectedItem !== item)
+      : currentSelectedItems.concat(item);
+
+    onSelect(newSelectedItems);
+  };
+
+  const isItemSelectable = (item: IPickerListItem) =>
+    !item.disabled && !item.groupHeader && item.key !== 'select-all';
```
sgraczyk, line 114: Consider having `'select-all'` in a shared constant

packages/react-components/src/components/Picker/Picker.tsx
```diff
@@ -70,32 +77,72 @@ export const Picker: React.FC<IPickerProps> = ({
...
+
+    if (newSelectedItems?.length === 0) {
+      return onSelect(null);
+    }
+
+    onSelect(newSelectedItems);
+  };
+
+  const getSelectedItemsKeys = (items: IPickerListItem[]) =>
+    items ? items.map((item) => item.key) : null;
```
sgraczyk, line 144: Consider the following approach
```suggestion
  const selectedItemsKeys = useMemo(() => {
    if (!selected || !items) {
      return null;
    }
    
    return items.map((item) => item.key);
  }, [selected, items]);
```

sgraczyk (approved):

kstvsko (approved):

jpyzio123:
1. Picker with `multiselect` set to `true` has different height CSS than picker with it set to false. Non-multiselect picker has specific height values depending on `size` property, while multiselect picker has the `height` set to `auto` 
https://user-images.githubusercontent.com/46003125/198558578-f423ad95-7552-41d0-9d87-6600d0e236b5.mp4 

2. Option to `Select all` is still visible and clickable with `multiselect` set to false, I think it shouldn't be. 
https://user-images.githubusercontent.com/46003125/198559669-a206b8bc-733a-412e-8997-42a2bec0a90c.mp4

jpyzio123:
in type: multi, the option `selectAllOption` set to False is still visible and possible to add as regular option
image

jpyzio123 (approved):

Issue #343, opened 2022-07-14
[Picker] List items should have possibility to put own component there


marcinsawicki:
## Description
Currently, the `Picker` only takes the item's name to select and displays plain text in the items list.
This functionality should be expanded to allow displaying the component as a list item (e.g. an avatar with the username).
The same functionality must be available in multi-select mode, but components should be displayed also as selected item tags.

PR #362, opened 2022-07-27
#343 Extend the Picker options interface and logic to enable passing the component as list item

marcinsawicki:
Resolves: #343 

## Description
Some changes related to the interface and logic of `Trigger` and `PickerList`, enabling the possibility to pass the own component as list item.

## Checklist

**Obligatory:**

- [ ] Self-review
- [ ] Storybook cases
- [ ] Design review
- [ ] Functional (QA) review

**Optional:**

### DIFF
--- a/packages/react-components/src/components/Picker/PickerList.tsx
+++ b/packages/react-components/src/components/Picker/PickerList.tsx
@@ -12,6 +12,7 @@ const itemClassName = `${baseClass}__item`;
 export interface IPickerListItem {
   key: string;
   name: string;
+  customElement?: React.ReactElement;
   groupHeader?: boolean;
   disabled?: boolean;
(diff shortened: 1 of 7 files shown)


packages/react-components/src/components/Picker/PickerList.tsx
```diff
@@ -143,6 +144,18 @@ export const PickerList: React.FC<IPickerListProps> = ({
...
+    if (item.customElement) {
+      return (
+        <div className={styles[`${itemClassName}__custom`]}>
+          {item.customElement}
+        </div>
+      );
+    }
+
+    return item.name;
+  };
```
sgraczyk, line 157: It's more of a component. Consider extracting.

Issue #339, opened 2022-07-01
[Input][TagInput] Missing background design tokens


sgraczyk:
## Describe the bug
Mentioned components do not apply different colors to the background while switching themes.
We need to add a token for background (`--surface-basic`)

## Steps to reproduce
https://www.loom.com/share/252f553e2a9247da90890deff2e97b73

## Expected behavior
`--surface-basic` token applied for `TagInput` and `Input` components on theme change

## Screenshots
https://www.loom.com/share/252f553e2a9247da90890deff2e97b73

## Environment
 - Design system package(s) version: `@livechat/design-system-react-components@1.0.0-beta.14`

## Additional context
n/a

sgraczyk:
Released in [v1.0.0-alpha.18](https://github.com/livechat/design-system/releases/tag/v1.0.0-alpha.18)

PR #364, opened 2022-07-27
#339 Apply missing tokens to Input and TagInput components

sgraczyk:
Resolves: #339 

## Description
Initially, the task assumed only adding missing background color tokens, however, I've also found out it requires some work around stories and other styles. Here's what changed:

* Added background and placeholder color tokens for `Input`
* Added background and disabled color tokens for `TagInput`
* Added `disabled` prop to `InputProps`
* Reorganized stories to cover sizes and states of Input components

## Storybook
* [Chromatic UI review](https://www.chromatic.com/build?appId=613a8e945a5665003a05113b&number=514)
* [Input](https://613a8e945a5665003a05113b-mtkyesnbym.chromatic.com/?path=/story/forms-input--default)
* [TagInput](https://613a8e945a5665003a05113b-mtkyesnbym.chromatic.com/?path=/story/forms-tag-input--default-tag-input)

## Checklist

**Obligatory:**

- [x] Self-review
- [x] Storybook cases
- [ ] Design review
- [ ] Functional (QA) review

**Optional:**

### DIFF
--- a/packages/react-components/src/components/TagInput/TagInput.stories.tsx
+++ b/packages/react-components/src/components/TagInput/TagInput.stories.tsx
@@ -1,16 +1,21 @@
 import * as React from 'react';
-import { ComponentMeta } from '@storybook/react';
+import { ComponentMeta, Story } from '@storybook/react';
+
+import noop from '../../utils/noop';
+import { StoryDescriptor } from '../../stories/components/StoryDescriptor';
 
 import {
-  TagInput as TagInputComponent,
+  TagInput,
   TagInputProps,
-  EmailTagInput as EmailTagInputComponent,
+  EmailTagInput,
   EmailTagInputProps,
 } from './index';
 
+const placeholderText = 'Placeholder text';
+
 export default {
   title: 'Forms/Tag Input',
-  component: TagInputComponent,
+  component: TagInput,
   argTypes: {
     tags: {
       control: {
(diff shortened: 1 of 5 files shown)


packages/react-components/src/components/TagInput/TagInput.stories.tsx
```diff
@@ -1,49 +1,68 @@
...
   return (
     <div>
-      <EmailTagInputComponent {...args} tags={tags} onChange={setTags} />
+      <EmailTagInput {...args} tags={tags} onChange={setTags} />
     </div>
   );
 };
-
-EmailTagInput.args = {
+DefaultEmailTagInput.storyName = 'EmailTagInput';
```
marcinsawicki, line 54: Shouldn't we use spaces in the story name?

marcinsawicki (approved):

packages/react-components/src/components/TagInput/TagInput.stories.tsx
sgraczyk: The default story name is simply the name of the component. I'm using this workaround to avoid the unnecessary rename of the imported component from `EmailTagInput` to `EmailTagInputComponent`. It's better this way as the `Docs` section in Storybook is not polluted with the `Component` suffix.

AlithAnar (approved):

mreszke (approved):

Issue #345, opened 2022-07-19
[Design Tokens] Spacing


sgraczyk:
## Description
Introduce `spacing` tokens as `CSS variables`.

## Design
[Figma - Spacing](https://www.figma.com/file/2pFu80PXO5A2tfyrAGnx91/Product-Components?node-id=6020%3A47693)

sgraczyk:
Released in [v1.0.0-alpha.18](https://github.com/livechat/design-system/releases/tag/v1.0.0-alpha.18)

PR #367, opened 2022-07-27
#345 Spacing variables for the css usage

marcinsawicki:
Resolves: #345 

## Description
Scss variables of spacing, to the use in styling for components.

## Storybook
- [Storybook](https://613a8e945a5665003a05113b-czjyoqmqne.chromatic.com/?path=/story/foundations-spacing--page)

## Checklist

**Obligatory:**

- [x] Self-review
- [x] Storybook cases
- [ ] Design review
- [ ] Functional (QA) review

**Optional:**


sgraczyk:
Why aren't we using `CSS variables` for that, or at least both with SCSS?

packages/react-components/.storybook/preview.js
```diff
@@ -1,6 +1,7 @@
 import '../src/themes/legacy.scss';
 import '../src/themes/light.scss';
 import '../src/themes/dark.scss';
+import '../src/spacing/spacing.scss';
```
sgraczyk, line 4: Do we need `scss` at all? Using `CSS variables` feels convenient

sgraczyk: Ok, nevermind it's just an extension of file, it still uses CSS vars

packages/react-components/src/index.scss
```diff
@@ -1,4 +1,7 @@
 @use './themes/legacy';
+@use './themes/light';
+@use './themes/dark';
+@use './spacing/spacing';
```
sgraczyk, line 4: how about?:
```suggestion
@use './foundations/spacing';
```

packages/react-components/src/index.ts
```diff
@@ -1,6 +1,7 @@
 import './index.scss';
 
 export { DesignTokens } from './themes/designTokens';
+export { SpacingTokens } from './spacing/spacingTokens';
```
sgraczyk, line 4: Enumerations (even realized as consts) should have a singular form.
```suggestion
export { SpacingToken } from './foundation/spacing-tokens';
```

packages/react-components/src/index.ts
```diff
@@ -1,6 +1,7 @@
 import './index.scss';
 
 export { DesignTokens } from './themes/designTokens';
```
sgraczyk, line 3: We could fix this also
```suggestion
export { DesignToken } from './themes/design-oken';
```

packages/react-components/src/spacing/spacing.scss
```diff
@@ -0,0 +1,16 @@
+:root {
```
sgraczyk, line 1: @kstvsko @martagranisz Are we able to confirm that `spacing` tokens should be `theme-agnostic`? What I mean is if we are sure that there will be no `theme` that could introduce different spacings?

packages/react-components/src/spacing/spacingTokens.ts
```diff
@@ -0,0 +1,16 @@
+export const SpacingTokens = {
```
sgraczyk, line 1: Alternatively `SPACING_TOKENS`, hard to say what's our approach here, as it seems more of a `enumeration` than a set of data.
```suggestion
export const SpacingToken = {
```

packages/react-components/.storybook/preview.js
marcinsawicki: 👍

packages/react-components/src/index.ts
marcinsawicki: 👍 But we need to remember to include this as (small) breaking change in release docs, as we use `DesignTokens` enum in another app.

packages/react-components/src/spacing/spacing.scss
sgraczyk: For now we assumed spacings are constant for all themes

sgraczyk (approved):

kstvsko (approved):
I think everything is right, my only concern is the **Examples** in the Foundations/Spacing. I think we should address this during our new approach to the documentation to be more descriptive and clear, I will note it down.

Issue #342, opened 2022-07-01
[Input] Add missing functionalities and enable possibility the type of input


sgraczyk:
## Description
The current `Input` implementation is missing:
- Input with icon option

Also, the `Input` could provide a native password input that would handle at least two functionalities:
- value is hidden by default
- the ability to show password when eg. visibility icon is clicked

## Design
https://www.figma.com/file/2pFu80PXO5A2tfyrAGnx91/Product-Components?node-id=6%3A18

PR #369, opened 2022-07-28
#342 Support type prop to set the type of input, adjusted styles to fit with designs, added Icon functionality

marcinsawicki:
Resolves: #342 

## Description
The `Input` component is missing some functionalities after migration from the old implementation. Also, we found that the component has incorrect or missing styles.
What's changed:
- Added support for native `type` prop to enable switching between `text` and `password` mode, where `text` is the default
- In `password` mode, added an icon to switch between password visibility
- Added `icon` prop for passing the component, which will be displayed as decoration in input
- Refactored code and styles to fit with the designs
- Changes `size` prop to `kind` to set the input size, because `size` is reserved in native props and has a different desitnation

## Storybook
- [Storybook](https://613a8e945a5665003a05113b-eeawbyvcqc.chromatic.com/?path=/story/forms-input--kinds)

## Checklist

**Obligatory:**

- [x] Self-review
- [x] Storybook cases
- [x] Design review
- [x] Functional (QA) review

**Optional:**


packages/react-components/src/components/Input/Input.stories.tsx
```diff
@@ -1,15 +1,20 @@
 import * as React from 'react';
 import { ComponentMeta, Story } from '@storybook/react';
 
-import { StoryDescriptor } from '../../stories/components/StoryDescriptor';
+import * as MaterialIcons from '@livechat/design-system-icons/react/material';
```
sgraczyk, line 4: You can remove the line break from `3`

packages/react-components/src/components/Input/Input.stories.tsx
```diff
@@ -49,3 +51,23 @@ export const States = (): JSX.Element => (
     </StoryDescriptor>
   </>
 );
+
+export const Kinds = (): JSX.Element => (
+  <>
+    <StoryDescriptor title="Text">
+      <Input placeholder={placeholderText} />
```
sgraczyk, line 58: While defining specific variants, we should not rely on `default` values as it's intended to include prop value for docs

kstvsko (approved):
As for now, it is compliant with the designs. The only thing I would rethink is if we want to have a hover state on the clickable icons like the one for showing/hiding password. I will take care of the topic and bring a new issue if needed.

packages/react-components/src/components/Input/Input.stories.tsx
```diff
@@ -49,3 +50,23 @@ export const States = (): JSX.Element => (
...
+      <Input kind="text" placeholder={placeholderText} />
+    </StoryDescriptor>
+    <StoryDescriptor title="Password">
+      <Input kind="password" placeholder={placeholderText} />
+    </StoryDescriptor>
+  </>
+);
+
+export const WithIcon = (): JSX.Element => (
+  <>
```
sgraczyk, line 66: `React.Fragment` is not necessary here

packages/react-components/src/components/Input/Input.tsx
```diff
@@ -1,34 +1,91 @@
...
+import { Icon } from '../Icon';
 
-type InputSize = 'xsmall' | 'small' | 'medium' | 'large';
+type InputSize = 'compact' | 'medium' | 'large';
 
 export interface InputProps extends React.HTMLAttributes<HTMLInputElement> {
   size?: InputSize | undefined;
   error?: boolean | undefined;
   disabled?: boolean | undefined;
+  kind?: 'text' | 'password';
```
sgraczyk, line 19: To be honest, I'm not sure if it's a good approach. We already extend `React.HTMLAttributes<HTMLInputElement>` which has a prop called `type` And it should handle `password` type -> https://www.w3schools.com/html/html_form_input_types.asp

MateuszRomek: ☝️

kstvsko:
@sgraczyk @MichalPaszowski @marcinsawicki Any idea how the disabled input with icon looks? I had to fix that issue in Figma and I also wanted to take care of it here, but I cannot see the disabled input with the icon on Storybook.

packages/react-components/src/components/Input/Input.tsx
```diff
@@ -1,34 +1,90 @@
...
+import { Icon } from '../Icon';
 
-type InputSize = 'xsmall' | 'small' | 'medium' | 'large';
+type InputSize = 'compact' | 'medium' | 'large';
 
-export interface InputProps extends React.HTMLAttributes<HTMLInputElement> {
-  size?: InputSize | undefined;
+export interface InputProps
+  extends React.InputHTMLAttributes<HTMLInputElement> {
+  kind?: InputSize | undefined;
```
MichalPaszowski, line 17: why do you need to `| undefined` in this and the next three lines

MichalPaszowski (changes requested):

packages/react-components/src/components/Input/Input.spec.tsx
```diff
@@ -15,28 +15,59 @@ describe('<Input> component', () => {
...
   });
 
   it('should have error class on error', () => {
     const { container } = render(<Input error />);
     expect(container.firstChild).toHaveClass(styles['input--error']);
   });
+
+  it('should have disabled class and input should be disabled if "disabled" prop is set', () => {
+    const { container, getByTestId } = render(<Input disabled />);
+    expect(container.firstChild).toHaveClass(styles['input--disabled']);
```
sgraczyk, line 40: Maybe it should also have `aria-disabled`?

packages/react-components/src/components/Input/Input.stories.tsx
sgraczyk: Bump ☝🏼

packages/react-components/src/components/Input/Input.tsx
```diff
@@ -1,34 +1,90 @@
...
-type InputSize = 'xsmall' | 'small' | 'medium' | 'large';
+type InputSize = 'compact' | 'medium' | 'large';
 
-export interface InputProps extends React.HTMLAttributes<HTMLInputElement> {
-  size?: InputSize | undefined;
-  error?: boolean | undefined;
-  disabled?: boolean | undefined;
+export interface InputProps
+  extends React.InputHTMLAttributes<HTMLInputElement> {
+  wrapperSize?: InputSize;
```
sgraczyk, line 17: Maybe `inputSize`. I'm not sure if `wrapper` is easy to understand.

packages/react-components/src/components/Input/Input.tsx
```diff
@@ -1,34 +1,90 @@
...
+          disabled={disabled}
+          type={type && !isPasswordVisible ? type : 'text'}
+        />
+        {type === 'password' && (
+          <Button
+            disabled={disabled}
+            kind="plain"
+            icon={
+              <Icon
+                customColor={
```
sgraczyk, line 74: I would move these conditions to constants

packages/react-components/src/components/Input/Input.tsx
```diff
@@ -1,34 +1,90 @@
...
+                  !disabled
+                    ? 'var(--content-default)'
+                    : 'var(--content-disabled)'
+                }
+                source={
+                  !isPasswordVisible ? VisibilityOnIcon : VisibilityOffIcon
```
sgraczyk, line 80: As above, the target is to make JSX as simple as possible

sgraczyk (approved):

packages/react-components/src/components/Input/Input.tsx
MateuszRomek: I definitely agree with that. `inputSize` feels much better

packages/react-components/src/components/Input/Input.tsx
```diff
@@ -1,34 +1,90 @@
...
+  VisibilityOn as VisibilityOnIcon,
+  VisibilityOff as VisibilityOffIcon,
+} from '@livechat/design-system-icons/react/material';
+
 import styles from './Input.module.scss';
+import { Button } from '../Button';
+import { Icon } from '../Icon';
 
-type InputSize = 'xsmall' | 'small' | 'medium' | 'large';
+type InputSize = 'compact' | 'medium' | 'large';
```
MateuszRomek, line 13: For me it could be a generic `size` type lifted up in folders structure as we are slowly starting to repeat ourselves 

Screenshot 2022-11-15 at 08 42 54

MateuszRomek (approved):

packages/react-components/src/components/Popover/Popover.module.scss
```diff
@@ -1,13 +1,13 @@
...
 .popover {
-  background-color: var(--surface-basic-default);
-  border-radius: 4px;
-  box-shadow: 0 1px 10px 0 rgba(66, 77, 87, 0.3);
   display: none;
-  max-width: 336px;
-  min-width: 168px;
   z-index: $stacking-context-level-popover;
+  border-radius: 4px;
+  box-shadow: 0 1px 10px 0 rgb(66 77 87 / 30%);
```
sgraczyk, line 7: How about a shadow color from tokens here?

sgraczyk (approved):

jpyzio123:
mission icon options to put icon to the right side or two icons on both sides as in designs 
image

jpyzio123 (approved):

Issue #371, opened 2022-07-28
[Switch] Add loading state, medium size and slightly refresh the UI


MrVorgoth:
Refactor of switch component:
- adding `loading` state,
- adding `medium` size,
- visual refresh of component

Design: https://www.figma.com/file/2pFu80PXO5A2tfyrAGnx91/Product-Components?node-id=1881%3A156

sgraczyk:
Released in [v1.0.0-alpha.19](https://github.com/livechat/design-system/releases/tag/v1.0.0-alpha.19)

PR #372, opened 2022-07-29
#371 Add loading state, medium size and slightly refresh the UI

MrVorgoth:
Resolves: https://github.com/livechat/design-system/issues/371

## Description
Refresh of switch component. What's changed?

Screenshot 2022-08-01 at 16 31 06

1) Added new icon states to Switch component:
- `loading` state
- `on` state
- `off` state

2) The loading state is always on top, when passed with other props (like disabled, which also adds icon). Loading state also needs to have `disabled` state, because we can't perform any action during loading

3) New `medium` size available in Switch component. I've also renamed the sizes:
- `basic` -> `large`
- `compact` -> `small` After discussion I can't do it as it's a standard to name it `compact`, so it'll stay as is
- `medium` (it's new size)
- so to sum up - the 3 sizes are now: `compact` / `medium` / `large`

## Storybook
- added `medium` sized switch
- added `loading` state of switch

## Checklist
**Obligatory:**
- [x] Self-review
- [x] Storybook cases
- [ ] Design review
- [x] Functional (QA) review

**Optional:**


MrVorgoth:
Note: I'm still waiting for any last-call changes from Marta. For sure the `X` icon on off switch should be changed or be deleted - it's only my idea of it

packages/react-components/src/components/Loader/Loader.stories.tsx
```diff
@@ -23,3 +23,13 @@ export const labeled = (args: LoaderProps): React.ReactElement => (
 export const customColors = (args: LoaderProps): React.ReactElement => (
   <Loader primaryColor="#d64646" secondaryColor="#eec4c5" {...args} />
 );
+
+export const coverSize = (args: LoaderProps): React.ReactElement => {
```
MateuszRomek, line 27:
```suggestion
export const coverSize = (args: LoaderProps): ReactElement => {
```

Maybe we can adjust imports from `react` because we don't need to import all in v17

packages/react-components/src/components/Loader/Loader.tsx
```diff
@@ -9,7 +9,7 @@ const baseClass = 'loader';
 const spinnerClass = `${baseClass}__spinner`;
 
 export interface LoaderProps extends React.HTMLAttributes<HTMLDivElement> {
-  size?: 'small' | 'medium' | 'large';
+  size?: 'small' | 'medium' | 'large' | 'cover';
```
MateuszRomek, line 12: For me we should introduce `fullWidth` prop instead of adding new `size`.
Size prop is generic and used in other components, why would we be able to choose more options for prop size here?

packages/react-components/src/components/Switch/Switch.tsx
```diff
@@ -1,19 +1,25 @@
...
 import noop from '../../utils/noop';
 
 import styles from './Switch.module.scss';
+import { Loader } from '../Loader';
 
-const baseClass = 'switch';
+export const baseClass = 'switch';
 
-export type SwitchSize = 'basic' | 'compact';
+export type SwitchSize = 'large' | 'medium' | 'compact';
```
MateuszRomek, line 17: From my perspective `Size` should be a common interface that is inherited by multiple components. However, this is not the time to change this in this PR

MateuszRomek (approved):

packages/react-components/src/components/Switch/Switch.tsx
MrVorgoth: Actually there was a discussion about it, and it should be fixed in the future (the issue should be in issues tab)

packages/react-components/src/components/Loader/Loader.tsx
MrVorgoth: Hmm, in <Loader /> component, `size` defaulted to `medium`. I wanted to provide another size prop, because it felt natural for me. So you're saying that adding another prop `fullWidth` is more benefitial instead of having another `size`?

MateuszRomek: This is my subjective opinion, but yes - we shouldn't be afraid of adding new props. For me, the size should be constant not that one component has 4 size options and the other has 3.
When looking at such MUI or Chakra UI they have clearly defined sizes of their components and some additional behaviors come in the form of new props.

If cover is going to take 100% width and 100% height then in this case maybe I would change my proposal for prop name to `cover` instead of `fullWidth` or any other name but I would not extend  size prop

packages/react-components/src/components/Loader/Loader.module.scss
```diff
@@ -14,6 +14,11 @@
   flex-direction: row;
   flex-wrap: nowrap;
 
+  &--cover {
```
sgraczyk, line 17: Is this naming convention fine for @martagranisz?

packages/react-components/src/components/Loader/Loader.tsx
sgraczyk: I also have a bit of mixed feelings about this one. Do we have a valid use case for Design System users? If we would like to include the ability for components to expand to full width we would need to support it for most of them. Personally, I feel it might not be that necessary. If we want it to handle the specific case within the DS itself I would simply implement it using CSS within the `Switch` component.

packages/react-components/src/components/Switch/Switch.spec.tsx
```diff
@@ -36,4 +37,32 @@ describe('Switch', () => {
...
   });
+
+  it('should be disabled if disabled is set to true', () => {
+    const { getByRole, getByTestId } = render(<Switch disabled={true} />);
+    const checkbox = getByRole('checkbox') as HTMLInputElement;
+    expect(checkbox.disabled).toEqual(true);
+    expect(getByTestId(`${baseClass}__disabled-icon`)).toBeVisible();
+  });
+
+  it('should have loader icon visible if loading is set to true', () => {
```
sgraczyk, line 48: You could also check for `disabled` state

packages/react-components/src/components/Switch/Switch.tsx
sgraczyk: Can we sort it ascending?

packages/react-components/src/components/Switch/Switch.tsx
```diff
@@ -42,9 +49,41 @@ export const Switch: React.FC<SwitchProps> = ({
     }
   }, [on]);
 
-  const iconSize: IconSize = size === 'basic' ? 'small' : 'xsmall';
+  const getIcon = (iconSize: IconSize): JSX.Element => {
```
sgraczyk, line 52: It should be converted into a component. Please avoid helpers returning `JSX.Element`.

packages/react-components/src/components/Switch/Switch.tsx
```diff
@@ -42,9 +49,41 @@ export const Switch: React.FC<SwitchProps> = ({
...
+    if (loading) {
+      return <Loader size="cover" />;
+    }
+
+    if (disabled) {
+      return (
+        <Icon
+          data-testid={`${baseClass}__disabled-icon`}
```
sgraczyk, line 60: Wouldn't classname be enough?

packages/react-components/src/components/Switch/Switch.tsx
```diff
@@ -42,9 +49,41 @@ export const Switch: React.FC<SwitchProps> = ({
...
+          source={LockIcon}
+          kind={'primary'}
+        />
+      );
+    }
+
+    if (on) {
+      return (
+        <Icon
+          className={styles[`${baseClass}__icon--on`]}
```
sgraczyk, line 71: There is an `inconsistency` ->
`disabled-icon` vs `icon--on`

packages/react-components/src/components/Switch/Switch.tsx
```diff
@@ -42,9 +49,41 @@ export const Switch: React.FC<SwitchProps> = ({
...
+          size={iconSize}
+          source={CheckIcon}
+          kind={'primary'}
+        />
+      );
+    }
+
+    return <Icon size={iconSize} source={CloseIcon} kind={'primary'} />;
```
sgraczyk, line 79: Why this one does not require a classname? I think you can handle them without separating them by the if clause. Just make sure the props provided are defined conditionally (icon and className)

packages/react-components/src/components/Switch/Switch.tsx
```diff
@@ -73,7 +112,7 @@ export const Switch: React.FC<SwitchProps> = ({
         checked={checked}
         name={name}
         ref={innerRef}
-        disabled={disabled}
+        disabled={shouldBehaveAsDisabled}
```
sgraczyk, line 115: What is the expected `pointer` in case of loading?

sgraczyk (changes requested):
I've left some comments to be applied before merging.

packages/react-components/src/components/Loader/Loader.tsx
MrVorgoth: Okay so you are saying to abandon `cover` size for `Loader` component. And instead we should provide our own loading Icon to the `Switch` component and make animation there and not use <Loader /> inside `Switch` at all?

sgraczyk: As discussed F2F I'd suggest using `Loader` component with custom styling within `Switch` for the loading state.

packages/react-components/src/components/Loader/Loader.module.scss
MrVorgoth: It's removed already

packages/react-components/src/components/Loader/Loader.tsx
MrVorgoth: Fixed - [1d02631](https://github.com/livechat/design-system/pull/372/commits/1d026314ab2284f8ecfd71d1988e81e0841a0383)

packages/react-components/src/components/Loader/Loader.stories.tsx
MrVorgoth: Cover size is removed

packages/react-components/src/components/Switch/Switch.tsx
MrVorgoth: Fixed it -> I've removed the incon--on (as we don't need it for additional styling of icon inside). Moreover I've renamed the data-testid to just `disabled-icon`

packages/react-components/src/components/Switch/Switch.tsx
MrVorgoth: I tried it with `className` before, but it's hard to query elements for it. Got 3 reasons for that:
1) If you want to check if the element is not in the DOM then `getElement..` returns undefined, because it can't find such element and you don't have 100% sure if you didn't mess up your test (made a typo or something)
2) Now the other method - `queryBy..` really does return information (even if element is not in the DOM), but I can't seem to use it on classNames
3) I want it to be unique so I want to use specific id for that

packages/react-components/src/components/Switch/Switch.tsx
MrVorgoth: Removed this helper - we have only 2 icons for now and I used simpler approach

packages/react-components/src/components/Switch/Switch.tsx
MrVorgoth: ☝️ it's also removed (there won't be any icon for `off` state)

packages/react-components/src/components/Switch/Switch.tsx
```diff
@@ -92,8 +96,14 @@ export const Switch: React.FC<SwitchProps> = ({
             styles[`${baseClass}__slider--${size}--${toggleStyles}`]
           )}
         >
-          {disabled && (
-            <Icon size={iconSize} source={LockIcon} kind={'primary'} />
+          {loading && <Loader className={styles[`${baseClass}__loader`]} />}
+          {disabled && !loading && (
```
sgraczyk, line 100: Move this condition to a constant

sgraczyk (approved):

sgraczyk:
I'm not sure if it looks well for dark theme. It might be old issue, but lock icon is not visible for dark theme.

https://user-images.githubusercontent.com/8076673/182609716-855b8f53-189e-442b-86e3-bccd80560012.mov

sgraczyk:
The design (size of loader) do not match the actual implementation:
image
vs
image

sgraczyk (changes requested):
Requesting changes due to design issues

MrVorgoth:
According to design indeed the padding inside loading icon is 1.5px for compact / 2px for medium and 3px for large switch component. I've adjusted it here: [363f347](https://github.com/livechat/design-system/pull/372/commits/363f3471e7efc4730751efe745375e7b1957b61f)

MichalPaszowski (approved):

MichalPaszowski:
In compact size distance between the edge of the container and the switch is a bit off. 
distance_edge

MrVorgoth:
> In compact size distance between the edge of the container and the switch is a bit off. distance_edge

Hey, nice catch! I've adjusted it in new commit - [62c5ae8](https://github.com/livechat/design-system/pull/372/commits/62c5ae806fa9b51f4f9aaf57df6e23b0b3b7d682)

Also in every design the inner circle of loader has border-width of 2px. I could acheive that by just passing prop `size="small"' to Loader component, but it seemed unnatural for me (maybe it isn't). Instead I've decided to target that element with CSS and set border-width to 2px

MrVorgoth:
> Looks fine now. Could you please report an issue with `lock icon invisible` in dark theme?

Reported it here - https://github.com/livechat/design-system/issues/380

packages/react-components/src/components/Switch/Switch.tsx
MrVorgoth: I'm waiting for @martagranisz for that info

packages/react-components/src/components/Switch/Switch.module.scss
```diff
@@ -188,4 +188,11 @@ $medium-move: $medium-width - $medium-size;
       border-width: 2px;
     }
   }
+
+  &__disabled-icon {
```
sgraczyk, line 192: Shouldn't we add `xxsmall` size for the icon?
https://github.com/livechat/design-system/blob/294d06d5eab1570d29db59e8d0f1e47c2892d33e/packages/react-components/src/components/Icon/Icon.tsx#L16-L37

MrVorgoth: It turned out the icon was wrong -> after changing the lock icon [094a699](https://github.com/livechat/design-system/pull/372/commits/094a6996224c3cc7d832d5e27cc66826f4da7fc5), everything is working as expected and this specific CSS size is removed

martagranisz (approved):

Issue #352, opened 2022-07-20
[Documentation] Include information about importing styles to target project, update starting project in global README


marcinsawicki:
## Description
We need to update our documentation, and add to README information that in order to use components from DS v1, you need to import styles in your application:
`import '@livechat/design-system-react-components/dist/style.css';`

Also please add information to global README about starting the project. It's not that simple to just run `npm install` and `npm run start`. It's also worth noting that you need to install `volta` and then set proper node version according to our `package.json`

sgraczyk:
Released in [v1.0.0-alpha.20](https://github.com/livechat/design-system/releases/tag/v1.0.0-alpha.20)

PR #374, opened 2022-08-01
update README that includes the details how to setup the project

marcinsawicki:
Resolves: #352 

## Description
Update of the README file, which includes the details related to instructions on how to setup the project.

## Storybook
null

## Checklist

**Obligatory:**

- [x] Self-review
- [ ] Storybook cases
- [ ] Design review
- [ ] Functional (QA) review

**Optional:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,33 @@
 
 [See the Storybook](https://v1--613a8e945a5665003a05113b.chromatic.com/)
 
+# How to install?
+
+Install [Homebrew Package Manager](http://mxcl.github.io/homebrew/) and then check it by running commands:
+
+```
+ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
+```
+
+```
+brew doctor
+```
+
+Install NVM (Node Version Manager):
+
+```
+brew install nvm
+```
+
+and follow instructions presented by brew (you will need to add a proper entry to .bash_profile).
+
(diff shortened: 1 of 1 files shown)


TomaszRatajczyk (approved):

MichalPaszowski:
I think I would add mention about how to work with the monorepo, why we have divided packages in that way and not the other,  maybe drop a word or two about lerna and how we use it. How to add a new package, how to bootstrap it etc.

README.md
```diff
@@ -2,6 +2,15 @@
 
 [See the Storybook](https://v1--613a8e945a5665003a05113b.chromatic.com/)
 
+# How to install?
+
+To run the project, first You need to use a the tool for managing the node version, which is convenient for you.
+We suggest using [Volta](https://volta.sh/).
+
+Required version of node is `16.13.2` and npm `8.1.0`.
```
mreszke, line 10: how I can easily switch to this version? Is there some file like `.nvmrc` with the node version so I can switch using one command like `nvm use` when using nvm?

sgraczyk: We've configured the project to use volta, we could also introduce `.nvmrc` file if you think it's necessary.

mreszke: ok, I just read the documentation, and I discovered that node version is changed automatically, so there is no need to do something like `nvm use` every time you change the project 👍 
I think it's worth mentioning that

Issue #373, opened 2022-07-29
[Avatar] Component


sgraczyk:
## Description
Avatars are used to represent a person or a group.

Avatars should support:
* Variants
  * Person
  * Group
* Sizes
  * XXXS
  * XXS
  * XS
  * S
  * M
  * L
  * XL
  * XXL
* Status
  * Online
  * Offline
  * Unknown
* Rim
* Initials built from name
* Fallback image
* Stacking (optional)

## Design
[Figma](https://www.figma.com/file/2pFu80PXO5A2tfyrAGnx91/Product-Components?node-id=6%3A22)

sgraczyk:
Released in [v1.0.0-alpha.18](https://github.com/livechat/design-system/releases/tag/v1.0.0-alpha.18)

PR #375, opened 2022-08-01
#373 Avatar component

sgraczyk:
Resolves: #373 

## Description
Implements the `Avatar` component from scratch. During the development phase I made some decisions to be confirmed by the rest of the team:
* I decided to make the API more generic. For example instead of `kind: 'person' | 'group'` I've used `shape: 'circle' : 'rounded-square'`. A similar change was for the status, as not all products are chat-focused, so I avoided the `accepts chats` name.
* Figma project does not specify the behavior of rim in case of `rounded-square`, nevertheless, I implemented it.
* I'm a bit confused on how to address the position of the status icon as the design seems to do it differently for each size and square. I tried to follow that assumption with some slight modifications.

## Storybook
- [Avatar](https://feature-373--613a8e945a5665003a05113b.chromatic.com/?path=/story/components-avatar--default)

## Checklist

**Obligatory:**

- [x] Self-review
- [x] Storybook cases
- [x] Design review by @kstvsko 
- [x] Functional (QA) review by @TomaszRatajczyk 

**Optional:**

### DIFF
--- a/packages/react-components/src/components/Avatar/Avatar.helpers.ts
+++ b/packages/react-components/src/components/Avatar/Avatar.helpers.ts
@@ -11,6 +11,6 @@ export function getInitials(name = '', count = 2): string {
 
 export function getFontColor(color: string): string {
   return getContrast(color, '#FFFFFF') > 4.5
-    ? 'var(--content-white-locked)'
-    : 'var(--content-subtle)';
+    ? 'var(--color-white)'
+    : 'var(--color-black)';
 }
--- a/packages/react-components/src/components/Avatar/Avatar.helpers.ts
+++ b/packages/react-components/src/components/Avatar/Avatar.helpers.ts
@@ -1,0 +1,16 @@
+import { getContrast } from 'polished';
+
+export function getInitials(name = '', count = 2): string {
+  return name
+    .split(' ')
+    .map((el) => el.charAt(0))
+    .join('')
+    .substring(0, count)
+    .toUpperCase();
+}
+
+export function getFontColor(color: string): string {
+  return getContrast(color, '#FFFFFF') > 4.5
+    ? 'var(--content-white-locked)'
+    : 'var(--content-subtle)';
+}
--- a/packages/react-components/src/components/Avatar/Avatar.spec.tsx
+++ b/packages/react-components/src/components/Avatar/Avatar.spec.tsx
@@ -1,0 +1,139 @@
+import * as React from 'react';
+import { render, fireEvent, vi } from 'test-utils';
+import { Avatar, AvatarProps } from './Avatar';
+
+const renderComponent = (props: AvatarProps) => {
+  return render(<Avatar {...props} />);
+};
+
+const baseClass = 'avatar';
+
+describe('<Alert> component', () => {
+  it('should allow for custom CSS class', () => {
+    const customClass = 'custom-class';
+    const { container } = renderComponent({
+      type: 'text',
+      className: customClass,
+    });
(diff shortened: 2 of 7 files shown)


packages/react-components/src/components/Avatar/Avatar.helpers.ts
```diff
@@ -0,0 +1,16 @@
+import { getContrast } from 'polished';
+
+export function getInitials(name = '', count = 2): string {
```
sgraczyk, line 3: At this moment it does not support diacritics or emojis.

kstvsko:
@sgraczyk 

I've added updated avatars, also with status indicators, that are now scaled and positioned at the same place at every size. I've added also the scale for the avatars with rims. :) 

Rims: https://www.figma.com/file/2pFu80PXO5A2tfyrAGnx91/Product-Components?node-id=8167%3A31183
Status Indicators: https://www.figma.com/file/2pFu80PXO5A2tfyrAGnx91/Product-Components?node-id=736%3A8527

MichalPaszowski:
Something not pykło here https://monosnap.com/file/1aOi1ZZ2lR00rwnWGF2JmDl9Ks13r3

MichalPaszowski (approved):

sgraczyk:
> Something not pykło here https://monosnap.com/file/1aOi1ZZ2lR00rwnWGF2JmDl9Ks13r3

Should be fixed within https://github.com/livechat/design-system/pull/375/commits/99b29a141f0edf68f0946b178b36731967670c83

sgraczyk:
> @sgraczyk
> 
> I've added updated avatars, also with status indicators, that are now scaled and positioned at the same place at every size. I've added also the scale for the avatars with rims. :)
> 
> Rims: https://www.figma.com/file/2pFu80PXO5A2tfyrAGnx91/Product-Components?node-id=8167%3A31183 Status Indicators: https://www.figma.com/file/2pFu80PXO5A2tfyrAGnx91/Product-Components?node-id=736%3A8527

Status indicators are implemented as proposed, however, I'd like to wait before introducing changes designed for rims. It looks that now the size of the image is adjusted so the rim fits into the original size of the avatar. It'll probably introduce a flickering effect, especially that the rim is a UI element that may appear and disappear when a specific condition changes. Can we move the work around rims to another issue, and accept the current state for now?

kstvsko (approved):

mreszke (approved):

packages/react-components/src/components/Avatar/Avatar.spec.tsx
```diff
@@ -0,0 +1,139 @@
...
+    expect(container.firstChild).toHaveStyle(
+      `background-color: ${customColor}`
+    );
+  });
+
+  const extraSmallSizes: AvatarProps['size'][] = [
+    'xxxsmall',
+    'xxsmall',
+    'xsmall',
+  ];
```
MichalPaszowski, line 48: missing space - perhaps it should be put up top of the file?

packages/react-components/src/components/Avatar/Avatar.spec.tsx
```diff
@@ -0,0 +1,139 @@
...
+  );
+
+  const smallAndAboveSizes: AvatarProps['size'][] = [
+    'small',
+    'medium',
+    'large',
+    'xlarge',
+    'xxlarge',
+  ];
+  it.each(smallAndAboveSizes)(
```
MichalPaszowski, line 69: same as above

Issue #385, opened 2022-08-08
[Toast] Change CloseIcon size and background colour for 'info' variant


sgraczyk:
## Description
* `CloseIcon` should use `medium` size (20x20)
* `Info` variant should use `Surface Invert/Default --surface-invert-default` colour

## Design
[Figma](https://www.figma.com/file/2pFu80PXO5A2tfyrAGnx91/Product-Components?node-id=6904%3A31508)

sgraczyk:
Released in [v1.0.0-alpha.19](https://github.com/livechat/design-system/releases/tag/v1.0.0-alpha.19)

PR #388, opened 2022-08-08
#385 Change CloseIcon size and background colour for 'info' variant

sgraczyk:
Resolves: #385

## Description
* Added basic `shadow` tokens
* Reorganized `Toast` styles (mostly simplified)
* Removed `notification` kind, fixed color tokens for `info` kind
* Added more advanced stories for `Toast`
* Added missing export of `Avatar`
* Renamed some `Toast` props to follow convention

## Storybook
https://feature-385--613a8e945a5665003a05113b.chromatic.com/

## Checklist

**Obligatory:**

- [x] Self-review
- [x] Storybook cases
- [x] Design review
- [x] Functional (QA) review

**Optional:**

### DIFF
--- a/packages/react-components/src/components/Toast/Toast.stories.tsx
+++ b/packages/react-components/src/components/Toast/Toast.stories.tsx
@@ -16,7 +16,6 @@ export default {
     It disappears on its own after a few seconds. It provides a feedback about an operation 
     for user.
     `,
-    exclude: { controls: { exclude: ['action'] } },
   },
   argTypes: {
     onClose: { action: 'closed' },
--- a/packages/react-components/src/components/Toast/Toast.stories.tsx
+++ b/packages/react-components/src/components/Toast/Toast.stories.tsx
@@ -1,11 +1,12 @@
 import * as React from 'react';
 import { ComponentMeta, Story } from '@storybook/react';
 
-import { ToastProps, Toast as ToastComponent } from './Toast';
+import { StoryDescriptor } from '../../stories/components/StoryDescriptor';
+import { ToastProps, Toast } from './Toast';
 
 export default {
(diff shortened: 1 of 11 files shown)


kstvsko (approved):

packages/react-components/src/components/Toast/Toast.stories.tsx
```diff
@@ -1,64 +1,86 @@
...
-  removable: true,
-  onClose: () => alert('onClose click'),
-};
+export const WithCustomAction = (): JSX.Element => (
+  <>
+    <StoryDescriptor title="Without close icon">
+      <Toast
+        action={{
+          label: 'Show details',
+          onClick: () => alert('Custom action button clicked'),
```
marcinsawicki, line 54: image

We can use the SB functionality to track such events in UI (Story > Canvas > Actions)

Example from `Picker` component stories:
`export default {
  title: 'Components/Picker',
  component: PickerComponent,
  parameters: {
    componentSubtitle: `TBD`,
  },
  argTypes: {
    onSelect: { action: 'changed' },
  },
} as ComponentMeta<typeof PickerComponent>;`

marcinsawicki (approved):

packages/react-components/src/components/Toast/Toast.stories.tsx
sgraczyk: Ok, I think I should do so in the main story, as in the remaining ones I use regular components. I'll think about how to handle it in this specific case.

sgraczyk: I've applied some changes e2b6981

MichalPaszowski (approved):

Issue #332, opened 2022-06-24
[Popover] should have z-index set to prevent content overlapping


MichalPaszowski:
## Describe the bug
One image is worth a thousand words: 
![Image](https://user-images.githubusercontent.com/7773964/175543476-119fd3f4-e202-4da2-8639-ca38035bcc4c.png)

## Steps To Reproduce
Place popover near element with higher z-index.

## Expected behaviour
A popover should be over(hehe) other elements.

## Screenshots
![Image](https://user-images.githubusercontent.com/7773964/175543476-119fd3f4-e202-4da2-8639-ca38035bcc4c.png)

## Environment
N/A

## Additional context
N/A

sgraczyk:
Released in [v1.0.0-alpha.22](https://github.com/livechat/design-system/releases/tag/v1.0.0-alpha.22)

PR #398, opened 2022-08-26
#332 z-index for Popover component

MichalPaszowski:
Resolves: #[332](https://github.com/livechat/design-system/issues/332)

## Description
It seems to be enough to fix stuff like:
https://monosnap.com/file/agb5EyL9CrbJqO8E0vNQhtUAo6BYOo

## Storybook
https://613a8e945a5665003a05113b-iaeypnmfan.chromatic.com/

## Checklist

**Obligatory:**

- [x] Self-review
- [ ] Storybook cases
- [ ] Design review
- [x] Functional (QA) review

**Optional:**

### DIFF
--- a/packages/react-components/src/components/Popover/Popover.module.scss
+++ b/packages/react-components/src/components/Popover/Popover.module.scss
@@ -5,7 +5,7 @@
   display: none;
   max-width: 336px;
   min-width: 168px;
-  z-index: 1000;
+  z-index: $stacking-context-level-popover;
 
   &:focus {
     outline: none;
--- a/packages/react-components/src/components/Popover/Popover.module.scss
+++ b/packages/react-components/src/components/Popover/Popover.module.scss
@@ -5,6 +5,7 @@
   display: none;
   max-width: 336px;
   min-width: 168px;
+  z-index: 1000;
 
   &:focus {
     outline: none;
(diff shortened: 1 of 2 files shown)


packages/react-components/src/components/Popover/Popover.module.scss
```diff
@@ -5,6 +5,7 @@
   display: none;
   max-width: 336px;
   min-width: 168px;
+  z-index: 1000;
```
sgraczyk, line 8: Could we use a dedicated variable from
https://github.com/livechat/design-system/blob/39fcca44937242ba3bd3bc7c8b855269d335b2f1/packages/react-components/src/utils/StackingContextLevel.scss

I believe we would need to introduce a new level:

W could follow this concept from LiveChat's Agent App:

```ts
export enum StackingContextLevel {
  Tooltip = 1000,
  Dropdown = 2000,
  ...
}
```

jpyzio123 (approved):

packages/react-components/src/components/Popover/Popover.module.scss
```diff
@@ -5,6 +5,7 @@
   display: none;
   max-width: 336px;
   min-width: 168px;
+  z-index: $stacking-context-level-popover;
```
sgraczyk, line 8: You need to import it before usage `@import '../../utils/StackingContextLevel';`

sgraczyk (changes requested):
Build fails, the variable is not imported

sgraczyk (approved):

Issue #380, opened 2022-08-04
[Switch] Lock icon invisible in dark theme


MrVorgoth:
## Describe the bug
There is no `lock` icon inside Switch component, when `disabled` prop is passed and dark theme is applied

## Steps to reproduce
1. Go to Switch component on Storybook
2. Click on `Change the theme on the preview` (in case it's hard to find - it's the 4th icon at the top in the middle column)
3. Choose `dark`
4. Scroll down to Stories - States
5. See the disabled state -> no icon is visible

## Expected behavior
Lock icon should be visible (either icon color needs to be changed or the switch color). Right now the switch has forced white design token, that's why it's failing on the dark theme, because on dark theme the icon is also white so white icon on white background is not visible

## Screenshots
How it looks on the white theme:
<img width="128" alt="Screenshot 2022-08-04 at 14 11 43" src="https://user-images.githubusercontent.com/20707838/182846522-755814e3-240e-4ef8-af11-8f388a4f734d.png">

How it looks on dark theme (bug):
<img width="125" alt="Screenshot 2022-08-04 at 14 11 58" src="https://user-images.githubusercontent.com/20707838/182846556-29c1bc57-9141-44c6-a01e-8afc610f8334.png">

## Environment
It doesn't matter - everywhere it'll be the same as it's about wrong design tokens

## Additional context
None

sgraczyk:
Released in [v1.0.0-alpha.22](https://github.com/livechat/design-system/releases/tag/v1.0.0-alpha.22)

PR #402, opened 2022-09-02
#380 introduce black-locked token and use it with switch icon

MichalPaszowski:
Resolves: #380

## Storybook
https://lock-icon-switch-380--613a8e945a5665003a05113b.chromatic.com/

## Checklist

**Obligatory:**

- [x] Self-review
- [ ] Storybook cases
- [ ] Design review
- [x] Functional (QA) review


sgraczyk:
It's still not working in linked Storybook
image

packages/react-components/src/components/Switch/Switch.tsx
```diff
@@ -110,6 +110,7 @@ export const Switch: React.FC<SwitchProps> = ({
           )}
           {isLocked && (
             <Icon
+              customColor={styles[`${baseClass}__icon`]}
```
sgraczyk, line 113: I think it would be better to just apply a className here eg,  `className={styles[`${baseClass}__icon`]}`, and set up a color there. It will be easier due to the fact, that in order to use CSS var in JavaScript it is required to compute it -> https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties#values_in_javascript.

I believe the current solution does not work as `customColor` expects a valid color value, while you provide a className with CSS styles.

sgraczyk (changes requested):
Current solution does not work

sgraczyk (approved):

packages/react-components/src/components/Switch/Switch.module.scss
```diff
@@ -188,4 +188,8 @@ $medium-move: $medium-width - $medium-size;
       border-width: 2px;
     }
   }
+
+  &__icon {
+    color: var(--content-black-locked);
+  }
```
sgraczyk, line 194:
```suggestion
  &__icon svg {
    fill: var(--content-black-locked);
  }
```

MichalPaszowski: just did
```
className={styles[`${baseClass}__icon`]}
```
instead of
```
className={`${baseClass}__icon`} 
```
and it clicked :)

jpyzio123 (approved):

PR #412, opened 2022-09-06
Remove disabled attribute from button

MateuszRomek:
## Description
Addresses issue with `disabled` attribute which creates more harm than it should be. Due to that, We run into problems especially with the tooltip component, which gets lost because his child has the disabled attribute.
Even floating-ui [recommends](https://floating-ui.com/docs/react-dom-interactions#disabled-elements) to handle `disabled` elements in that way instead of passing `disabled` attribute on the button.

Here are good article about why we should avoid disabled buttons https://css-tricks.com/making-disabled-buttons-more-inclusive/

## Storybook
https://feature-remove-button-disabled-attr--613a8e945a5665003a05113b.chromatic.com/

## Checklist

**Obligatory:**

- [x] Self-review
- [x] Storybook cases
- [x] Design review
- [x] Functional (QA) review

### DIFF
--- a/packages/react-components/src/components/Button/Button.spec.tsx
+++ b/packages/react-components/src/components/Button/Button.spec.tsx
@@ -107,16 +107,11 @@ describe('<Button> component', () => {
     expect((linkEl as HTMLElement).tagName).toBe('A');
   });
 
-  it('should set aria-disabled if we pass disabled prop', () => {
-    const { btnEl } = renderButton({ disabled: true });
+  it('should set aria-disabled if we pass disabled prop and should not fire onClick callback if disabled is true', () => {
+    const onClick = vi.fn();
+    const { btnEl } = renderButton({ disabled: true, onClick });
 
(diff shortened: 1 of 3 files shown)


jpyzio123 (approved):

packages/react-components/src/components/Button/Button.spec.tsx
```diff
@@ -107,4 +106,20 @@ describe('<Button> component', () => {
...
+
+  it('should not fire callback if we pass disabled prop as true', () => {
+    const onClick = vi.fn();
+
+    const { btnEl } = renderButton({ disabled: true, onClick });
+
+    fireEvent.click(btnEl as Element);
+
+    expect(onClick).not.toHaveBeenCalled();
+  });
```
sgraczyk, line 124: I think we can easily merge these tests

sgraczyk (approved):

kstvsko (approved):

PR #419, opened 2022-09-13
#418 Add closeOnOverlayPress prop to Modal

### LINKED ISSUE #418: [Modal] Do not close modal on overlay click
## Description
Adding a new property to the Modal component to allow not to close modal on overlay click.

## Design
N/A

VadymBezpalko:
Resolves: #418 
## Description
Adding a new property to the Modal component to allow not to close modal on overlay click.

## Storybook
https://feature-418--613a8e945a5665003a05113b.chromatic.com/

## Checklist

**Obligatory:**

- [x] Self-review
- [x] Storybook cases
- [ ] Design review
- [ ] Functional (QA) review

**Optional:**

### DIFF
--- a/packages/react-components/src/components/Modal/Modal.stories.tsx
+++ b/packages/react-components/src/components/Modal/Modal.stories.tsx
@@ -30,6 +30,7 @@ Modal.args = {
   heading: 'Modal',
   children: 'Modal content',
   closeOnEscPress: true,
+  closeOnOverlayPress: true,
   style: { width: '600px', height: '400px' },
   footer: (
     <React.Fragment>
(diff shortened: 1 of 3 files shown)


MateuszRomek (approved):
My only concern is we will suddenly introduce strange behavior for already used modals? 
I think maybe we should set this prop by default to `false` to keep the current behavior and use prop to specify if we want to close overlay on click?

JoannaSikora (approved):

MateuszRomek:
> My only concern is we will suddenly introduce strange behavior for already used modals? I think maybe we should set this prop by default to `false` to keep the current behavior and use prop to specify if we want to close overlay on click?

Nvm it's already working like that 🙂 My bad

sgraczyk (approved):

PR #424, opened 2022-09-19
#423 Provide way of passing flip parameters to Popover component

### LINKED ISSUE #423: [Popover] Provide way of passing options from floating-ui
## Description
One of the teams has reported that the Popover component lacks options when it comes to placement. 
This PR provides way of passing options for flip() method available in floatingUI library
https://floating-ui.com/docs/flip

MichalPaszowski:
Resolves #423

## Description
One of the teams has reported that the Popover component lacks options when it comes to placement.
This PR provides a way of passing options for flip() method available in floatingUI library
https://floating-ui.com/docs/flip

## Storybook
https://popover-more-options--613a8e945a5665003a05113b.chromatic.com/

## Checklist

**Obligatory:**

- [x] Self-review
- [ ] Storybook cases
- [ ] Design review
- [x] Functional (QA) review

**Optional:**

### DIFF
--- a/packages/react-components/src/components/Popover/Popover.tsx
+++ b/packages/react-components/src/components/Popover/Popover.tsx
@@ -6,7 +6,9 @@ import {
   flip,
   offset,
   autoUpdate,
+  DetectOverflowOptions,
 } from '@floating-ui/react-dom';
+import type { Options as FlipOptions } from '@floating-ui/core/src/middleware/flip';
 
 import cssStyles from './Popover.module.scss';
 
@@ -15,6 +17,7 @@ export interface IPopoverProps {
   className?: string;
   placement?: Placement;
   isVisible?: boolean;
+  flipOptions?: Partial<FlipOptions & DetectOverflowOptions>;
   triggerRenderer: () => React.ReactNode;
 }
 
@@ -24,6 +27,7 @@ export const Popover: React.FC<IPopoverProps> = (props) => {
     children,
     className,
     placement,
+    flipOptions,
     isVisible = false,
   } = props;
   const [visible, setVisibility] = React.useState(false);
@@ -38,7 +42,7 @@ export const Popover: React.FC<IPopoverProps> = (props) => {
     update,
     placement: updatedPlacement,
   } = useFloating({
-    middleware: [offset(4), flip()],
+    middleware: [offset(4), flip(flipOptions)],
     placement: placement,
   });
 
--- a/packages/react-components/src/components/Popover/Popover.tsx
+++ b/packages/react-components/src/components/Popover/Popover.tsx
@@ -6,9 +6,7 @@ import {
   flip,
   offset,
   autoUpdate,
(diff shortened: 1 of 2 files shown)


packages/react-components/src/components/Popover/Popover.tsx
```diff
@@ -6,7 +6,9 @@ import {
   flip,
   offset,
   autoUpdate,
+  DetectOverflowOptions,
 } from '@floating-ui/react-dom';
+import type { Options as FlipOptions } from '@floating-ui/core/src/middleware/flip';
```
sgraczyk, line 11: The path for this import looks very strange. Can't we somehow derive this type?

packages/react-components/src/components/Popover/Popover.tsx
```diff
@@ -15,6 +17,7 @@ export interface IPopoverProps {
   className?: string;
   placement?: Placement;
   isVisible?: boolean;
+  flipOptions?: Partial<FlipOptions & DetectOverflowOptions>;
```
sgraczyk, line 20: Can we showcase this within `Storybook`?

sgraczyk: Consider this solution

```suggestion
  flipOptions?: Parameters<typeof flip>[0];
```

kstvsko (approved):

packages/react-components/src/components/Popover/Popover.tsx
marcinsawicki: The provided link doesn't show this (and some of the others) props available for this component.
Please fix that, because currently there is impossible to check how their props work to check the component behavior.

marcinsawicki (commented):
`flipOptions` prop isn't visible in the Storybook, so we can't fully check how the component works.

jpyzio123 (approved):

packages/react-components/src/components/Popover/Popover.tsx
MichalPaszowski: @sgraczyk It works and if you think it's better that way, I can leave it but IMHO reaching the first parameter with "[0]" is quite cryptic.

MichalPaszowski:
> `flipOptions` prop isn't visible in the Storybook, so we can't fully check how the component works.

I have provided a way of testing it, please look again.

packages/react-components/src/components/Popover/Popover.tsx
sgraczyk: @MichalPaszowski It's easier to maintain compatibility with the library that way. It may look strange, but we're actually interested in the first (and only) parameter of a function.

sgraczyk: There may be a better way in TS-land but I'm not aware of it.

sgraczyk (approved):

mreszke (approved):

marcinsawicki (approved):

PR #439, opened 2022-09-28
#440 Change on focus property from box shadow to filter

rafalluksza:
Resolves: #440 and https://livechatinc.atlassian.net/browse/AA-13976

## Description
When we use buttons and add the margin, the button's highlights are not in the right place. I changed from `box-shadow` to the `filter: drop-shadow` property. 

## Design
Before: 
Screenshot 2022-10-03 at 10 21 30

After: 
Screenshot 2022-10-03 at 10 26 32

## Storybook
https://feature-AA-13976--613a8e945a5665003a05113b.chromatic.com/

## Checklist

**Obligatory:**

- [x] Self-review
- [ ] Storybook cases
- [ ] Design review
- [ ] Functional (QA) review

**Optional:**

### DIFF
--- a/packages/react-components/src/components/Button/Button.module.scss
+++ b/packages/react-components/src/components/Button/Button.module.scss
@@ -28,7 +28,7 @@ $base-class: 'btn';
   }
 
   &:focus {
-    box-shadow: 0 0 1px 2px rgb(var(--color-action-default-rgb) 0.7);
+    filter: drop-shadow(0 0 1px 2px rgb(var(--color-action-default-rgb) 0.7));
   }
 
   &--disabled {


marcinsawicki:
Please add this task to our GH Board and assign the correct link in the PR description.
https://github.com/orgs/livechat/projects/7/views/1
Also, the link for the Storybook in the description is incorrect.

packages/react-components/src/components/Button/Button.module.scss
```diff
@@ -28,7 +28,7 @@ $base-class: 'btn';
   }
 
   &:focus {
-    box-shadow: 0 0 1px 2px rgb(var(--color-action-default-rgb) 0.7);
+    filter: drop-shadow(0 0 1px 2px rgb(var(--color-action-default-rgb) 0.7));
```
marcinsawicki, line 31: It will be nice if You add "before" and "after" images to the description because now we can't easily verify this change without playing with code in the inspector. WDYT?

rafalluksza: Yes, master!

shwarcu (approved):

sgraczyk (approved):

MichalPaszowski (approved):

jpyzio123 (approved):

Issue #444, opened 2022-10-17
[Switch] Allow setting accessible name for switch input element


shwarcu:
## Description

At the moment looks like there is no way to set either `aria-label` or `aria-labelledby` for checkbox input element inside the switch. I am trying to improve accessibility in a place that uses Switch from design system but both old and new component doesn't give option for that.

Having ability to set `aria-label` for input element inside checkbox would unblock me.

## Design
<img width="1320" alt="Screenshot 2022-10-17 at 08 27 49" src="https://user-images.githubusercontent.com/58426925/196104898-ee13109d-2a97-45a0-a390-fb6ff712ea5a.png">

PR #445, opened 2022-10-17
[Switch] Allow accessible name for switch input element

shwarcu:
Resolves: #444

## Description
At the moment looks like there is no way to set either `aria-label` or `aria-labelledby` for checkbox input element inside the switch. 
I propose to add new optional property for Switch interface to allow setting accessible name (one of several ways to achieve it). 
I also removed test-id that seems to be some kind of leftover after moving components to new DS setup. I didn't find any usage of this `foo` test id in the repository. I think leaving it there was not intentional.

Screenshot 2022-10-17 at 08 27 49

## Storybook
https://feature-v1-switch-accessibility-props--613a8e945a5665003a05113b.chromatic.com/

## Checklist

**Obligatory:**

- [x] Self-review
- [ ] Storybook cases
- [ ] Design review
- [ ] Functional (QA) review

**Optional:**

- [x] Accessibility cases (keyboard control, correct HTML markup, etc.)

### DIFF
--- a/packages/react-components/src/components/Switch/Switch.spec.tsx
+++ b/packages/react-components/src/components/Switch/Switch.spec.tsx
@@ -62,4 +62,12 @@ describe('Switch', () => {
     expect(checkbox.disabled).toEqual(true);
     expect(lockIcon).toBeVisible();
   });
+
+  it('should allow setting accessible name for input element', () => {
+    const label = 'Hello world';
+    const { getByRole } = render(<Switch on={true} ariaLabel={label} />);
+
+    const checkbox = getByRole('checkbox', { name: label });
+    expect(checkbox).toHaveAccessibleName(label);
+  });
 });
--- a/packages/react-components/src/components/Switch/Switch.tsx
+++ b/packages/react-components/src/components/Switch/Switch.tsx
@@ -14,6 +14,7 @@ export type SwitchSize = 'compact' | 'medium' | 'large';
 export type SwitchState = 'regular' | 'loading' | 'locked';
 
 export interface SwitchProps {
+  ariaLabel?: string;
   className?: string;
   defaultOn?: boolean;
   disabled?: boolean;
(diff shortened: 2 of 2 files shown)


packages/react-components/src/components/Switch/Switch.spec.tsx
```diff
@@ -62,4 +62,12 @@ describe('Switch', () => {
     expect(checkbox.disabled).toEqual(true);
     expect(lockIcon).toBeVisible();
   });
+
+  it('should allow setting accessible name for input element', () => {
+    const label = 'Hello world';
+    const { getByRole } = render(<Switch on={true} ariaLabel={label} />);
+
+    const checkbox = getByRole('checkbox', { name: label });
```
shwarcu, line 70: Or 
```suggestion
    const checkbox = getByRole('checkbox');
```

? 🤔

sgraczyk: `const checkbox = getByRole('checkbox');` should be enough, but I think it doesn't matter that much in this particular test case.

sgraczyk (approved):

packages/react-components/src/components/Switch/Switch.tsx
```diff
@@ -23,6 +23,7 @@ export interface SwitchProps {
   onChange?(e: React.FormEvent, value: boolean): void;
   size?: SwitchSize;
   state?: SwitchState;
+  ariaLabel?: string;
```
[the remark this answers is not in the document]
shwarcu: Done in https://github.com/livechat/design-system/pull/445/commits/53e8b99595e44e1c37ac213581030a5a5a8ff227

packages/react-components/src/components/Switch/Switch.spec.tsx
shwarcu: I will change it, as with `const checkbox = getByRole('checkbox');` we should get better error if the test fails

MichalPaszowski (approved):

dawid-drelichowski (approved):

jpyzio123 (approved):

PR #473, opened 2023-01-10
[fixes] Card width, Tag background color, Picker interface

MichalPaszowski:
This is a PR that fixes few obstacles that we have encountered during introduction of new DS to AA.

### DIFF
--- a/packages/react-components/src/components/Card/Card.module.scss
+++ b/packages/react-components/src/components/Card/Card.module.scss
@@ -9,7 +9,6 @@ $base-class: card;
   border-radius: 4px;
   background-color: var(--background);
   padding: 16px;
-  min-width: 320px;
 
   &__header {
     display: flex;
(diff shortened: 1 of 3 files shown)


packages/react-components/src/components/Card/Card.module.scss
```diff
@@ -9,7 +9,6 @@ $base-class: card;
   border-radius: 4px;
   background-color: var(--background);
   padding: 16px;
-  min-width: 320px;
```
sgraczyk, line 12: Shouldn't we have some sort of `min-width`?  How will the card render if the content is lacking?

sgraczyk (approved):

packages/react-components/src/components/Card/Card.module.scss
sgraczyk: Ok, I've tested this solution and there seems to be no negative impact after removing `min-width`.

jpyzio123 (approved):

packages/react-components/src/components/Card/Card.module.scss
MichalPaszowski: 👌🏼

PR #484, opened 2023-02-02
#485 Adding missing children type

kamilmateusiak:
Resolves: #485

## Description
When project is using React 18 some components have incorrect types. From React v18 React.FC doesn't have `children` prop by default and it need to be added manually to component props. Since react-components has defined following peer dependencies:
```json
  "peerDependencies": {
    "react": ">= 17.0.2",
    "react-dom": ">= 17.0.2"
  },
```
we should support React v18 as well.

## Storybook
This should be available: 
https://v1-react-v18-types-613a8e945a5665003a05113b.chromatic.com/ or https://kamilmateusiak:v1-react-v18-types-613a8e945a5665003a05113b.chromatic.com/
but is not. Probably because I need to use fork to contribute

## Checklist

**Obligatory:**

- [x] Self-review
- [ ] Storybook cases
- [ ] Design review
- [ ] Functional (QA) review


packages/react-components/src/components/UploadBar/UploadBar.tsx
```diff
@@ -23,6 +23,7 @@ const wrapperHeaderClass = `${baseClass}__wrapper__header`;
 const TRANSITION_TIMEOUT = 300;
 
 export interface UploadBarProps {
+  children?: React.ReactNode;
```
kamilmateusiak, line 26: I noticed that UploadBar is not exported. Is it on purpose?

kamilmateusiak:
@marcinsawicki can you help me with this one and ask the team for feedback? Sorry to bother you :D

Issue #406, opened 2022-09-04
[Search] Improvements 


martagranisz:
## Description

- When the loading state is active, there should be a variant where search is active and can be modified
- When the loading state is active, searching phrase should be visible 
- Collapsed search should not have a stroke/ border

## Design
TBA https://www.figma.com/file/2pFu80PXO5A2tfyrAGnx91/Product-Components?node-id=3479%3A19213

martagranisz:
Think about how to show spinner on the disabled input> for now on it's not visible enough

PR #488, opened 2023-02-07
#406 - Search - adjusted logic and styles

marcinsawicki:
Resolves: #406

## Description
Changes related to the behavior of `isCollapsable` version of `Search`.
Some of the logic responsible for UI has been moved to the component container (background, borders, etc.), to make it easier to control in the `isCollapsable` version.
Loader has his own background to be better visible in the state `disabled` + `isLoading`.
Also added a separate story for the `isCollapsable` version, to show behavior with other, potential elements in the UI.

## Storybook
https://feature-406--613a8e945a5665003a05113b.chromatic.com/

## Checklist

**Obligatory:**
- [x] Self-review
- [x] Storybook cases
- [x] Design review
- [x] Functional (QA) review

**Optional:**
- [x] Accessibility cases (keyboard control, correct HTML markup, etc.)

### DIFF
--- a/packages/react-components/src/components/Search/Search.spec.tsx
+++ b/packages/react-components/src/components/Search/Search.spec.tsx
@@ -1,13 +1,15 @@
 import * as React from 'react';
-import { render, fireEvent, vi } from 'test-utils';
+import { render, userEvent, fireEvent, vi } from 'test-utils';
 import noop from '../../utils/noop';
-import { ISearchInputProps, SearchInput, SearchSize } from './Search';
+import { ISearchInputProps, SearchInput } from './Search';
 import styles from './Search.module.scss';
 
 const baseClass = 'search-input';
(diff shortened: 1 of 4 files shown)


packages/react-components/src/components/Search/Search.spec.tsx
```diff
@@ -95,43 +188,56 @@ describe('<Search> component', () => {
...
+          }}
+        />
+      );
+    };
+
+    const { getByRole } = render(<SearchWrapper />);
+    const text = 'test value';
+
+    userEvent.type(getByRole('textbox'), text, { delay: 0 });
+    expect(onChangeFunction).toHaveBeenCalledWith(text);
```
marcinsawicki, line 229: Since this test is for a controlled component, we need to render it along with the external state, if we want to use `userEvent` to simulate a user entering text.

MichalPaszowski (approved):

Issue #477, opened 2023-01-17
[Picker] Fix selected options when using search


MrVorgoth:
When selecting some options in `multi` mode and then searching for something and selecting such options all other options, which were selected and are not visible right now (are filtered via search) are gone. See movie below

https://user-images.githubusercontent.com/20707838/212837952-127c8045-9e07-48a6-96b4-3b56aacb301d.mov

Issue #395, opened 2022-08-19
[Picker] Displaying long strings


kstvsko:
## Description
We should add the handling of the long strings in `Picker`. 
If it doesn't fit into the Picker field we should add the ellipsis at the end and display as much text as possible.

<img width="377" alt="image" src="https://user-images.githubusercontent.com/109948903/185565303-63f4dc03-ec13-41c6-981e-8c3dadb6ca71.png">

PR #491, opened 2023-02-08
#479 #477 #395 - Picker - bug fixes

marcinsawicki:
Resolves: #479 #477 #395 

## Description
Some bug fixes:
- added ellipsis for long strings in `Trigger Body` (also added to list items)
- fixed the issue in `multi` mode, where items could be removed even with a disabled state by just blocking the possibility to click the "x" icon
- fixed the issue with selecting the items on a filtered list

## Storybook
https://feature-picker-fixes--613a8e945a5665003a05113b.chromatic.com/

## Checklist

**Obligatory:**

- [x] Self-review
- [x] Storybook cases
- [ ] Design review
- [ ] Functional (QA) review

**Optional:**

### DIFF
--- a/packages/react-components/src/components/Picker/Picker.stories.tsx
+++ b/packages/react-components/src/components/Picker/Picker.stories.tsx
@@ -1,167 +1,123 @@
 import * as React from 'react';
 import { ComponentMeta, Story } from '@storybook/react';
 
-import { IPickerProps, Picker as PickerComponent } from './Picker';
+import { IPickerProps, Picker } from './Picker';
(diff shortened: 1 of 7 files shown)


kstvsko (changes requested):
When I have an option in multi-select with a long copy, it shows an ellipsis on the list, but when I select it, it glitches somehow, and it looks weird, it almost touches the "clear" button. But I am not 100% sure that we should fix it, because then we won't have a place where that kind of option will display fully. (I do not know if that's a problem)

image

MichalPaszowski (approved):

kstvsko (approved):

Issue #379, opened 2022-08-03
Remove React.x prefix across all components


MichalPaszowski:
## Description
React. prefix makes a lot of noise across all components. Introduce global import.

PR #515, opened 2023-04-12
remove general react import

MichalPaszowski:
Resolves: #379

## Description
Currently, across all components, we use hooks with React.x prefix. It makes code noisier than it should; This PR fixes that

## Storybook
https://feature-379--613a8e945a5665003a05113b.chromatic.com/

## Checklist

**Obligatory:**

- [x] Self-review
- [ ] Storybook cases
- [ ] Design review
- [ ] Functional (QA) review

**Optional:**

### DIFF
--- a/packages/react-components/src/components/DatePicker/types.ts
+++ b/packages/react-components/src/components/DatePicker/types.ts
@@ -1,10 +1,11 @@
-import { Reducer } from 'react';
+import { ReactElement } from 'react';
+import { Reducer, Ref } from 'react';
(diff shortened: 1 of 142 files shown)


packages/react-components/src/components/DatePicker/types.ts
```diff
@@ -1,10 +1,11 @@
-import { Reducer } from 'react';
+import { ReactElement } from 'react';
+import { Reducer, Ref } from 'react';
```
marcinsawicki, line 2: ?

marcinsawicki (approved):

Issue #510, opened 2023-03-29
[Checkbox] - Improve implementation details with a11y features


marcinsawicki:
Improve implementation details with `a11y` features like `role=checkbox` and `aria-selected="true"`, etc.
It will also improve the possibility to write the tests for this component.

ashbork:
There's another, minor issue which I can fix separately and it doesn't break anything; the div is actually selectable which causes some issues in our app. We have a custom implementation for Shift-clicking checkboxes to select all items between two specific boxes; this breaks with the DS component because the items are selectable and default behavior kicks in. This can be fixed with `user-select:none` on the parent and `user-select:text` on inner elements that should be selectable.

vladko-uxds:
Please use another color token for check icon
--content-invert-primary

ashbork:
Done, thanks @vladko-uxds

PR #518, opened 2023-04-14
rewrite the Checkbox component to input

ashbork:
Resolves: #510 

Note that introducing this would be a breaking change - the internals would change and some of DS' consumers' tests will break.

## Description
The Checkbox component being a `div` makes it less accessible and less usable through the keyboard. It isn't plug-and-play; introducing it into our app broke a couple of functionalities that were previously there. As such, I've tried rewriting the component to a native `input` element. I'd also suggest splitting the component into two parts: `Checkbox` and `LabeledCheckbox`; the first one would contain only the checkbox itself, while the latter would have the whole package (just like `Checkbox` does now. This would simplify testing greatly as we wouldn't need to dive into the div and label that's otherwise there.

I'd appreciate your help on the matter; this is my first contribution to DS, and I'm fairly new to FE altogether.

## Storybook
https://feature-510--613a8e945a5665003a05113b.chromatic.com/

## Checklist

**Obligatory:**

- [x] Self-review
- [x] Unit & integration tests - green
- [x] Storybook cases - they work as they did previously
- [x] Design review - design is 1:1 with the old one
- [ ] Functional (QA) review

**Optional:**

- [x] Accessibility cases (keyboard control, correct HTML markup, etc.) - this solution is much better in terms of keyboard navigability and follows ARIA rules for accessible interactive elements.

### DIFF
--- a/packages/react-components/src/components/Checkbox/Checkbox.module.scss
+++ b/packages/react-components/src/components/Checkbox/Checkbox.module.scss
@@ -9,7 +9,14 @@
     width: 100%;
   }
 
-  &__square {
+  &__text {
+    flex: 1 0 auto;
+    margin-left: 8px;
+    color: var(--content-basic-primary);
+  }
+
+  &__input {
+    appearance: none;
     box-sizing: border-box;
     display: flex;
     position: relative;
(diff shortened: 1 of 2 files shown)


packages/react-components/src/components/Checkbox/Checkbox.module.scss
```diff
@@ -21,44 +28,28 @@
...
+      opacity: 0;
+      z-index: 1;
+      background-color: var(--color-white);
+      width: 24px;
+      height: 24px;
+      content: '';
+      clip-path: path(
+        'M9 16.2001L5.5 12.7001C5.11 12.3101 4.49 12.3101 4.1 12.7001C3.71 13.0901 3.71 13.7101 4.1 14.1001L8.29 18.2901C8.68 18.6801 9.31 18.6801 9.7 18.2901L20.3 7.70007C20.69 7.31007 20.69 6.69007 20.3 6.30007C19.91 5.91007 19.29 5.91007 18.9 6.30007L9 16.2001Z'
+      );
+      scale: 0.5;
```
marcinsawicki, line 45: It won't be easier to change the background color of the input and just show the icon (positioned by CSS) instead of a that pseudo element? 🤔

ashbork: How would you see that? I think having one less thing to worry about in the markup is a better approach. I suppose both the Checkbox and the icon could be siblings, positioned absolutely on top of each other, but this solution feels better to me, WDYT?

sgraczyk:
> The Chromatic snapshot seems to be wrong, the actual render works correctly

image

You can click `1up` mode in snapshot comparison and you can observe that there are slight differences:
- the default state changed to `checked`. I think it should remain unchecked.
- The new one seems to be a bit shifted to the right and paddings seems a bit greater.

packages/react-components/src/components/Checkbox/Checkbox.module.scss
marcinsawicki: Yes, I agree that one less thing in markup will be better.
How about to use the checkmark icon directly in the `content` of that pseudo-element?
`content: url([path]/icon.svg);`

ashbork: I wanted to do that in the first place, would you be able to help me out? I'm not sure where to point the path

marcinsawicki: It could be something like this `content: url('./check.svg');` and icon file should be placed in the same place where we keep the styles for this component. Of course, the icon should have the correct color in every mode, so we must change the `fill` or `stroke` param to token: `--content-invert-primary`.

marcinsawicki:
Currently, we can test it only manually because Chromatic isn't mandatory in case of testing our component. We use it to help us catch the changes, but as we have only a free plan only, the snapshot limit could be reached in case of many changes.

ashbork:
It looks like some autoformatter/linter is breaking the `scss` module on commit; specifically by changing prefixed `-webkit-mask` to `mask`.

ashbork:
I believe the spacing changes should now be fixed

marcinsawicki (approved):
I think that this approach will give us more value than before, and still, we keep the primary design.
Nice job 👍

ashbork:
Are there plans to merge this sometime soon?

Issue #573, opened 2023-06-01
Popover's closeOnEsc prop does not react to dynamic changes


norbertprocak:
When I change `closeOnEsc` property in `Popover` from `true` to `false`, the change is ignored. Only the initial value is working.

PR #584, opened 2023-06-02
573: reacting properly on closeOnEsc prop change

VadymBezpalko:
Resolves: #573
## Description
Adding closeOnEsc to the dependencies array of a hook to properly react on prop changes

## Storybook
https://feature-573--613a8e945a5665003a05113b.chromatic.com/

## Checklist

**Obligatory:**

- [x] Self-review
- [ ] Storybook cases
- [ ] Design review
- [ ] Functional (QA) review

**Optional:**

### DIFF
--- a/packages/react-components/src/components/Popover/Popover.tsx
+++ b/packages/react-components/src/components/Popover/Popover.tsx
@@ -106,29 +106,20 @@ export const Popover: React.FC<IPopoverProps> = ({
   }
 
   const handleHideOnEscape = (event: KeyboardEvent) => {
-    if (event.key === 'Escape') {
+    if (closeOnEsc && event.key === 'Escape') {
       setVisibility(false);
     }
   };
 
   React.useEffect(() => {
     document.addEventListener('mousedown', handleDocumentClick);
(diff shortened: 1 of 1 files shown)


packages/react-components/src/components/Popover/Popover.tsx
```diff
@@ -113,14 +113,22 @@ export const Popover: React.FC<IPopoverProps> = ({
...
+  }, []);
+
+  React.useEffect(() => {
     if (closeOnEsc) {
       document.addEventListener('keydown', handleHideOnEscape);
     }
     return () => {
-      document.removeEventListener('keydown', handleHideOnEscape);
-      document.removeEventListener('mousedown', handleDocumentClick);
+      if (closeOnEsc) {
```
norbertprocak, line 127: I think when you set `closeOnEsc` dynamically, the handler will not be removed. If I were you, I would always add the handler to `keydown` but return from handler if current value of `closeOnEsc` is false.

VadymBezpalko: Great idea! Thanks 👍

marcinsawicki (approved):

Issue #556, opened 2023-05-19
[Modal] - Fixes


marcinsawicki:
## Description
Modal footer - Top/Bottom spacing should be 16px

To consider whether footer and header should not be separate components

## Design
https://www.figma.com/file/2pFu80PXO5A2tfyrAGnx91/Product-Components?type=design&node-id=6904-33955&t=sOA1xqxapwnqCBou-4

Issue #551, opened 2023-05-18
[Date picker] - Fixes


marcinsawicki:
## Description
Spacing 4px between `>` and `>>` icons

## Design
https://www.figma.com/file/2pFu80PXO5A2tfyrAGnx91/Product-Components?type=design&node-id=3484-19051&t=nj3kPivwZOmLrwKq-4

PR #586, opened 2023-06-02
#551 - Modal - seperate header and footer components for modal

jpyzio123:
Resolves: #556
## Description

Fix padding in footer component and new ModalHeader component to speed up the implementation of the Modal component. Backwards compatibility is covered, no changes required by current implementations. 

## Storybook
https://modal-fixes-v3--613a8e945a5665003a05113b.chromatic.com/

## Checklist

**Obligatory:**

- [x] Self-review
- [x] Storybook cases
- [ ] Design review
- [ ] Functional (QA) review

**Optional:**


packages/react-components/src/components/Modal/Modal.stories.tsx
```diff
@@ -29,45 +28,65 @@ export default {
 } as ComponentMeta<typeof ModalComponent>;
 
 const defaultModalProps = {
-  children: <ModalContent />,
+  // children: <ModalContent />,
```
marcinsawicki, line 31: Why comment?

packages/react-components/src/components/Modal/Modal.tsx
```diff
@@ -1,27 +1,22 @@
...
+import { ModalCloseButton } from './ModalCloseButton';
 
 import styles from './Modal.module.scss';
-import { ModalCloseButton } from './ModalCloseButton';
 
 export interface ModalProps extends ModalBaseProps {
-  heading?: React.ReactNode;
-  labelHeading?: React.ReactNode;
   fullSpaceContent?: boolean;
-  footer?: React.ReactNode;
```
marcinsawicki, line 13: I'm afraid that we can't just remove these props from the interface and put in the new version.
We should consider to keep it, and also use atoms in the component `children`.

packages/react-components/src/components/Modal/ModalHeader.tsx
```diff
@@ -0,0 +1,39 @@
...
+  return (
+    <div className={mergedClassNames}>
+      {iconProps && (
+        <Icon className={styles[`${baseClass}__heading-icon`]} {...iconProps} />
+      )}
+      <div className={styles[`${baseClass}__heading-header`]}>
+        <div>{title}</div>
+
+        {children && (
+          <div className={styles[`${baseClass}__heading-description`]}>
```
VadymBezpalko, line 32: Maybe we can use our typography here (Text component)?

packages/react-components/src/components/Modal/ModalHeader.tsx
```diff
@@ -0,0 +1,39 @@
...
+  className = '',
+}) => {
+  const mergedClassNames = cx(styles[baseClass], className);
+
+  return (
+    <div className={mergedClassNames}>
+      {iconProps && (
+        <Icon className={styles[`${baseClass}__heading-icon`]} {...iconProps} />
+      )}
+      <div className={styles[`${baseClass}__heading-header`]}>
```
VadymBezpalko, line 28: Why is the header component wrapping the description? Or should it be named differently (I'm writing about class name)?

packages/react-components/src/components/Modal/ModalFooter.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+interface ModalFooterProps {
+  children: React.ReactNode;
+  className?: string;
+}
+
+export const ModalFooter: React.FC<ModalFooterProps> = ({
```
VadymBezpalko, line 7: I'm not quite sure what the purpose of this component is. At the moment, it doesn't seem to offer any benefits to the developer. Are there any plans to include additional styling or features in the future?

VadymBezpalko (commented):
Great job! Please take a look at the comment I left. Additionally, I have a question about TODO bullet points. If this is just a POC and not the final implementation, maybe it will be better to have it on a separate branch until the final version is ready? WDYT?

packages/react-components/src/components/Modal/ModalFooter.tsx
marcinsawicki: I know that creating separate components for the header and footer was my suggestion, but after thinking about this, in the current state, that footer component doesn't bring any benefit, so I believe we can drop that pattern.
Maybe for the moment it will be enough if you can pass the class to the footer to be able to style it?
WDYT? @VadymBezpalko

VadymBezpalko: @marcinsawicki ok 👌

marcinsawicki (approved):

PR #623, opened 2023-06-23
Firing open-close callbacks for managed components

VadymBezpalko:
## Description
Firing open-close callbacks for managed components

## Storybook
https://feature-tooltip-callbacks-improvement--613a8e945a5665003a05113b.chromatic.com/

## Checklist

**Obligatory:**

- [x] Self-review
- [ ] Storybook cases
- [ ] Design review
- [ ] Functional (QA) review

**Optional:**

### DIFF
--- a/packages/react-components/src/components/Tooltip/Tooltip.tsx
+++ b/packages/react-components/src/components/Tooltip/Tooltip.tsx
@@ -91,7 +91,9 @@ export const Tooltip: React.FC<ITooltipProps> = (props) => {
     } else {
       visible && onClose?.();
     }
-    setVisibility(newVisibility);
+    if (!isManaged) {
+      setVisibility(newVisibility);
+    }
   };
 
   React.useEffect(() => {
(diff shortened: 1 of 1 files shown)


MichalPaszowski (approved):

packages/react-components/src/components/Tooltip/Tooltip.tsx
```diff
@@ -91,7 +91,9 @@ export const Tooltip: React.FC<ITooltipProps> = (props) => {
     } else {
       visible && onClose?.();
```
MichalPaszowski, line 92: shouldn't it fire regardless of the fact that tooltip is managed or not?

VadymBezpalko: I've modified this component, so callback will be fired whenever `isVisible` value is changing

JCown (approved):
IMO this component gets a bit complicated

VadymBezpalko:
> IMO this component gets a bit complicated

I agree. We will work on refactoring and improving Tooltip component 👌

VadymBezpalko:
I created a task for the Tooltip refactor in our board #624

Issue #651, opened 2023-07-14
[Picker] - Multiple picker logic fix


vladko-uxds:
Problem:
Current picker does not clear search and list with results with each selection. So user need to change focus or click outside of picker, and go back again to reset Picker list view to default state
https://user-images.githubusercontent.com/125188605/253559669-41637d23-f9c3-4363-8a1b-b84233d99b3e.mp4

Soulution:
- Default behaviour "Clear search with each selection".
- Search field after each selection will be cleared/replaced with a selected value/tag automatically
- List of results after selection return to initial view (selected items should be in selected state)

https://user-images.githubusercontent.com/125188605/253573576-3c2fbd77-3aea-4872-960d-98d27ba2e819.mp4

PR #658, opened 2023-07-18
#651 - Picker - Auto clear search option

marcinsawicki:
Resolves: #651 

## Storybook

https://feature-651--613a8e945a5665003a05113b.chromatic.com/?path=/story/components-picker--default&args=type:multi;clearSearchAfterSelection:!true

## Checklist

**Obligatory:**

- [x] Add correct label

### DIFF
--- a/packages/react-components/src/components/Picker/TriggerBody.tsx
+++ b/packages/react-components/src/components/Picker/TriggerBody.tsx
@@ -1,5 +1,4 @@
 import * as React from 'react';
-import cx from 'clsx';
 
 import { IPickerListItem } from './PickerList';
 import { Tag } from '../Tag';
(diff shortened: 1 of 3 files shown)


packages/react-components/src/components/Picker/TriggerBody.tsx
```diff
@@ -30,10 +31,22 @@ export const TriggerBody: React.FC<ITriggerBodyProps> = ({
...
+
+  React.useEffect(() => {
+    if (clearSearchAfterSelection) {
+      onFilter('');
+
+      if (inputRef.current) {
+        inputRef.current.value = '';
+      }
+    }
+  }, [items, clearSearchAfterSelection]);
```
VadymBezpalko, line 49: Is it intended that `items` are in the dependency array and `onFilter` is not?

marcinsawicki: Yes. `items` will change after selection so it should triggers `useEffect`. If we put `onFliter` it will break search functionality as this handler is triggered when search input value changes.

vladko-uxds (approved):
🤘

Issue #566, opened 2023-05-23
[ActionMenu] - prop for controlling the close behaviour


marcinsawicki:
## Description
For Action Menu component we need to have the additional property for Menu Item to control “close ActionMenu on click or not close (ON/OFF)“.
So the users when they need can Turn ON  this property for case when an action menu should not close on click (Checkbox/RadioButton or Toggle etc)

PR #678, opened 2023-07-21
#566 - ActionMenu - Option to prevent closing when option is selected

marcinsawicki:
Resolves: #566 

## Description
Added optional prop that will prevent closing the menu after the item click. It can be used to put the form-like elements to the menu like radio buttons or switch.
I also adjusted the `Radio button` styles to fit the design if implemented inside the menu.

## Storybook

https://feature-566--613a8e945a5665003a05113b.chromatic.com/?path=/story/components-actionmenu--keep-open-on-item-click

## Checklist

**Obligatory:**

- [x] Add correct label

### DIFF
--- a/packages/react-components/src/components/ActionMenu/ActionMenu.tsx
+++ b/packages/react-components/src/components/ActionMenu/ActionMenu.tsx
@@ -92,7 +92,7 @@ export const ActionMenu: React.FC<ActionMenuProps> = ({
   const handleItemClick = (itemOnClick: () => void) => {
     itemOnClick();
 
-    if (keepOpenOnClick !== true) {
+    if (!keepOpenOnClick) {
       setIsVisible(false);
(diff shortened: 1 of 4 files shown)


packages/react-components/src/components/ActionMenu/ActionMenu.tsx
```diff
@@ -86,14 +91,16 @@ export const ActionMenu: React.FC<ActionMenuProps> = ({
 
   const handleItemClick = (itemOnClick: () => void) => {
     itemOnClick();
-    setIsVisible(false);
+
+    if (keepOpenOnClick !== true) {
```
VadymBezpalko, line 95:
```suggestion
    if (!keepOpenOnClick) {
```

VadymBezpalko (approved):

Issue #681, opened 2023-07-24
[Buttons] - Updated border radius for all buttons


vladko-uxds:
Please update border radius for all buttons:

Token: `-radius-3`

Components:
- Primary
- Tertiary
- Subtle
- Float
- Destructive
- Icons only button

PR #691, opened 2023-07-24
#681 - Button - Updated button border radius

marcinsawicki:
Resolves: #681 

## Description
Used correct `radius-token` for border radius.

## Storybook

https://feature-680--613a8e945a5665003a05113b.chromatic.com/?path=/story/components-button--kinds-and-states

## Checklist

**Obligatory:**

- [x] Add correct label

### DIFF
--- a/packages/react-components/src/foundations/radius.css
+++ b/packages/react-components/src/foundations/radius.css
@@ -1,6 +1,6 @@
 :root {
-  --spacings-border-radius-none: 0;
-  --spacings-border-radius-xs: 4px;
-  --spacings-border-radius-sm: 6px;
-  --spacings-border-radius-md: 8px;
+  --radius-0: 0;
+  --radius-1: 4px;
+  --radius-2: 6px;
+  --radius-3: 8px;
 }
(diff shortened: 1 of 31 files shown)


vladko-uxds (approved):

packages/react-components/src/foundations/radius.css
```diff
@@ -0,0 +1,6 @@
+:root {
+  --spacings-border-radius-none: 0;
+  --spacings-border-radius-xs: 4px;
+  --spacings-border-radius-sm: 6px;
+  --spacings-border-radius-md: 8px;
```
VadymBezpalko, line 5: We already have a `spacing` keyword used in a singular form. Maybe we should use the singular form here as well, for consistency? WDYT? @marcinsawicki @vladko-uxds

marcinsawicki: I agree with that, but we need to unify the naming with our documentation. CC: @vladko-uxds

VadymBezpalko (approved):

Issue #685, opened 2023-07-24
[Picker list] - Border radius and new styles


vladko-uxds:
Please update border radius for Picker list component
https://v1--613a8e945a5665003a05113b.chromatic.com/?path=/story/components-picker--default

List border radius:
`--radius-3`

Please update picker list items states
https://www.figma.com/file/2pFu80PXO5A2tfyrAGnx91/Product-Components?type=design&node-id=2192-19603&mode=design&t=hCer9WgZwKA8nUjV-4

Design spec
https://www.figma.com/file/2pFu80PXO5A2tfyrAGnx91/Product-Components?type=design&node-id=2329-20904&mode=design&t=hCer9WgZwKA8nUjV-4

PR #695, opened 2023-07-24
#685 - Picker - UI refresh

marcinsawicki:
Resolves: #685 

## Description
Updated UI for the `Picker` component.

## Storybook

https://feature-685--613a8e945a5665003a05113b.chromatic.com/?path=/docs/components-picker--docs

## Checklist

**Obligatory:**

- [x] Add correct label


vladko-uxds:
Right top/bottom corner should also have
Monosnap Components : Picker - Docs ⋅ Storybook 2023-07-24 15-05-07
 border radius

VadymBezpalko (approved):

packages/react-components/src/components/Picker/PickerListItem.tsx
```diff
@@ -0,0 +1,91 @@
...
+      role="option"
+      aria-current={currentItemKey === item.key}
+      aria-selected={isItemSelected}
+      aria-disabled={item.disabled}
+      id={item.key}
+      key={item.key}
+      className={cx(styles[itemClassName], {
+        [styles[`${itemClassName}__custom`]]: item?.customElement,
+      })}
+      onClick={() => !item.disabled && handleOnClick(item)}
```
VadymBezpalko, line 77: we can extract this condition into the handler function, and this part can simply look like this
```suggestion
      onClick={handleOnClick}
```

VadymBezpalko (approved):
Good job, I left some comments. And what do you think about placing these additional components (which are not exported in `intex.ts`) into `components` folder?

PR #699, opened 2023-07-26
Upgrading Lerna and Storybook versions

VadymBezpalko:
Upgrading Lerna from 4.0.0 to 7.x.x version caused some additional changes:
- now Lerna is no longer responsible for updating and linking dependencies; they are managed by NPM workspaces directly
- there is a strange case: Storybook detects React v18, and for that, I've added `options: {legacyRootApi: true}` into `.storybook/main.js` because otherwise, Storybook throws error:`m.createRoot is not a function`.

Additionally, I've upgraded SB to a new version and enabled `Lerna caching` mechanism (powered by Nx)


sgraczyk (approved):
Looks fine.

Can you check if this document is still up to date after these changes?
https://github.com/livechat/design-system/blob/main/docs/RELEASE.md

marcinsawicki (approved):

PR #701, opened 2023-07-27
Updating TSConfig and ESLint rules

VadymBezpalko:
## Description

Updating TSConfig and ESLint rules for a better developer experience.

## Checklist

**Obligatory:**

- [ ] Add correct label


.eslintrc
```diff
@@ -15,11 +16,29 @@
   },
   "plugins": ["@typescript-eslint"],
   "rules": {
+    "eslint-comments/no-unused-disable": "error",
     "prefer-const": "error",
     "@typescript-eslint/no-unsafe-member-access": "off",
     "@typescript-eslint/no-unsafe-call": "off",
     "@typescript-eslint/no-unsafe-assignment": "off",
     "@typescript-eslint/no-unsafe-return": "off",
```
sgraczyk, line 24: Can we validate if these should still be off?

sgraczyk (approved):

marcinsawicki (approved):

mreszke (approved):

.eslintrc
```diff
@@ -15,11 +16,29 @@
...
-    "@typescript-eslint/restrict-template-expressions": "off"
+    "@typescript-eslint/restrict-template-expressions": "off",
+    "no-console": "error",
+    "no-undef": "off",
```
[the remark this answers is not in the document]
VadymBezpalko: I've enabled this rule and revised the code accordingly 👍

.eslintrc
VadymBezpalko: @sgraczyk I went through these rules and enabled the rules that could be applied to the project ✅.

But there are issues with `no-unsafe-member-access` and `no-unsafe-assignment`. Our `design-system-icons` package does not generate proper types for icons, and because of this, TypeScript is unable to match the type from the `source` prop in the `<Icon>` component with generated icon components from the icons library. I've marked these two rules with TODO for this PR.

image

Issue #696, opened 2023-07-25
[Avatar] - New size variant and naming(breaking changes)


vladko-uxds:
Please add 
- Avatar 40x40px , size name: " Large"

Note: Other "Larger" avatars should change their names accordingly
- Large(48px) -> XLarge
- XLarge (64px) -> 2XLarge
- 2XLarge (96px) -> 3XLarge

See design 
https://www.figma.com/file/2pFu80PXO5A2tfyrAGnx91/Product-Components?type=design&node-id=736-8527&mode=design&t=rE9GHkMx29a6CsOG-4

marcinsawicki:
Blocked until we release the new version with fixes, because here we implement the breaking changes that may cause problems in currently implemented components

PR #705, opened 2023-07-28
#696 - Avatar - Changes in avatar sizes

marcinsawicki:
Resolves: #696 

## Description
Changed the sizes of the component according to the design documentation. Added `xxxlarge` size, values from `large` has been moved to the next size, and `large` has the new value.
This is the breaking change for every component usage with size `large` and above.

## Storybook

https://feature-696--613a8e945a5665003a05113b.chromatic.com/?path=/story/components-avatar--sizes-with-status

## Checklist

**Obligatory:**

- [x] Add correct label


VadymBezpalko (approved):

vladko-uxds (approved):
👍

Issue #709, opened 2023-07-31
[PromoBanner] - New version for promo banner


vladko-uxds:
Please prepare new banner component

Design spec
https://www.figma.com/file/2pFu80PXO5A2tfyrAGnx91/Product-Components?type=design&node-id=16225-64174&mode=design&t=87exqnbMGHsDBP6D-4

**NOTE**
- New banner should not replace current PromoBanner, It should be added as an additional component. 
- The current Promo banner should be market as "outdated" but exist in a DS.

PR #716, opened 2023-08-03
#709 - PromoBannerV2 - New component

marcinsawicki:
Resolves: #709 

## Description
New version of `PromoBanner` component, that will be called as `PromoBannerV2`.
Old `PromoBanner` will stay in lib and will be marked as deprecated.

## Storybook

https://feature-709--613a8e945a5665003a05113b.chromatic.com/?path=/docs/components-promobannerv2--docs

## Checklist

**Obligatory:**

- [ ] Add correct label


sgraczyk:
Consider informing about `PromoBanner` being deprecated and that `PromoBannerV2` should be used.

Possible approaches:

1. Using `useEffect`

```tsx
useEffect(() => {
        console.warn(
            "Deprecation Notice: The PromoBanner component is deprecated and will be removed in the future. " +
            "Please migrate to PromoBannerV2. Refer to the migration guide at [LINK_TO_GUIDE]."
        );
    }, **[]);**
```

2. Using `JSDoc`

```tsx
/**
 * @deprecated Since version X.Y.Z. Will be deleted in version Z.Y.X. Use {@link PromoBannerV2} instead.
 */
const PromoBanner: React.FC<PromoBannerProps> = (props) => {
    // rest of the code
};
```

packages/react-components/src/components/PromoBannerV2/PromoBannerV2.module.scss
```diff
@@ -0,0 +1,71 @@
...
+    }
+  }
+
+  &__content {
+    display: flex;
+    flex: 0 1 720px;
+    flex-flow: column;
+    justify-content: center;
+    padding: var(--spacing-5);
+    max-width: 720px;
```
VadymBezpalko, line 34: I think, that we can create a variable (e.g. `$promo-banner-content-max-width`) to use here and a few lines above.

VadymBezpalko (approved):
Good job! 
I have one more question: shouldn't the additional content be aligned to the right? Because for now, it looks a little bit odd on long viewports.
image

packages/react-components/src/components/PromoBannerV2/PromoBannerV2.module.scss
vladko-uxds: + $promo-banner-height-min (to manage **height** for Horizontal banner)

packages/react-components/src/components/PromoBannerV2/PromoBannerV2.module.scss
```diff
@@ -0,0 +1,78 @@
...
+      background: 0;
+      padding: 0;
+
+      &:hover {
+        cursor: pointer;
+      }
+    }
+  }
+
+  @container (max-width: 559px) {
```
VadymBezpalko, line 60: I'd suggest putting this number into `constants.ts` and later reuse it here and in `.mdx` file

packages/react-components/src/components/PromoBannerV2/PromoBannerV2.mdx
```diff
@@ -0,0 +1,80 @@
...
+  justify-content: flex-end;
+  align-items: flex-end;
+
+  @container (max-width: 559px) {
+    justify-content: center;
+  }
+}
+```
+
+The above example will cause the transferred element to fill the container in which it was placed, and the layout change will be on the side of implementation in your code. Using the `container query` set to the breakpoint indicated by us (<b>559px</b>), the element will be moved accordingly in accordance with the provided styles.
```
VadymBezpalko, line 67: Please use here a constant from constants.ts

packages/react-components/src/components/PromoBannerV2/PromoBannerV2.mdx
```diff
@@ -0,0 +1,80 @@
...
+      Title goes here, up 2 lines of text
+    </Heading>
+  </div>
+  A description with a <b>maximum of 100 characters</b>. That usually means
+  only one or two sentences.
+</PromoBannerV2>
+```
+
+##### How to control the elements
+By default, items in the left column are centered, while items in the right column are aligned to the bottom right. The component contains a breakpoint depending on the width of the main container, which, if it is <b>less than 559px</b>, will change the view to "vertical" and the contents of the right column will be moved above the left column. In "vertical" view, the contents of the right column will be left-aligned by default, the left column will remain unchanged. If you want more control of these elements (left and right columns) in both views, below is an instruction on how to control the position of individual elements via CSS styles.
```
VadymBezpalko, line 41: and here

VadymBezpalko (commented):
Please check the sizes of the illustrations for PromoBannerV2 storybook - currently the newly added `Illustration.svg` is around 1.02 MB, which is too big. Additionally we need to check our building process, cause the building library in library mode in Vite can have problems with tree-shaking. 
Also, please, add the information with the JSDoc to the old PromoBanner component, that it will be marked as deprecated

vladko-uxds (approved):
👍

PR #733, opened 2023-08-17
[Action Bar] - MVP experimental new component

marcinsawicki:
## Description
MVP of the new component called `ActionBar`. (experimental)

The purpose of this component is to display the options in the form of buttons. Contrary to `SegmentedControl` this component is supposed to adapt to the container in which it is embedded and in a situation where there is not enough space to display all buttons, those that lack space will be hidden and available in the form of options in the `ActionMenu`.

The component contains an observer that will hide or show additional buttons in case of changing the width of the container, depending on the available space. There is also an option `type: 'scroll'` that disables the menu and enables the scrolling in the main container.

Remember that this is an **experimental** component and may not meet all user requirements at the moment. It will be extended in the future so that it can be better adapted to the design.

## Storybook

https://feature-action-bar-mvp--613a8e945a5665003a05113b.chromatic.com/?path=/docs/experimental-actionbar--docs

## Checklist

**Obligatory:**

- [x] Add correct label


packages/react-components/src/components/ActionBar/ActionBar.tsx
```diff
@@ -0,0 +1,190 @@
...
+      }
+
+      entry.target.removeAttribute('tabindex');
+
+      const newMenuItems = menuItemsKeys;
+
+      if (newMenuItems.includes(entry.target.id)) {
+        const index = newMenuItems.indexOf(entry.target.id);
+        newMenuItems.splice(index, 1);
+        setMenuItemsKeys([...newMenuItems]);
```
VadymBezpalko, line 82: What is the purpose of a spread operator here?

packages/react-components/src/components/ActionBar/ActionBar.tsx
```diff
@@ -0,0 +1,190 @@
...
+  };
+
+  return (
+    <div id={id} className={mergedClassNames}>
+      <div
+        className={cx(styles[`${baseClass}__items`], {
+          [styles[`${baseClass}__items--scroll`]]: isScrollType,
+        })}
+      >
+        {options.map((o) => getItem(o))}
```
VadymBezpalko, line 178:
```suggestion
        {options.map(getItem)}
```

packages/react-components/src/components/ActionBar/ActionBar.tsx
```diff
@@ -0,0 +1,190 @@
...
+        option.key
+      ),
+      [styles[`${baseClass}__items__button--active`]]:
+        option.key === activeOptionKey,
+    });
+
+    if (option.label) {
+      const tooltipVisibility = menuItemsKeys.includes(option.key)
+        ? false
+        : undefined;
```
VadymBezpalko, line 132: We should not set values as `undefined` by ourselves, as it indicates a very specific case when the value is not initialized. If we want to declare 'emptiness' of a value we can use `null` for example. But in the current case I'd rather expect values more like `true` or `false`.

packages/react-components/src/components/ActionBar/ActionBar.tsx
```diff
@@ -0,0 +1,190 @@
...
+        key: o.key,
+        element: (
+          <ActionMenuItem leftNode={o.element}>{o.label}</ActionMenuItem>
+        ),
+        onClick: o.onClick,
+      };
+    });
+  };
+
+  const getItem = (option: IActionBarOption) => {
```
VadymBezpalko, line 120: I think we can make this item a standalone component to separate the codebase of these two entities, wdyt?

packages/react-components/src/components/ActionBar/ActionBar.tsx
```diff
@@ -0,0 +1,190 @@
...
+  return (
+    <div id={id} className={mergedClassNames}>
+      <div
+        className={cx(styles[`${baseClass}__items`], {
+          [styles[`${baseClass}__items--scroll`]]: isScrollType,
+        })}
+      >
+        {options.map((o) => getItem(o))}
+      </div>
+      {!isScrollType && menuItemsKeys.length !== 0 && (
```
VadymBezpalko, line 180: We can put this condition into a constant.

VadymBezpalko:
And regarding the visual aspects, can we decrease the "gap" between normal icons and menu icon button? Because now it looks like there is no more icons there. 
image

packages/react-components/src/components/ActionBar/ActionBar.tsx
```diff
@@ -0,0 +1,190 @@
...
+    if (!isScrollType) {
+      const target = document.querySelectorAll(
+        `.${styles[`${baseClass}__items__button`]}`
+      );
+
+      target.forEach((e) => observer.observe(e));
+    }
+
+    return () => observer.disconnect();
+  }, []);
```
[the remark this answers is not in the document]
VadymBezpalko: I think we can create a custom hook for this IntersectionObserver. Something like this for example: https://usehooks-ts.com/react-hook/use-intersection-observer

vladko-uxds (approved):

packages/react-components/src/components/ActionBar/ActionBar.tsx
```diff
@@ -0,0 +1,132 @@
...
+      entry.target.removeAttribute('tabindex');
+
+      if (menuItemsKeys.includes(entry.target.id)) {
+        setMenuItemsKeys(menuItemsKeys.filter((i) => i !== entry.target.id));
+      }
+    });
+  };
+
+  React.useEffect(() => {
+    if (!isScrollType) {
```
VadymBezpalko, line 75:
```suggestion
    const hasIOSupport = !!window.IntersectionObserver;
    if (!isScrollType && hasIOSupport) {
```

packages/react-components/src/components/ActionBar/ActionBar.tsx
```diff
@@ -0,0 +1,132 @@
...
+    threshold: 1,
+  };
+  const shouldDisplayMenu = !isScrollType && menuItemsKeys.length !== 0;
+
+  const handleIntersection = (entries: IntersectionObserverEntry[]) => {
+    entries.map((entry) => {
+      if (!entry.isIntersecting) {
+        entry.target.setAttribute('tabindex', '-1');
+
+        if (!menuItemsKeys.includes(entry.target.id)) {
```
VadymBezpalko, line 59: To avoid double-checking the same condition (here and next `if`, in the line 68) we can store this condition in a separate constant or use `if-else` statement.

packages/react-components/src/components/ActionBar/ActionBar.tsx
```diff
@@ -0,0 +1,132 @@
...
+      const observer = new IntersectionObserver(
+        handleIntersection,
+        observerOptions
+      );
+
+      target.forEach((e) => observer.observe(e));
+
+      return () => observer.disconnect();
+    }
+  }, [menuItemsKeys]);
```
VadymBezpalko, line 89: What about `isScrollType` prop? Maybe we should also include it here?

VadymBezpalko (approved):

packages/react-components/src/components/ActionBar/ActionBar.tsx
```diff
@@ -0,0 +1,139 @@
...
+   */
+  id?: string;
+  /**
+   * Array of action bar options
+   */
+  options: IActionBarOption[];
+  /**
+   * Set the key for active element
+   */
+  activeOptionKey?: string | null;
```
mreszke, line 29: do we need this `Key` suffix?
```suggestion
  activeOption?: string | null;
```

mreszke (commented):
IMO we should display the tooltip in opposed direction to dropdown, or even disable it when dropdown is open
image

packages/react-components/src/components/ActionBar/ActionBar.tsx
marcinsawicki: Name references to the part of the `option` object, so the `key` suffix is correct for me.

PR #742, opened 2023-08-24
Auto assign issues to the project

VadymBezpalko:
## Description
Auto assigning new feature requests / bug reports to the DS project.


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,61 @@
...
+        3. Scroll down to '....'
+        4. See an error
+    validations:
+      required: false
+  - type: textarea
+    attributes:
+      label: Expected Behavior
+      description: A clear and concise description of what you expected to happen.
+    validations:
+      required: false
```
mreszke, line 36: IMO it should be required.
I'm also missing the `actual behavior`

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,61 @@
...
+      description: |
+        examples:
+          - **Design system package(s) version**: [e.g. @livechat/design-system-react-components@1.0.0-beta.11]
+          - **Device**: [e.g. iPhone6]
+          - **OS**: [e.g. macOS]
+          - **Browser**: [e.g. chrome, safari]
+          - **Version**: [e.g. 22]
+      render: markdown
+    validations:
+      required: false
```
mreszke, line 55: IMO the package version is also important and should be required. Maybe it should be a separate field?

VadymBezpalko: Good point. I'll add it separately.

.github/ISSUE_TEMPLATE/bug_report.yml
VadymBezpalko: Ok, I will make it as required.
Regarding `actual behavior` - I thought that this kind of information would be included in the bug description and steps to reproduce sections.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,69 @@
...
+    attributes:
+      label: Steps To Reproduce
+      description: Steps to reproduce the behavior.
+      placeholder: |
+        1. Go to '...'
+        2. Click on '....'
+        3. Scroll down to '....'
+        4. See an error
+    validations:
+      required: false
```
mreszke, line 30: would set that as required also
```suggestion
      required: true
```

mreszke (approved):

PR #753, opened 2023-08-29
Avatar background color handling based on initials

sgraczyk:
Resolves: [Slack thread](https://text.slack.com/archives/C05G7GS1HB9/p1691501762305939)

## Description
It was decided that background color of the `Avatar` should have a default value if it uses `text` `type`. It uses a basic algorithm that makes sure that a set of 10 colors are used, and for the exact same initials the color will be always the same.

❗️ `Attention:` This change brings some sort of breaking change as it will apply new color in for avatars using `text` `type` without `color` provided. By default it was `grey`.

I've also enabled the possibility of using `CSS variable` for `color` prop as earlier it was impossible due to contrast calculation requiring direct values. It was solved by using `window.getComputedStyle`

image

## Storybook

https://feature-avatar-default-backgrounds--613a8e945a5665003a05113b.chromatic.com/?path=/story/components-avatar--colors

## Checklist

**Obligatory:**

- [x] Add correct label

### DIFF
--- a/packages/react-components/src/components/Avatar/Avatar.helpers.spec.ts
+++ b/packages/react-components/src/components/Avatar/Avatar.helpers.spec.ts
@@ -6,14 +6,14 @@ import {
 
 describe('getBackgroundColor', () => {
   it('should generate the same background color for the same initials', () => {
-    const color1 = getBackgroundColor('John Doe');
-    const color2 = getBackgroundColor('John Doe');
+    const color1 = getBackgroundColor('JD');
+    const color2 = getBackgroundColor('JD');
     expect(color1).toBe(color2);
   });
 
   it('should generate different background colors for different initials', () => {
-    const color1 = getBackgroundColor('John Doe');
-    const color2 = getBackgroundColor('Jane Doe');
+    const color1 = getBackgroundColor('JD');
+    const color2 = getBackgroundColor('AB');
     expect(color1).not.toBe(color2);
   });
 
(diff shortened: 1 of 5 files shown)


mreszke (approved):

rafalluksza (approved):

VadymBezpalko (approved):

vladko-uxds (approved):

packages/react-components/src/components/Avatar/Avatar.helpers.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+  });
+
+  it('should handle empty initials and return undefined', () => {
+    const color = getBackgroundColor('');
+    expect(color).toBeUndefined();
+  });
+});
+
+describe('getInitials', () => {
+  it('should return initials for a full name', () => {
```
shwarcu, line 27: It used to be possible to set emoji as customer name, consider checking if such unusual input is handled

sgraczyk: I think it's not handled in here. I believe there is such an implementation in LiveChat Web

```ts
export function getInitials(name = ''): string {
  const isEmojiInName = EMOJI_REGEX.test(name);
  const finalName = name ? name : '';
  let initials = '';

  initials = finalName
    .split(' ')
    .map((el) => el.charAt(0))
    // eslint-disable-next-line @typescript-eslint/no-unsafe-call
    .filter((el) => isAlphaNumeric(removeDiacritics(el)))
    .join('')
    .substring(0, 2)
    .toUpperCase();

  // fallback for emoji-only and non-alpha-numeric characters
  if (!initials) {
    initials = isEmojiInName
      ? finalName.match(EMOJI_REGEX)?.[0] || ''
      : finalName.toLocaleUpperCase().trim().substring(0, 2);
  }

  return initials;
}
```

@vladko-uxds Do we want to support `emojis` somehow in `Design System`?

PR #772, opened 2023-09-11
Fix for import issue in Icon component

VadymBezpalko:
## Description
Fix for import issue in Icon component

## Checklist

**Obligatory:**

- [x] Add correct label

### DIFF
--- a/packages/react-components/src/components/Picker/types.ts
+++ b/packages/react-components/src/components/Picker/types.ts
@@ -1,6 +1,6 @@
 import * as React from 'react';
 
-import { IconSource } from 'components/Icon';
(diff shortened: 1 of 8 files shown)


packages/react-components/src/components/Picker/types.ts
```diff
@@ -1,6 +1,6 @@
 import * as React from 'react';
 
-import { IconSource } from 'components/Icon';
+import { IconSource } from '../Icon';
```
mreszke, line 3:
```suggestion
import { IconSource } from '../Icon/types';
```

mreszke (approved):

marcinsawicki (approved):

PR #775, opened 2023-09-12
[ActionMenu] - fixes after RefreshUI tests

marcinsawicki:
## Description
Changed the trigger implementation, it uses the default `button` element, which wraps the element, with the possibility to modify it by `triggerClassName` prop.
Also added `groupHeader` prop to `options` to better group menu elements.

## Storybook

https://feature/action-menu-trigger-fix--613a8e945a5665003a05113b.chromatic.com/?path=/story/components-actionmenu--keep-open-on-item-click

## Checklist

**Obligatory:**

- [x] Add correct label


packages/react-components/src/components/ActionMenu/ActionMenu.tsx
```diff
@@ -104,28 +102,67 @@ export const ActionMenu: React.FC<ActionMenuProps> = ({
...
-    itemOnClick();
+  const handleItemClick = (itemOnClick?: () => void) => {
+    itemOnClick && itemOnClick();
 
     if (!keepOpenOnClick) {
       setIsVisible(false);
     }
   };
 
+  const getOptionElement = (option: IActionMenuOption, index: number) => {
```
VadymBezpalko, line 113: I'd suggest to put this element as a separate component into a `components` folder

VadymBezpalko (approved):

vladko-uxds (approved):

PR #780, opened 2023-09-14
feat(Tag): Add tooltip support for a Tag component content

VadymBezpalko:
## Description
Adding tooltip support for a Tag component content, when content is overflows the tag container

## Storybook

https://feature-tooltip-in-tag--613a8e945a5665003a05113b.chromatic.com/?path=/story/components-tag--kinds

## Checklist

**Obligatory:**

- [x] Add correct label


vladko-uxds (approved):
👍

packages/react-components/src/components/Tag/Tag.module.scss
```diff
@@ -6,7 +6,6 @@
   border-radius: var(--radius-3);
   border-color: var(--border-basic-primary);
   background-color: var(--surface-tertiary-default);
-  overflow: hidden;
```
marcinsawicki, line 9: This change will break the default behavior, it will break ellipsis in case of small container for tag

marcinsawicki (changes requested):
I think that `Tooltip` should wrap the whole component, not content only.

packages/react-components/src/components/Tag/Tag.module.scss
VadymBezpalko: @marcinsawicki Hi, can you please re-check it? Now tooltips are working differently than before, so maybe it would work properly now?

marcinsawicki (approved):

VadymBezpalko:
One more thing to discuss before merging - how this "tooltip wrapper" impacts the app performance 🤔

Issue #781, opened 2023-09-15
[ActionMenu] - Issue with keyboard control


marcinsawicki:
Using the keyboard control, if we want to move from clickable element to group header, it causes to crash the app.

PR #784, opened 2023-09-15
#781 - [ActionMenu] - Fix for selecting items by keyboard logic

marcinsawicki:
Resolves: #781 

## Description
When the first element on the list was a header element, or disabled element, when using the arrow up to select an element the logic crushes the app, trying to check an element with a negative index number.
This fix will return the last correct index number, if previous elements from the list are blocked (disabled, header)

## Storybook

https://feature-781--613a8e945a5665003a05113b.chromatic.com/?path=/story/components-actionmenu--keep-open-on-item-click

## Checklist

**Obligatory:**

- [x] Add correct label

### DIFF
--- a/packages/react-components/src/components/ActionMenu/ActionMenu.tsx
+++ b/packages/react-components/src/components/ActionMenu/ActionMenu.tsx
@@ -61,17 +61,20 @@ export const ActionMenu: React.FC<ActionMenuProps> = ({
   const [isVisible, setIsVisible] = React.useState(openedOnInit);
   const indexRef = React.useRef(-1);
 
-  const getIndex = (val: number) => {
-    indexRef.current = indexRef.current + val;
+  const getIndex = (val: number): number => {
+    const currentValue = indexRef.current;
+    let newValue = indexRef.current + val;
 
-    while (
-      options[indexRef.current].disabled ||
-      options[indexRef.current].groupHeader
-    ) {
-      indexRef.current = indexRef.current + val;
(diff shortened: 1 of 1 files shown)


packages/react-components/src/components/ActionMenu/ActionMenu.tsx
```diff
@@ -61,17 +61,20 @@ export const ActionMenu: React.FC<ActionMenuProps> = ({
...
-    indexRef.current = indexRef.current + val;
-
-    while (
-      options[indexRef.current].disabled ||
-      options[indexRef.current].groupHeader
-    ) {
-      indexRef.current = indexRef.current + val;
+  const getIndex = (val: number): number => {
+    const currentValue = indexRef.current;
+    let newValue = indexRef.current + val;
```
VadymBezpalko, line 66:
```suggestion
    let newValue = currentValue + val;
```

packages/react-components/src/components/ActionMenu/ActionMenu.tsx
```diff
@@ -61,17 +61,20 @@ export const ActionMenu: React.FC<ActionMenuProps> = ({
...
+
+    while (options[newValue]?.disabled || options[newValue]?.groupHeader) {
+      newValue = newValue + val;
```
VadymBezpalko, line 69:
```suggestion
      newValue += val;
```

VadymBezpalko (approved):

PR #796, opened 2023-09-22
[Avatar] - Fix for moving avatar when rim is visible

marcinsawicki:
## Description
Removed margin when rim is visible.

## Storybook

https://feature/avatar-with-rim-fix--613a8e945a5665003a05113b.chromatic.com/?path=/story/components-avatar--sizes-with-rim

## Checklist

**Obligatory:**

- [x] Add correct label


sgraczyk (approved):
Seems to be working fine in Storybook

Issue #807, opened 2023-10-04
[ActionMenu] - OnClick and Focus states


vladko-uxds:
Update mouse click state - no focus ring on mouse interaction.
Focus ring should appear only for keyboard interaction

See design spec
https://www.figma.com/file/2pFu80PXO5A2tfyrAGnx91/Product-Components?type=design&node-id=18161-4587&mode=design&t=R9wXp7vjFBbainL8-4

PR #812, opened 2023-10-05
#807 - ActionMenu - Group header styles change, improved logic for keyboard control

marcinsawicki:
Resolves: #807 

## Description
Changed the styles for the group header if that element is first on the menu list.
Added required `id` prop for component that will be used in the keyboard control logic. (this fixes the issue If there are multiple instantiations of that component in a single view).
Changed the styles for `focus` state, it will now work only for keyboard control.

## Storybook

https://feature-807--613a8e945a5665003a05113b.chromatic.com/?path=/story/components-actionmenu--keep-open-on-item-click

## Checklist

**Obligatory:**

- [x] Add correct label

### DIFF
--- a/packages/react-components/src/components/ActionMenu/ActionMenu.tsx
+++ b/packages/react-components/src/components/ActionMenu/ActionMenu.tsx
@@ -77,9 +77,10 @@ export const ActionMenu: React.FC<ActionMenuProps> = ({
     indexRef.current = getIndex(val);
     const elements = ref.current?.children;
     const elementToFocus =
-      elements && (elements[indexRef.current].children[0] as HTMLButtonElement);
+      elements &&
+      (elements[indexRef.current]?.children[0] as HTMLButtonElement);
 
-    return elementToFocus && elementToFocus.focus();
+    return elementToFocus?.focus();
   };
 
   const onKeyDown = (e: KeyboardEvent) => {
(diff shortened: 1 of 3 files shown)


packages/react-components/src/components/ActionMenu/ActionMenu.tsx
```diff
@@ -77,17 +73,24 @@ export const ActionMenu: React.FC<ActionMenuProps> = ({
     return newValue;
   };
 
+  const focusElement = (val: number) => {
+    indexRef.current = getIndex(val);
+    const elements = ref.current?.children;
+    const elementToFocus =
+      elements && (elements[indexRef.current].children[0] as HTMLButtonElement);
```
VadymBezpalko, line 80:
```suggestion
      elements && (elements[indexRef.current]?.children[0] as HTMLButtonElement);
```

packages/react-components/src/components/ActionMenu/ActionMenu.tsx
```diff
@@ -77,17 +73,24 @@ export const ActionMenu: React.FC<ActionMenuProps> = ({
...
+
+    return elementToFocus && elementToFocus.focus();
```
VadymBezpalko, line 82:
```suggestion
    return elementToFocus?.focus();
```

VadymBezpalko (approved):

vladko-uxds (approved):
👍

Issue #768, opened 2023-09-08
[BUG] Cannot read properties of undefined (reading 'disabled') in Picker Component


MateuszRomek:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

After adding prop `clearSearchAferSelection` from beta.4 and clicking enter without anything typed in input we receive crash in picker component.a 

### Steps To Reproduce

1. Open picker with search enabled
2. Click enter
3. Error `Cannot read properties of undefined (reading 'disabled')`

### Expected Behavior

Component shouldn't crash 

### Screenshots

https://github.com/livechat/design-system/assets/46934668/71842b2e-1ffc-4307-82a7-56014bd83881

### Design system package(s) version

examples:
  -  @livechat/design-system-react-components": "^1.0.0-beta.4

### Environment

```markdown
Chrome
```

MateuszRomek:
Same error for  "@livechat/design-system-react-components": "1.0.0-beta.12",

PR #818, opened 2023-10-09
#768 - Picker - Fix for TS error when input is empty and no item is selected

marcinsawicki:
Resolves: #768 

## Description
Fix for the issue described in the task.
It was caused by the event listener for key down when the list is opened. By capturing the `enter` key, executed logic causes to throw the error, because it wants to refer to the selected item from the list, where there is any item selected.

## Storybook

https://feature-768--613a8e945a5665003a05113b.chromatic.com/?path=/docs/components-picker--docs

## Checklist

**Obligatory:**

- [x] Add correct label

### DIFF
--- a/packages/react-components/src/components/Picker/PickerList.tsx
+++ b/packages/react-components/src/components/Picker/PickerList.tsx
@@ -75,7 +75,7 @@ export const PickerList: React.FC<IPickerListProps> = ({
 
     if (
       e.key === KeyCodes.enter &&
-      !!items[indexRef.current] &&
+      items[indexRef.current] &&
       !items[indexRef.current].disabled
     ) {
       e.preventDefault();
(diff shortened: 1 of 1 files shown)


packages/react-components/src/components/Picker/PickerList.tsx
```diff
@@ -73,7 +73,11 @@ export const PickerList: React.FC<IPickerListProps> = ({
       setCurrentItemKey(items[indexRef.current].key);
     }
 
-    if (e.key === KeyCodes.enter && !items[indexRef.current].disabled) {
+    if (
+      e.key === KeyCodes.enter &&
+      !!items[indexRef.current] &&
```
VadymBezpalko, line 78: Do you think the double negation here is necessary?

VadymBezpalko (approved):

PR #846, opened 2023-10-19
fix(ActionMenu): Change divider position after element with this flag

marcinsawicki:
## Description
Change styles for divider, now it will be displayed after element with this flag set.

## Storybook

https://feature-action-menu-divider-fix--613a8e945a5665003a05113b.chromatic.com/?path=/story/components-actionmenu--keep-open-on-item-click

## Checklist

**Obligatory:**

- [x] Add correct label


VadymBezpalko (approved):
👍

Issue #876, opened 2023-11-06
[Details card] - Improvements


marcinsawicki:
- Card label should accept custom elements (image, avatar, custom element)
- Divider should be displayed on the bottom
-

vladko-uxds:
Add Icon left to a card title
![image](https://github.com/livechat/design-system/assets/125188605/720873b1-eee7-4ffb-8e57-f511dfd3c87e)

PR #879, opened 2023-11-07
feat(PromoBannerV2): Style changes, added possibility to pass own element to card label

marcinsawicki:
Resolves: #876 

## Storybook

https://feature-876--613a8e945a5665003a05113b.chromatic.com/?path=/story/components-detailscard--example-usage

## Checklist

**Obligatory:**

- [ ] Add correct label

### DIFF
--- a/packages/react-components/src/components/DetailsCard/DetailsCard.module.scss
+++ b/packages/react-components/src/components/DetailsCard/DetailsCard.module.scss
@@ -13,7 +13,7 @@ $base-class: 'details-card';
   }
 
   &--with-divider {
-    border-top: 1px solid var(--border-basic-tertiary);
+    border-bottom: 1px solid var(--border-basic-tertiary);
   }
 
   &__button {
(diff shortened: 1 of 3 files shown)


packages/react-components/src/components/DetailsCard/DetailsCard.module.scss
```diff
@@ -56,6 +51,27 @@ $base-class: 'details-card';
     }
   }
 
+  &__label {
+    display: flex;
+    align-items: center;
+    overflow: hidden;
+
+    &--with-margin {
+      margin-right: 36px;
```
mreszke, line 60: why not using spacing token?

mreszke (approved):

packages/react-components/src/components/DetailsCard/DetailsCard.module.scss
marcinsawicki: This is closely related to the behavior of the elements. In the case of the `hideLabelOnOpen` option, the chevron icon is a floating element that moves with the animation when the tab is opened. The value of this margin is not subject to spacing token because it is a space reserved for the described case, so that the text does not overlap the floating element.

vladko-uxds (approved):
👍

Issue #899, opened 2023-11-23
[ActionMenu, Tooltip, Popover] - add "strategy" param


marcinsawicki:
https://floating-ui.com/docs/usefloating#strategy

We should add `strategy` param to our components. which will enable the possibility to change the behavior of floating elements, because in some cases where a component is nested in a small box with `overflow: hidden`, the tooltip is not rendered correctly.

PR #900, opened 2023-11-23
feat(Tooltip): Added flag to set the position strategy

marcinsawicki:
Resolves: #899 

## Description
This change should give the possibility to solve the issues with rendering floating elements inside containers with `overflow: hidden` style, by using the `floatingStrategy` flag, that can set the position of floating element to `fixed`.
This change includes `Tooltip`, `ActionMenu` and `Popover` components.

## Storybook

https://feature-899--613a8e945a5665003a05113b.chromatic.com/?path=/story/components-tooltip--default

## Checklist

**Obligatory:**

- [x] Add correct label

### DIFF
--- a/packages/react-components/src/components/ActionMenu/ActionMenu.tsx
+++ b/packages/react-components/src/components/ActionMenu/ActionMenu.tsx
@@ -11,6 +11,7 @@ import {
   useDismiss,
   useRole,
   useTransitionStyles,
+  UseFloatingOptions,
 } from '@floating-ui/react';
 import cx from 'clsx';
 
@@ -65,6 +66,10 @@ export interface ActionMenuProps {
    * Optional handler called on menu open
    */
   onOpen?: () => void;
+  /**
+   * Set the type of CSS position property to use
+   */
+  floatingStrategy?: UseFloatingOptions['strategy'];
 }
 
 const baseClass = 'action-menu';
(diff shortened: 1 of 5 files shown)


packages/react-components/src/components/ActionMenu/ActionMenu.tsx
```diff
@@ -81,6 +86,7 @@ export const ActionMenu: React.FC<ActionMenuProps> = ({
   visible,
   onClose,
   onOpen,
+  floatingStrategy = 'absolute',
```
VadymBezpalko, line 89: there is no need to set the fault value here - it already has this value by the library side

packages/react-components/src/components/ActionMenu/ActionMenu.tsx
```diff
@@ -65,6 +66,10 @@ export interface ActionMenuProps {
    * Optional handler called on menu open
    */
   onOpen?: () => void;
+  /**
+   * Set the type of CSS position property to use
```
VadymBezpalko, line 70: I'd suggest adding a link to the floating UI docs here as well

packages/react-components/src/components/ActionMenu/ActionMenu.tsx
```diff
@@ -11,6 +11,7 @@ import {
   useDismiss,
   useRole,
   useTransitionStyles,
+  UseFloatingOptions,
```
VadymBezpalko, line 14: We can directly import `Strategy` type

VadymBezpalko (approved):
Great 💪 
P.S.: I left some minor comments - please apply them to other components as well.

vladko-uxds (approved):

Issue #920, opened 2023-12-12
[Button] - Update button height


vladko-uxds:
Button should have instead of fixed `height`->`min-height` and specific `padding(top-bottom)`

- [ ] Primary
- [ ] Secondary
- [ ] Destructive
- [ ] Float button
- [ ] Plain button
- [ ] HighContrast

Compact size:
`min-height: 32px`
`padding -top/bottom: 6px`

Medium size:
`min-height: 36px`
`padding -top/bottom: var(--spacing-2)`

Large size:
`min-height: 44px`
`padding -top/bottom: var(--spacing-3)`

VadymBezpalko:
Unfortunately, implementing these changes will lead to a breaking change, as altering the `height` parameter to `min-height` could disrupt the visual appearance in scenarios where a custom `height` is used, both within and outside DS components. Setting a `min-height` value could override any custom `height` that is set to be less than the proposed `min-height`.
To avoid causing visual issues in existing implementations, it is better to introduce this improvement in the next `major` version of the library.

Issue #896, opened 2023-11-22
[Button] - styles issue


marcinsawicki:
Issue with the styles for `compact/large` size with icon only:

![Image](https://github.com/livechat/design-system/assets/23378800/1b943b67-c350-4959-85ff-82b047300625)

PR #926, opened 2023-12-15
fix(Button): Fix for Button component height

VadymBezpalko:
Resolves: #920 #896 

## Description
Properly handling button height for elements with border and without it

## Storybook

https://feature-920--613a8e945a5665003a05113b.chromatic.com

## Checklist

**Obligatory:**

- [ ] Add correct label

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -2,6 +2,6 @@
   "packages": [
     "packages/*"
   ],
-  "version": "1.12.0",
+  "version": "1.13.0",
   "$schema": "node_modules/lerna/schemas/lerna-schema.json"
 }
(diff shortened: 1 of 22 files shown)


marcinsawicki (approved):

VadymBezpalko:
Unfortunately, implementing these changes will lead to a breaking change, as altering the `height` parameter to `min-height` could disrupt the visual appearance in scenarios where a custom `height` is used, both within and outside DS components. Setting a `min-height` value could override any custom `height` that is set to be less than the proposed `min-height`.
To avoid causing visual issues in existing implementations, it is better to introduce this improvement in the next `major` version of the library.

Issue #934, opened 2023-12-19
docs(PR request): Enhance PR request template 


VadymBezpalko:
### Description

Improve existing PR templates to automatically fill the task number in the "Resolve: #" section and generate a storybook link based on the PR branch

PR #938, opened 2023-12-19
docs(PR): adding GH Action to modify branch name in the PR description

VadymBezpalko:
Resolves: #934

## Description
Adding a GitHub Action to automatically replace {issue-number} in the Pull Request template with the actual branch name from Git (using the pattern feature/XXX, where there is a check to ensure that XXX is a number).

Unfortunately, this is the only way to obtain the branch name, as there is no native support for this feature in GitHub yet.

## Storybook

https://feature-934--613a8e945a5665003a05113b.chromatic.com

## Checklist

**Obligatory:**

- [ ] Add correct label


.github/workflows/update-pr-description.yml
```diff
@@ -0,0 +1,41 @@
+name: Update PR Description
+on:
+  pull_request:
+    types: [opened]
+jobs:
+  update-description:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Update PR Description
+        uses: actions/github-script@v3
```
sgraczyk, line 10: It seems that the latest version if `7`

sgraczyk (approved):

marcinsawicki (approved):

.github/workflows/update-pr-description.yml
VadymBezpalko: thanks, I will update it than 👍

PR #973, opened 2024-01-16
fix(components): Adding aria-label to close buttons in components

VadymBezpalko:
## Description
Adding aria-label to close buttons in components

## Storybook

https://feature-close-buttons-aria--613a8e945a5665003a05113b.chromatic.com

## Checklist

**Obligatory:**

- [x] Add correct label

### DIFF
--- a/packages/react-components/src/components/ActionMenu/ActionMenu.tsx
+++ b/packages/react-components/src/components/ActionMenu/ActionMenu.tsx
@@ -77,6 +77,10 @@ export interface ActionMenuProps {
    * https://floating-ui.com/docs/usefloating#strategy
    */
   floatingStrategy?: Strategy;
+  /**
+   * Test id passed to the wrapper element
+   */
+  ['data-testid']?: string;
 }
 
 const baseClass = 'action-menu';
@@ -95,6 +99,7 @@ export const ActionMenu: React.FC<ActionMenuProps> = ({
   onOpen,
   floatingStrategy,
   selectedOptions,
+  ['data-testid']: testId = 'action-menu-trigger-button',
   ...props
 }) => {
   const isControlled = visible !== undefined;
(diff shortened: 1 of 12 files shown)


MateuszRomek (approved):

packages/react-components/src/components/ActionMenu/ActionMenu.tsx
```diff
@@ -77,6 +77,10 @@ export interface ActionMenuProps {
    * https://floating-ui.com/docs/usefloating#strategy
    */
   floatingStrategy?: Strategy;
+  /**
+   * Test id passed to the wrapper element
+   */
+  ['data-testid']?: string;
```
sgraczyk, line 83: @VadymBezpalko Have you considered an extendable Props interface for all components, so some core fields would be part of every single component? `['data-testid']` could be a part of them. 

Such approach would prevent you from necessity to add them from time to time when needed.

sgraczyk (approved):

packages/react-components/src/components/ActionMenu/ActionMenu.tsx
```diff
@@ -251,7 +256,8 @@ export const ActionMenu: React.FC<ActionMenuProps> = ({
   return (
     <>
       <div
-        data-testid="action-menu-trigger-button"
```
MrVorgoth, line 254: If it was passed before, shouldn't it stay and be overwritten by custom data-test id if passed?

packages/react-components/src/components/ActionMenu/ActionMenu.tsx
VadymBezpalko: Great idea 👍

MrVorgoth (approved):

vladko-uxds (approved):

packages/react-components/src/components/ActionMenu/ActionMenu.tsx
MateuszRomek: Its set as default value in props

Issue #992, opened 2024-01-30
[BUG] Allow control over Picker's state


ashbork:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

The `selected` prop doesn't control `Picker`'s state completely. For example, setting `selected` to `null` should intuitively:

- display the placeholder
- prevent any of the list items from displaying as 'ticked'

However, while the placeholder actually is displayed correctly (and the clicked item doesn't have its `selectedItemBody` placed in the Trigger), any items that get selected (but not added to the `selected` prop) gain the styles associated with selected options.

### Steps To Reproduce

1. Set the `selected` prop on a `Picker` to `null` or `[]`
2. Open the `Picker` and select an option
3. Open the `Picker` again - the option will have a tick next to it despite not being in the `selected` prop

### Expected Behavior

The component should probably be completely controlled if given the `selected` prop (or at least disable any selection styles when `selected` is set to `null`?)

### Screenshots

![Monosnap screencast 2024-01-30 13-35-27](https://github.com/livechat/design-system/assets/82605611/2b1943fd-d3b6-4760-9be9-9ca53e708bb2)

### Design system package(s) version

@livechat/design-system-react-components@1.2.0

PR #995, opened 2024-02-01
fix(Picker): fix for data controlled mode

VadymBezpalko:
Resolves: #992

## Description

Adding data-controlled mode: when there is `selected` prop passed to the Picker component it will track only those indexes as selected and would ignore internal state.

## Storybook

https://feature-992--613a8e945a5665003a05113b.chromatic.com

## Checklist

**Obligatory:**

- [ ] Add correct label


vladko-uxds (approved):
👍

ashbork (approved):

PR #997, opened 2024-02-02
fix(Modal): Possibility to scroll down the overlay if modal is larger than window

marcinsawicki:
## Description
Implementation of `FloatingOverlay` component from `floating-ui` as our modals overlay. It will allow to scroll down the whole overlay if the modal size is larger than user window.

## Storybook
https://feature-modal-overflow-scroll--613a8e945a5665003a05113b.chromatic.com/?path=/story/components-modal--modal

## Checklist

**Obligatory:**

- [x] Add correct label

### DIFF
--- a/packages/react-components/src/components/Modal/components/ModalBase.tsx
+++ b/packages/react-components/src/components/Modal/components/ModalBase.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 
+import { FloatingOverlay } from '@floating-ui/react';
 import cx from 'clsx';
 
 import { KeyCodes } from '../../../utils/keyCodes';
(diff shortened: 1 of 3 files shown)


packages/react-components/src/components/Modal/components/ModalBase.tsx
```diff
@@ -69,7 +70,7 @@ export const ModalBase: React.FC<React.PropsWithChildren<ModalBaseProps>> = ({
   };
 
   return (
-    <div
+    <FloatingOverlay
```
VadymBezpalko, line 73: Have you checked how this works in AA? Does it clip with navbar or no?

VadymBezpalko (approved):

packages/react-components/src/components/Modal/components/ModalBase.tsx
marcinsawicki: Yep, added `z-index` on the overflow wrapper to prevent elements overflow :)

Issue #996, opened 2024-02-02
[Build] Tree-shakable components


VadymBezpalko:
### Description

Change the library build process to enable the tree-shanking feature and copy some handful static files like README.md

PR #1000, opened 2024-02-12
build(react-components,icons): Tree-shaking for react-components; icons package build improvements

VadymBezpalko:
Resolves: #996

## Description

- This PR contains an optimized way of building a component library with a tree-shaking feature available. The build source is now contains separated files for each component, separated js and ts declaration files and separated css file in the assets folder. Now style.css file contains only shared css variables (color, radius, spacing etc tokens). Also, we get rid of garbage typings for spec and storybook files mistakenly added to the build source previously. 
- Using [globbing](https://nodejs.org/api/packages.html#package-entry-points) for icons exports in `icons` package instead of script
- Simplifying Icons build process by removing custom scripts and replacing them with globbing name patterns.

The example of DS lib usage in a clean CRA app (consuming only the Button component):
Before the changes
cra_default_ds

After the changes
cra_improved_ds

Example of the AA bundle analysis with unchaged DS vs Tree-shakable DS libs:
Before:
AA_bundle_default

After:
AA_bundle_optimized

Worth mentioning:
When comparing the current DS lib installed vs the tree-shakable version we have a minor AA build size increase, despite `vendors` chunk decreasing in size. It is illustrated in the image below.

AA_comparation

## Storybook

https://feature-996--613a8e945a5665003a05113b.chromatic.com

## Checklist

**Obligatory:**

- [x] Add correct label


vladko-uxds (approved):
🔥

mreszke (approved):
💪

marcinsawicki (approved):

Issue #1020, opened 2024-02-27
[Input field] - Update disabled state


vladko-uxds:
For `Disabled` state update text color to `ContentBasicDisabled`

![Image](https://github.com/livechat/design-system/assets/125188605/7a459491-5d56-4a1b-8a6f-1024fcb6b710)

PR #1021, opened 2024-02-27
fix(Input): change color for disabled state if value is given

marcinsawicki:
Resolves: #1020

## Description
Bug fix - correct text color for disabled state if any value is given to the input

## Storybook

https://feature-1020--613a8e945a5665003a05113b.chromatic.com

## Checklist

**Obligatory:**

- [x] Add correct label


vladko-uxds (approved):
👍

PR #1035, opened 2024-03-06
fix(ActionBar): New menu placement logic, styles fixes, docs and stories update

marcinsawicki:
## Description
Menu placement will be calculated on the total number of visible elements and positioned by`absolute`. In cases where the menu is visible, the container will add extra space to display it.

## Storybook

https://feautre-action-bar-upgrade--613a8e945a5665003a05113b.chromatic.com/?path=/docs/experimental-actionbar--docs

## Checklist

**Obligatory:**

- [x] Add correct label


vladko-uxds (approved):
👍

packages/react-components/src/components/ActionBar/ActionBar.mdx
```diff
@@ -0,0 +1,51 @@
...
+  ]} 
+/>
+```
+
+#### Types <a id="Types" />
+
+By default, the component is set to `menu` mode, in which hidden items will be moved to the menu. <br />
+This behavior can be changed using the `type` flag set to `scroll`, then elements will not be hidden and the container will be scrollable.
+
+<Canvas of={ActionBarStories.Horizontal} sourceState="none" />
```
VadymBezpalko, line 47: You can also show the Vertical mode as you already have the story for that

VadymBezpalko (approved):

Issue #1034, opened 2024-03-06
[Colors] - Add New & updates


vladko-uxds:
We need to have new group of color tokens with opacity

**SurfaceOpacity**
- Basic
- Invert

- [ ] Tokens for SurfaceOpacity**Basic**
Rename token:
 `--surface-primary-hover-opacity ` - to new name - `--surface-opacity-basic-hover` Light `rgba(19, 19, 23, 0.07)` | Dark `rgba(255, 255, 255, 0.08)`
Add new token:
`--surface-opacity-basic-active`Light `rgba(19, 19, 23, 0.09)` | Dark `rgba(255, 255, 255, 0.12)`

- [ ] Tokens for SurfaceOpacity**Invert**
**Add new tokens:**
`--surface-opacity-invert-hover` Light `rgba(255, 255, 255, 0.1)` | Dark `rgba(255, 255, 255, 0.1)`
`--surface-opacity-invert-active` Light `rgba(255, 255, 255, 0.15)` | Dark `rgba(255, 255, 255, 0.12)`

VadymBezpalko:
I will not rename the `surface-primary-hover-opacity ` to not cause the conflicts, I'll mark it as deprecated instead and will create a new entry with new values.

PR #1038, opened 2024-03-07
fix(DesignToken): Adding new tokens; adding deprecated info to TS enum

VadymBezpalko:
Resolves: #1034

## Description

- Mark `--surface-primary-hover-opacity ` as deprecated
- Added 4 new tokens with new color group (implies new group in colours sb page)
- Added mechanism to mark deprecated colour in TS enum as well, so it will be visible in IDE

## Storybook

https://feature-1034--613a8e945a5665003a05113b.chromatic.com

## Checklist

**Obligatory:**

- [x] Add correct label


vladko-uxds (approved):
👍

Issue #1037, opened 2024-03-07
[Search] - incorrect content area width


vladko-uxds:
https://design.livechat.com/?path=/docs/components-search--docs

https://github.com/livechat/design-system/assets/125188605/b521422d-b026-4d80-b5d2-302abf234b8c

PR #1039, opened 2024-03-07
fix(Search): fix for short input container, added ellipsis

marcinsawicki:
Resolves: #1037

## Description
Fix for the short input container after typo changes.
Added ellipsis for the input for long string, enabled by default.

## Storybook

https://feature-1037--613a8e945a5665003a05113b.chromatic.com

## Checklist

**Obligatory:**

- [x] Add correct label

### DIFF
--- a/packages/react-components/src/components/Search/Search.tsx
+++ b/packages/react-components/src/components/Search/Search.tsx
@@ -47,6 +47,10 @@ export interface ISearchInputProps {
    * The event handler for onChange
    */
   onChange: (value: string) => void;
+  /**
+   * Set to enable ellipsis
+   */
+  cropOnBlur?: boolean;
 }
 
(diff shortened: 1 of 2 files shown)


vladko-uxds (approved):
👍

packages/react-components/src/components/Search/Search.tsx
```diff
@@ -128,11 +133,16 @@ export const SearchInput: React.FC<ISearchInputProps> = ({
         disabled={isDisabled}
         kind="primary"
       />
-      <Text as="div">
+      <Text
+        as="div"
+        className={cx(
+          styles[inputBaseClass],
```
VadymBezpalko, line 139: I think we need to rename this class, because it no longer addresses the input element

VadymBezpalko (approved):

Issue #1058, opened 2024-03-22
[Avatar] Component gets compressed, `minWidth` issue


kubaczerwinski77:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

When we use `Avatar` component next to element that has more width than it should and overflow the parent component, the `Avatar` gets compressed and looks a bit strange.

Slack: https://text.slack.com/archives/CA945NPQR/p1711005341172669

### Steps To Reproduce

1. Create parent container with width 100px, make it flex and display children in a row
2. Add `Avatar` and some text that overflows the width of parent component
3. The `Avatar`should be compressed like on the screenshot

### Expected Behavior

It should always be a square or circle with perfect aspect ratio.

### Screenshots

![Image](https://github.com/livechat/design-system/assets/86623851/1d93c74f-df48-402c-8675-de6784d15edc)

### Design system package(s) version

examples:
  - @livechat/design-system-react-components@1.19.2

### Environment

Macbook
MacOS
Chrome
Version 122.0.6261.129 (Official Build) (arm64)

PR #1061, opened 2024-03-26
fix(Avatar): minHeight and minWidth for avatar

VadymBezpalko:
Resolves: #1058

## Description
- Adding minHeight and minWidth for avatar sizes to prevent shrinking
- Refactoring styles using mixins

## Storybook

https://feature-1058--613a8e945a5665003a05113b.chromatic.com

## Checklist

**Obligatory:**

- [x] Add correct label


marcinsawicki (approved):
Only need to add `flex-shrink: 0` to prevent shrinking in `flex` container with another elements.

Issue #1082, opened 2024-04-11
[BUG] Picker sometimes is losing a foucs in search field


sirekpiotr:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

We've spotted and reported by HelpDesk clients that the picker we use for tag picker is sometimes using its focus on the search field when users use only a keyboard to navigate and use it. 

The issue is also visible in our Storybook.

https://github.com/livechat/design-system/assets/32463186/c0c0f3ad-9ea1-4ab1-8495-0fef749580c0
https://github.com/livechat/design-system/assets/32463186/ec824497-e452-46b9-8a8d-deef8f587e56

### Steps To Reproduce

1. Try to search for something in the picker
2. Use keyboard to move around the picker options 
3. Try to remove the phrase in the search field
4. Sometimes you will see that search field is losing focus

### Expected Behavior

Search field should always have focus enabled and react for events from the keyboard like backspace.

### Screenshots

See the description for two videos attached. 

### Design system package(s) version

examples:
  - @livechat/design-system-react-components@1.0.0-beta.11

kajaHD:
Hi, I've experienced similar issue with multipicker - after searching sth and navigating to the elements using arrows, the focus from search is gone and I am not able to edit search content

https://github.com/livechat/design-system/assets/49684667/511a3b5d-d36b-4632-a8a7-d0f189fa724b

PR #1107, opened 2024-05-16
fix(Picker): fix for search input focusing + controls improve

marcinsawicki:
Resolves: #1082

## Description
Removed list focusing on keyboard control.

## Storybook

https://feature-1082--613a8e945a5665003a05113b.chromatic.com

## Checklist

**Obligatory:**

- [x] Add correct label


packages/react-components/src/components/Picker/components/PickerTriggerBody.tsx
```diff
@@ -56,6 +60,30 @@ export const PickerTriggerBody: React.FC<ITriggerBodyProps> = ({
...
+      e.preventDefault();
+      onClear();
+    }
+  };
+
+  React.useEffect(() => {
+    if (!isOpen && isTriggerFocused) {
+      document.addEventListener('keydown', onKeyDown);
+
+      return () => document.removeEventListener('keydown', onKeyDown);
```
VadymBezpalko, line 83: I'm not sure if this hook will always remove added listeners - when a user opens the Picker then the `isOpen` prop will change its value and the removeEventListener will be inaccessible

VadymBezpalko (approved):

Issue #1088, opened 2024-04-18
[Picker] Broken ellipsis after adding typography


marcinsawicki:
After adding Typography to Picker, the ellipsis doesn't work if the component is small and don't use custom elements as picker list items.
Adding `max-width: 100%` to typography wrapper should handle this.

![Image](https://github.com/livechat/design-system/assets/23378800/405618dd-5c82-45fc-b96d-ada121163731)

PR #1108, opened 2024-05-17
fix(Picker): ellipsis for list items and selected item

marcinsawicki:
Resolves: #1088

## Storybook

https://feature-1088--613a8e945a5665003a05113b.chromatic.com

## Checklist

**Obligatory:**

- [x] Add correct label

### DIFF
--- a/packages/react-components/src/components/Picker/components/PickerListItem.tsx
+++ b/packages/react-components/src/components/Picker/components/PickerListItem.tsx
@@ -68,14 +68,15 @@ export const PickerListItem: React.FC<IPickerListItemProps> = ({
           />
         )}
         <div className={styles[`${itemClassName}__label-container`]}>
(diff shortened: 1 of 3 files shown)


packages/react-components/src/components/Picker/components/PickerListItem.tsx
```diff
@@ -138,9 +139,7 @@ export const PickerListItem: React.FC<IPickerListItemProps> = ({
             item.key === SELECT_ALL_OPTION_KEY,
         })}
       >
-        <Text as="span" className={styles[`${itemClassName}__content`]}>
```
VadymBezpalko, line 141: What about the `_content` class? Shouldn't we remove it as well?

VadymBezpalko (approved):

PR #1128, opened 2024-06-03
fix(DatePicker): prevent infinite update

marcinsawicki:
Resolves: #{issue-number}

## Description
This fix should prevent to infinite loop update of the component, on `month` prop change when `toMonth` is set.

## Storybook

https://feature-date-picker-update-loop-fix--613a8e945a5665003a05113b.chromatic.com

## Checklist

**Obligatory:**

- [x] Add correct label


VadymBezpalko (approved):

vladko-uxds (commented):
👍

Issue #669, opened 2023-07-20
Unit tests audit


marcinsawicki:
TODO:

- [x] Audit the currently implemented unit tests. The result of which should be a written short component testing policy (maybe a checklist of what such a test should cover)

When the first point is done:
- [x] Add missing tests, and removal of redundant ones

```[tasklist]
### Tasks
- [x] [ActionBar] - unit test check/update/rewrite
- [x] [ActionMenu] - unit test check/update/rewrite
- [x] [Alert] - unit test check/update/rewrite
- [x] [Avatar] - unit test check/update/rewrite
- [x] [Badge] - unit test check/update/rewrite
- [x] [Button] - unit test check/update/rewrite
- [x] [Card] - unit test check/update/rewrite
- [x] [Checkbox] - unit test check/update/rewrite
- [x] [DatePicker] - unit test check/update/rewrite
- [x] [DetailsCard] - unit test check/update/rewrite
- [x] [FieldGroup] - unit test check/update/rewrite
- [x] [FileUploadProgress] - unit test check/update/rewrite
- [x] [Form] - unit test check/update/rewrite
- [x] [FormField] - unit test check/update/rewrite
- [x] [FormGroup] - unit test check/update/rewrite
- [x] [Icon] - unit test check/update/rewrite
- [x] [Input] - unit test check/update/rewrite
- [x] [Link] - unit test check/update/rewrite
- [x] [Loader] - unit test check/update/rewrite
- [x] [Modal] - unit test check/update/rewrite
- [x] [NumericInput] - unit test check/update/rewrite
- [ ] [Picker] - unit test check/update/rewrite
- [x] [Popover] - unit test check/update/rewrite
- [x] [Progress] - unit test check/update/rewrite
- [x] [PromoBannerV2] - unit test check/update/rewrite
- [x] [RadioButton] - unit test check/update/rewrite
- [x] [Search] - unit test check/update/rewrite
- [x] [SegmentedControl] - unit test check/update/rewrite
- [x] [Skeleton] - unit test check/update/rewrite
- [x] [Switch] - unit test check/update/rewrite
- [x] [Tab] - unit test check/update/rewrite
- [x] [TagInput] - unit test check/update/rewrite
- [x] [Textarea] - unit test check/update/rewrite
- [x] [Toast] - unit test check/update/rewrite
- [x] [Tooltip] - unit test check/update/rewrite
- [x] [Typography] - unit test check/update/rewrite
[...]

PR #1154, opened 2024-06-10
fix(Card): tests updates

marcinsawicki:
Resolves: #669

## Storybook

## Checklist

**Obligatory:**

- [x] Add correct label


packages/react-components/src/components/Card/Card.spec.tsx
```diff
@@ -1,97 +1,122 @@
...
+    const { container } = renderComponent({ className: 'my-css-class' });
+
     expect(container.firstChild).toHaveClass('my-css-class');
   });
 
-  it('should display only title if image is not provided', () => {
-    const { getByText, getByRole } = render(<Card title="Hello" />);
-    expect(getByText('Hello')).toBeVisible();
-    expect(() => getByRole('img')).toThrow();
+  it('should display title if is provided', () => {
```
VadymBezpalko, line 22:
```suggestion
  it('should display title if it's provided', () => {
```

VadymBezpalko (approved):

PR #1162, opened 2024-06-11
fix(Textarea): tests updates

marcinsawicki:
Resolves: #669

## Storybook

## Checklist

**Obligatory:**

- [x] Add correct label


packages/react-components/src/components/Textarea/Textarea.spec.tsx
```diff
@@ -2,59 +2,69 @@ import * as React from 'react';
...
-
+    userEvent.click(getByRole('textbox'));
     expect(customOnFocus).toHaveBeenCalled();
 
-    textarea?.blur();
+    userEvent.click(document.body);
     expect(customOnBlur).toHaveBeenCalled();
   });
+
+  it('should call onChange if textarea value change', () => {
```
VadymBezpalko, line 49: I'm not sure we need this test, because onChange callback is handled by a native implementation

VadymBezpalko (approved):

PR #1168, opened 2024-06-12
fix(ActionMenu): fix for the menu height

marcinsawicki:
Resolves: #{issue-number}

## Description
Added `size` from `floating-ui` to define the floating element max height, and some style changes to make it possible to scroll the menu items.

## Storybook

https://feature-action-menu-size-fix--613a8e945a5665003a05113b.chromatic.com

## Checklist

**Obligatory:**

- [x] Add correct label


packages/react-components/src/components/ActionMenu/constants.tsx
```diff
@@ -57,4 +57,9 @@ export const exampleOptions = [
     ),
     onClick: noop,
   },
+  ...[...Array(10)].map((_, index) => ({
+    key: `option${index + 8}`,
+    element: <ActionMenuItem>{`Menu item #${index + 8}`}</ActionMenuItem>,
+    onClick: noop,
+  })),
```
VadymBezpalko, line 64: We need to separate those storybook options from our constants.ts file, because after the build the whole file ends up in output files, and we do not wont to have storybook examples in the final build

vladko-uxds (approved):
🔥

VadymBezpalko (approved):

Issue #1003, opened 2024-02-13
[Theme Provider] - Implement Theme Provider


VadymBezpalko:
### Description

- Implement Theme provider functionality for better theme management out-of-the-box in DS.
- Add an option to customize themes (dark/light).
- Create new documentation for theme usage and theme switching.

PR #1176, opened 2024-06-14
feat(ThemeProvider): adding ThemeProvider functionality and removing Legacy theme

VadymBezpalko:
Resolves: #1003

## Storybook

https://feature-1003--613a8e945a5665003a05113b.chromatic.com

## Checklist

**Obligatory:**

- [ ] Add correct label


marcinsawicki (approved):

packages/react-components/src/providers/ThemeProvider.tsx
```diff
@@ -0,0 +1,70 @@
...
+    }
+  }, [theme, variables]);
+
+  const toggleTheme = () => {
+    setTheme((prevTheme) => (prevTheme === 'light' ? 'dark' : 'light'));
+  };
+
+  const setCustomVariables = (newVariables: CustomVariables) => {
+    setVariables(newVariables);
+  };
```
mreszke, line 63: shouldn't we use `useCallbacks` to prevent re-rendering with unstable functions refs?

packages/react-components/src/providers/ThemeProvider.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+  return (
+    <ThemeContext.Provider value={{ theme, toggleTheme, setCustomVariables }}>
```
mreszke, line 66: Also here, the `value` should be moved to const and memoized

mreszke (approved):

PR #1184, opened 2024-06-18
fix(Tab): tests updates

marcinsawicki:
Resolves: #669

## Storybook

## Checklist

**Obligatory:**

- [ ] Add correct label


packages/react-components/src/components/Tab/Tab.tsx
```diff
@@ -56,6 +56,8 @@ export const Tab: React.FC<React.PropsWithChildren<TabProps>> = ({
     <Text
       {...restProps}
       as={restProps.href ? 'a' : 'button'}
+      role="tab"
+      aria-selected={isSelected ? true : false}
```
VadymBezpalko, line 60:
```suggestion
      aria-selected={isSelected}
```

VadymBezpalko (approved):

PR #1191, opened 2024-06-19
feat(Interactive): allow all Button props in tooltip button, make secondButton optional

### LINKED ISSUE #1189: [Tooltip / Interactive] Make the component props less restrictive
### Description

`Interactive` currently **forces** its consumers to define both a primary and a secondary button. As an user I may want to have only one button. As well as that, the buttons aren't as customisable as may be necessary for some use-cases: I'd recommend allowing **all** `Button` props in `primary-` and `secondaryButton`

### Design

_No response_

ashbork:
Resolves: #1189 

## Description

This improvement allows more flexibility when using the component:
- The secondary button is now optional
- You can provide all of the props that `Button` has.

For example: before, you couldn't provide an Icon to complement the button - now you can.
Note: existing `Interactive` usage should **not** break or change in any way - this is an API extension and isn't supposed to break current uses of the component

## Storybook

https://feature-1189-tooltip-interactive--613a8e945a5665003a05113b.chromatic.com

## Checklist

**Obligatory:**

- [x] Add correct label

### DIFF
--- a/packages/react-components/src/components/Tooltip/components/Interactive.tsx
+++ b/packages/react-components/src/components/Tooltip/components/Interactive.tsx
@@ -80,18 +80,26 @@ export const Interactive: React.FC<ITooltipInteractiveProps> = ({
         )}
       >
         <Button
+          {...primaryButton}
           kind={primaryButton.kind || getDefaultPrimaryButtonKind(theme)}
           onClick={primaryButton.handleClick}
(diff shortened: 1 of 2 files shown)


VadymBezpalko (approved):
👍

marcinsawicki (approved):

PR #1195, opened 2024-06-20
fix(Picker): tests updates

marcinsawicki:
Resolves: #669

## Storybook

## Checklist

**Obligatory:**

- [ ] Add correct label

### DIFF
--- a/packages/react-components/src/components/Picker/components/PickerTrigger.spec.tsx
+++ b/packages/react-components/src/components/Picker/components/PickerTrigger.spec.tsx
@@ -6,10 +6,6 @@ import noop from '../../../utils/noop';
 
 import { PickerTriggerProps, PickerTrigger } from './PickerTrigger';
 
-import styles from './PickerTrigger.module.scss';
-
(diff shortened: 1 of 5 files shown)


VadymBezpalko (commented):
Great job! 🔥 Looks great 👍 
Btw, what do you think of adding some more tests for keyboard interaction here, like: opening list with keyboard, selecting one/few items and removing them?

packages/react-components/src/components/Picker/components/PickerTrigger.spec.tsx
```diff
@@ -71,7 +51,7 @@ describe('<PickerTrigger> component', () => {
       onClear,
     });
 
-    fireEvent.click(getByTestId(`${baseClass}__clear-icon`));
+    fireEvent.click(getByTestId(`picker-trigger__clear-icon`));
```
VadymBezpalko, line 54: IMO, we can switch to userEvent entirely, it is safer to use and more similar to normal user interactions.

VadymBezpalko (approved):

Issue #1214, opened 2024-06-27
[ActionBar] - bug with last icon


vladko-uxds:
Chevron for expanding  appears over the last App icon, and block the click on integration app

![Image](https://github.com/livechat/design-system/assets/125188605/a841281a-78f9-4b64-9fa7-1015041b3b65)

![Image](https://github.com/livechat/design-system/assets/125188605/856dc8b2-94e8-4d38-bb38-da1551356379)

PR #1219, opened 2024-06-28
feat(ActionBar): logic refactor

marcinsawicki:
Resolves: #1214 

## Description
Logic refactor:
- changed `IntersectionObserver` to `ResizeObserver` - from now logic will observe the container size changes, then will calculate how many option buttons will fit in the currently available space:
   - nothing is changing with buttons possible to show in the container
   - buttons for which there is not enough space will not be rendered (that was necessary only for `IntersectionObserver`) and will be available in the menu
- menu will be placed relative to the other elements (no more `absolute` positioning)

## Storybook

https://feature-action-bar-menu-icon-fix--613a8e945a5665003a05113b.chromatic.com

## Checklist

**Obligatory:**

- [x] Add correct label


vladko-uxds (approved):
👍

packages/react-components/src/components/ActionBar/ActionBar.tsx
```diff
@@ -24,128 +24,83 @@ export const ActionBar: React.FC<IActionBarProps> = ({
...
     styles[baseClass],
     className,
     vertical && styles[`${baseClass}--vertical`]
   );
-  const observerOptions = {
-    root: document.querySelector(`${id}`),
-    threshold: 1,
-  };
-  const shouldDisplayMenu = !isScrollType && menuItemsKeys.length !== 0;
+  const singleElementSize = 44;
```
VadymBezpalko, line 38: Shouldn't it be available as a prop? If not I'd suggest to move it outside the component's function.

packages/react-components/src/components/ActionBar/ActionBar.tsx
```diff
@@ -24,128 +24,83 @@ export const ActionBar: React.FC<IActionBarProps> = ({
...
+    const hasIOSupport = !!window.ResizeObserver;
 
     if (!isScrollType && hasIOSupport) {
-      const target = document.querySelectorAll(
-        `button[data-actionbarid='${id}']`
-      );
+      const observer = new ResizeObserver((entries) => {
+        const { width, height } = entries[0].contentRect;
+        const containerSize = vertical ? height : width;
+        const exstraSpacing = menuOptions.length > 0 ? 60 : 0;
```
VadymBezpalko, line 60: `extraSpacing`

packages/react-components/src/components/ActionBar/ActionBar.tsx
```diff
@@ -24,128 +24,83 @@ export const ActionBar: React.FC<IActionBarProps> = ({
...
-  }, [menuItemsKeys, isScrollType]);
-
-  const getMenuItems = (keys: string[]) => {
-    const filteredOptions = options.filter((row) =>
-      keys.find((k) => k === row.key && !row.hideInMenu)
-    );
+  }, [menuOptions, isScrollType, visibleItemsCount]);
 
-    return filteredOptions.map((o) => {
+  const getMenuItems = (menuOptions: IActionBarOption[]) =>
```
VadymBezpalko, line 80: useCallback?

VadymBezpalko (approved):
👍 
P.S.: Have you checked the component with React profiler to see how it renders?

packages/react-components/src/components/ActionBar/ActionBar.tsx
marcinsawicki: Currently not, but in the future I plan to improve that component and give a possibility to set the buttons size.

Issue #1204, opened 2024-06-25
[Navigation] - new component


vladko-uxds:
Add `Main nav` with  `TopBar` and `SubNavigation` as a component
- [Navbar spec](https://www.figma.com/design/2pFu80PXO5A2tfyrAGnx91/Product-Components?node-id=17377-100533&t=cfzBwG1Dmx1J7Viw-4)

- [TopBar](https://www.figma.com/design/2pFu80PXO5A2tfyrAGnx91/Product-Components?node-id=22472-9762&t=cfzBwG1Dmx1J7Viw-4)

SubNavigation https://www.figma.com/design/2pFu80PXO5A2tfyrAGnx91/Product-Components?node-id=22260-41686&t=cfzBwG1Dmx1J7Viw-4

PR #1227, opened 2024-07-04
feat(AppFrame): new component

marcinsawicki:
Resolves: #1204 

## Description
New `AppFrame` component.
[See docs](https://feature-nav-bar-component--613a8e945a5665003a05113b.chromatic.com/?path=/docs/components-appframe--docs) for more details.

## Storybook

https://feature-nav-bar-component--613a8e945a5665003a05113b.chromatic.com/?path=/story/components-appframe--default

## Checklist

**Obligatory:**

- [x] Add correct label


ashbork:
2 accessibility issues - `li`s aren't immediate children of the list (due to the wrapper div which is caused by the tooltip) and `ul` isn't a valid `nav` element - let's wrap it with a `nav` instead, [much like the example](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/navigation_role#examples) in MDN docs

packages/react-components/src/components/AppFrame/AppFrame.stories.tsx
```diff
@@ -0,0 +1,223 @@
...
+      case 'archives':
+        return getArchivesSubMenu(activeSubItem, setActiveSubItem);
+      default:
+        return null;
+    }
+  };
+
+  return (
+    <AppFrameProvider>
+      <AppFrame
```
VadymBezpalko, line 52: I'd suggest using the AppFrameProvider in AppFrame internally, but to leave it exported externally as well -> thanks to that, developers wouldn't need to declare it explicitly but will have the ability to use the sidenav controls if needed separately

packages/react-components/src/components/AppFrame/AppFrame.stories.tsx
```diff
@@ -0,0 +1,223 @@
...
+        topNavBarNode={
```
VadymBezpalko, line 53: I'd suggest renaming it to `navigation` for simplicity.
And to consider following components structure:
```
<AppFrame
  navigation={
    <Navigation>
      <TopNavigation>
        <NavigationItem {...props} />
        ...
      </TopNavigation>
      <BottomNavigation>
        <NavigationItem {...props} />
        ...
      </BottomNavigation>
    </Navigation>
  }
```

`<Navigation>` - our component responsible for styling and placing the items
`<TopNavigation>` - our component responsible for navigation items placing at the top
`<BottomNavigation>` - our component responsible for navigation items placing at the bottom
`<NavigationItem>` - our nav item component

packages/react-components/src/components/AppFrame/AppFrame.stories.tsx
```diff
@@ -0,0 +1,223 @@
...
+              label="Home"
+              disableTooltip
+              icon={<Icon source={Icons.LiveChatMono} />}
+              onClick={(e) => {
+                e.preventDefault();
+                setActiveItem('home');
+              }}
+              isActive={activeItem === 'home'}
+              href="#"
+            />
```
VadymBezpalko, line 65: Regarding `NavigationItem` component I'd suggest a following changes:

- add the `id` prop.
- use the `id` as a key prop
- return this `id` value as a second parameter in the onClick callback, so it should help to simplify building callbacks on the dev side:
```
onClick={(e, id) => {
                e.preventDefault();
                setActiveItem(id);
              }}
```

packages/react-components/src/components/AppFrame/AppFrame.stories.tsx
```diff
@@ -0,0 +1,223 @@
...
+                >
+                  Custom element with tooltip
+                </Tooltip>
+              }
+              onClick={(e) => e.preventDefault()}
+              href="#"
+            />
+          </NavBarList>
+        }
+        subNavBar={getSubNav()}
```
VadymBezpalko, line 198: I'd suggest renaming it to just - `sideNavigation={<SideNavigation />}`

packages/react-components/src/components/AppFrame/AppFrame.stories.tsx
```diff
@@ -0,0 +1,223 @@
...
+        topBarNode={
```
VadymBezpalko, line 199: we can just name it - `topBar`, we do not need to add a "Node" suffix, it would be clear from TS typing declaration.

packages/react-components/src/components/AppFrame/stories-helpers.tsx
```diff
@@ -0,0 +1,240 @@
...
+      </>
+    </div>
+  );
+};
+
+export const getChatsMenu = (
+  activeSubItem: number,
+  handler: (o: number) => void
+) => (
+  <SubNavBar noGaps title="Simple list">
```
VadymBezpalko, line 60: I'd suggest naming it `SideNavigation` and `SideNavigationItem` as these naming patterns are already used in the AA application.

packages/react-components/src/components/AppFrame/stories-helpers.tsx
```diff
@@ -0,0 +1,240 @@
...
+    <SubNavBarListItem
+      label="Option 1"
+      shouldKeepIconSpace={false}
+      isActive={activeSubItem === 0}
+      onClick={() => handler(0)}
+    />
```
VadymBezpalko, line 66: - We can add here the `id` prop and use it in the onClick handler and the key prop in the same way as in NavigationItem component
- `label` prop should accept ReactNode values
- maybe we can rename `icon` prop to the `leftNode` as it already accepts not only icons but reactnode values

packages/react-components/src/components/AppFrame/stories-helpers.tsx
```diff
@@ -0,0 +1,240 @@
...
+  </SubNavBar>
+);
+
+export const getArchivesSubMenu = (
+  activeSubItem: number,
+  handler: (o: number) => void
+) => (
+  <SubNavBar noGaps title="Options with collapsable list">
+    <SubNavBarList isCollapsible label="Simple">
```
VadymBezpalko, line 122: For better readability, we can rename it to smth like `SideNavigationGroup`

packages/react-components/src/components/AppFrame/types.ts
```diff
@@ -0,0 +1,69 @@
+import * as React from 'react';
+
+export interface INavBarListProps {
+  className?: string;
+}
+
+export interface INavBarOption {
```
VadymBezpalko, line 7: I'd suggest storing the types near their actual usage - in the `<Component>/types.ts` files

packages/react-components/src/components/AppFrame/components/NavBar/NavBar.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+const baseClass = 'nav-bar';
+
+export const NavBar: React.FC<INavBarProps> = ({
+  className,
+  topNavBarNode,
+  bottomNavBarNode,
+}) => {
+  return (
+    <div role="navigation" className={cx(styles[baseClass], className)}>
```
VadymBezpalko, line 17: we can use `nav` element as it more semantically correct and there will be no need to add role attr explicitly. 
Also, this component could also extend the navite HTML <nav> props and spread them to allow developers pass more props if needed. I'd suggest using this strategy for other components as well.

packages/react-components/src/components/AppFrame/AppFrame.mdx
```diff
@@ -0,0 +1,154 @@
...
+- `NavigationList`
+- `NavigationItem`
+
+#### `Navigation` component
+
+It mainly serves as a container for other navigation elements. The styles have been prepared to provide enough space to place a list of items. If we want to make another group of navigation items, the next group will be displayed at the bottom of the navigation.
+
+#### `NavigationList` component
+
+Is another container for building navigation, responsible for the appropriate arrangement of navigation buttons. In addition to the container function, it is also responsible for enabling scrolling of elements when there is not enough space to display them using the `scrollable` prop, which is not defined by default.
```
VadymBezpalko, line 37: The `NavigationGroup` component is a container responsible for the appropriate grouping of `NavigationItem` elements. Should be a direct child of the `Navigation` component. In addition to its container function, it enables the scrolling of elements when there is insufficient space for their display. This feature can be activated by passing the prop `scrollable={true}`.

packages/react-components/src/components/AppFrame/AppFrame.mdx
```diff
@@ -0,0 +1,154 @@
...
+
+#### `NavigationItem` component
+
+The final element for building navigation, which is responsible for displaying the buttons. This component will display buttons with an icon (or any other element you pass, but remember about the size limitations of this element), a tooltip displayed on the hover with the name of the given element passed by `label` which can be turned off if necessary, and also badge which is optional.
```
VadymBezpalko, line 41: The `NavigationItem` component is responsible for displaying navigation elements. It displays buttons with an icon (or any other element you provide, keeping in mind the size limitations of this component). Additionally, a tooltip displaying the name of the element can appear on hover, as specified by the label prop; this can be disabled if necessary. The badge prop is also optional.

packages/react-components/src/components/AppFrame/AppFrame.mdx
```diff
@@ -0,0 +1,154 @@
...
+Side navigation passed by the `sideNavigation` prop will be displayed in an appropriate container, located next to the application frame. By default, the navigation will be visible, but the component allows you to control its visibility.
+For this purpose, `AppFrameProvider` was created, which is used by default inside the `AppFrame` component.
+
+`AppFrameProvider` gives us access to the following properties:
+- `isSubNavBarVisible` - defines whether side navigation is open/closed
+- `toggleSubNavBarVisibility` - function changing the value of `isSubNavBarVisible`
+
+Thanks to the provider, you can easily implement the ability to close and open this panel in your own UI.
+
+```jsx
```
VadymBezpalko, line 144: include the import statements in this example

packages/react-components/src/components/AppFrame/AppFrame.mdx
```diff
@@ -0,0 +1,154 @@
...
+Unlike `NavigationList` it is not a component required to create a side navigation structure. Its main task is to aggregate elements under a label, with the option of enabling expansion of the list of elements via the `isCollapsible` prop.
+
+#### `SideNavigationItem` component
+
+It is the final element for building side navigation, responsible for displaying the button. In addition to displaying the button with the given label, it can be supplemented by passing your own elements via the `leftNode` and `rightNode` props.
+
+#### Side navigation implementation example
+
+```jsx
+<AppFrame
```
VadymBezpalko, line 109: include the import statements in this example

packages/react-components/src/components/AppFrame/components/SideNavigationItem/types.ts
```diff
@@ -0,0 +1,46 @@
...
+
+import { ComponentCoreProps } from '../../../../utils/types';
+
+export interface ISideNavigationItemProps extends ComponentCoreProps {
+  /**
+   * The label of the item
+   */
+  label: React.ReactNode;
+  /**
+   * It will display the custom element on the right side of the side navigation item
```
ashbork, line 11: "It will display the custom element..." is janky wording, IMO. Could we go for "Displays the custom element...", or, better yet, "Custom element"? Same for other cases where similar wording appears.

ashbork:
Hi, please note that this comment is written from a much smaller product's perspective. 
We've got plenty of sections that aren't supposed to be collapsible - settings are a prominent example as we just don't have quite as many of them as LC does. As it stands, non-collapsible (pictured on the top) groups look a bit weird - the header doesn't really seem like one, it resembles an item that was just shifted a bit. The bottom (collapsible) one is better, but as I mentioned, we don't need collapsing for some sections and we explicitly **don't want it** in other places. I'd love it if you could reach out to one of our designers and see if there's some nice middle ground there.
image

packages/react-components/src/components/AppFrame/AppFrame.stories.tsx
```diff
@@ -0,0 +1,170 @@
...
+                url="#"
+              />
+            ))}
+            <NavigationItem
+              id="user"
+              label="User"
+              disableOpacity
+              disableTooltip
+              icon={
+                <Tooltip
```
ashbork, line 138: Non-blocking: it would be so cool to have tooltips out of the box, with no need for defining them each time

packages/react-components/src/components/AppFrame/components/NavigationItem/types.ts
```diff
@@ -0,0 +1,42 @@
...
+   */
+  url: string;
+  /**
+   * Specify whether the tooltip should be disabled after hover
+   */
+  disableTooltip?: boolean;
+  /**
+   * Specify whether the opacity should be disabled
+   */
+  disableOpacity?: boolean;
```
ashbork, line 29: unclear what that means, the description doesn't help much

packages/react-components/src/components/AppFrame/components/SideNavigationGroup/types.ts
```diff
@@ -0,0 +1,34 @@
...
+  /**
+   * The function to call on item hover
+   */
+  onItemHover?: () => void;
+  /**
+   * Specify whether the list should be open on the first render
+   */
+  shouldOpenOnInit?: boolean;
+  /**
+   * Specify whether the list should be open if item is active
```
ashbork, line 31: Could be made more specific,
```suggestion
   * Specify whether the list should be open if an item within the group is active
```

ashbork:
a11y: SideNavigationItems don't seem to be keyboard-focusable!

packages/react-components/src/components/AppFrame/AppFrame.stories.tsx
marcinsawicki: `NavigationItem` has built-in tooltips displaying the label.
This is only an example of how to create the navigation element with its own tooltip (like Avatar with profile menu in the AA)

ashbork: Gotcha, that's great. Resolving

vladko-uxds (approved):
👍

packages/react-components/src/components/AppFrame/components/SideNavigationGroup/SideNavigationGroup.tsx
```diff
@@ -0,0 +1,132 @@
...
+    if (!isCollapsible) {
+      return setIsSideMenuVisible(true);
+    }
+
+    if (isOpen) {
+      return setIsSideMenuVisible(true);
+    }
+
+    return setIsSideMenuVisible(false);
+  }, [isOpen]);
```
VadymBezpalko, line 78: Why `isCollapsible` is not in deps array?

The hook can be optimized to just this:

```suggestion
  React.useEffect(() => {
    return setIsSideMenuVisible(!isCollapsible || isOpen);
  }, [isOpen]);
```

packages/react-components/src/components/AppFrame/components/SideNavigationGroup/SideNavigationGroup.tsx
```diff
@@ -0,0 +1,132 @@
...
+  children,
+  isCollapsible,
+  onItemHover,
+  shouldOpenOnInit = false,
+  shouldOpenOnActive = false,
+}) => {
+  const [isOpen, setIsOpen] = React.useState(
+    !isCollapsible || shouldOpenOnInit
+  );
+  const [isSideMenuVisible, setIsSideMenuVisible] = React.useState(false);
```
VadymBezpalko, line 30: Why is this state named like this? How is this state related to side menu visibility?

VadymBezpalko, line 30: Why are there 2 states that are responsible for children visibility?

packages/react-components/src/components/AppFrame/stories-helpers.tsx
```diff
@@ -0,0 +1,349 @@
...
+        isActive={activeSubItem === 2}
+        onClick={() => handler(2)}
+        shouldKeepIconSpace={false}
+      />
+      <SideNavigationItem
+        label="Option 4"
+        isActive={activeSubItem === 3}
+        onClick={() => handler(3)}
+        shouldKeepIconSpace={false}
+      />
```
VadymBezpalko, line 220: Why not `.map`?

packages/react-components/src/components/AppFrame/stories-helpers.tsx
```diff
@@ -0,0 +1,349 @@
...
+      <SideNavigationItem
+        label="Option 11"
+        isActive={activeSubItem === 10}
+        onClick={() => handler(10)}
+      />
+      <SideNavigationItem
+        label="Option 12"
+        isActive={activeSubItem === 11}
+        onClick={() => handler(11)}
+      />
```
VadymBezpalko, line 301: Why not `map`?

packages/react-components/src/components/AppFrame/stories-helpers.tsx
```diff
@@ -0,0 +1,349 @@
...
+      shouldKeepIconSpace={false}
+      isActive={activeSubItem === 2}
+      onClick={() => handler(2)}
+    />
+    <SideNavigationItem
+      label="Option 4 with very long label name"
+      shouldKeepIconSpace={false}
+      isActive={activeSubItem === 3}
+      onClick={() => handler(3)}
+    />
```
VadymBezpalko, line 118: The code can be reduced using an array and a map function

packages/react-components/src/components/AppFrame/AppFrame.mdx
```diff
@@ -0,0 +1,164 @@
...
+- `SideNavigationGroup`
+- `SideNavigationItem`
+
+#### `SideNavigation` component
+
+It mainly serves as a container for side navigation elements. An addition in this component is props `title` which will display an additional header above the list of navigation buttons.
+
+#### `SideNavigationGroup` component
+
+Unlike `NavigationGroup` it is not a component required to create a side navigation structure. Its main task is to aggregate elements under a label, with the option of enabling expansion of the list of elements via the `isCollapsible` prop.
```
VadymBezpalko, line 103: In my opinion, it is more effective to first describe the purpose of this component to developers, rather than highlighting its differences from other components. Therefore, I suggest starting the description as follows: "The SideNavigationGroup component is used to create collapsible items within the SideNavigation. It groups the `SideNavigationItem` items under same category."

VadymBezpalko (approved):

rafalluksza (approved):

ashbork (approved):

PR #1252, opened 2024-07-18
feat(Picker): set clearSearchAfterSelection as true by default

VadymBezpalko:
Resolves: [Slack thread](https://text.slack.com/archives/CCRBKU2AD/p1721305366433939)

## Description
Set clearSearchAfterSelection as true by default

## Storybook

https://feature-picker-clean-by-default--613a8e945a5665003a05113b.chromatic.com

## Checklist

**Obligatory:**

- [x] Add correct label

### DIFF
--- a/packages/react-components/src/components/Picker/Picker.tsx
+++ b/packages/react-components/src/components/Picker/Picker.tsx
@@ -35,7 +35,7 @@ export const Picker: React.FC<IPickerProps> = ({
   onOpen,
   onClose,
   openedOnInit = false,
-  clearSearchAfterSelection,
+  clearSearchAfterSelection = true,
   onSelect,
   placement,
   floatingStrategy,


vladko-uxds (approved):

packages/react-components/src/components/Picker/Picker.tsx
```diff
@@ -35,7 +35,7 @@ export const Picker: React.FC<IPickerProps> = ({
   onOpen,
   onClose,
   openedOnInit = false,
-  clearSearchAfterSelection,
+  clearSearchAfterSelection = true,
```
marcinsawicki, line 38: IMO we shouldn't treat that change as a fix. 🤔 It's a breaking change, changing how the search works on every component implementation that does not use this prop.

marcinsawicki (approved):

Issue #1232, opened 2024-07-08
[Badge] - Update variants


vladko-uxds:
New variants
- Update/Setup
- Status badges (Do not accept, Accept, Offilne)

![Image](https://github.com/livechat/design-system/assets/125188605/655b538e-19d6-462d-a522-1040975de309)
![Image](https://github.com/livechat/design-system/assets/125188605/beced6d1-e8c1-424d-ad0d-a53e6de1f3a2)

Deprecated:
 - Dot
 - Secondary  

![Image](https://github.com/livechat/design-system/assets/125188605/322dcb76-3fc6-49a0-b579-cd77ac58784c)

[See design spec ](https://www.figma.com/design/9FDwjR8lYvincseDkKypC4/Component-Documentations?node-id=731-59622&t=ETkste7oqmpQJbL4-4)

PR #1258, opened 2024-07-25
feat(Badge): deprecate some props and new UpdateBadge and StatusBadge components

marcinsawicki:
Resolves: #1232 

## Description
Changes in `Badge` component:
- `kind` - `secondary` is deprecated (added information)
- `size` - `large` is deprecated (added information)
- `type` - `dot` is deprecated (added information)

Added new `UpdateBadge` and `StatusBadge` components with stories related to the `Badge` component.

## Storybook

https://feature-1232--613a8e945a5665003a05113b.chromatic.com

## Checklist

**Obligatory:**

- [x] Add correct label


packages/react-components/src/components/StatusBadge/StatusBadge.module.scss
```diff
@@ -0,0 +1,79 @@
...
+
+.#{$base-class} {
+  display: flex;
+  border-style: solid;
+  border-color: var(--surface-primary-default);
+
+  &--3XS {
+    @include size(4px);
+
+    border-width: 1px;
```
VadymBezpalko, line 17: I'd suggest including border-width into the size mixin and make it as a second parameter

VadymBezpalko (approved):
Good job 👍 
I left some minor comments and also have a suggestion to group the default badge files in the storybook as well - in the similar style as StatusBadge and UpdateBadge. So the Badge title probably should be smth like this: `title: 'Components/Badge/Badge',`

vladko-uxds (approved):
👍

Issue #1253, opened 2024-07-22
[Details card] - expand card by click on title/header


vladko-uxds:
Based on feedback we need to update expand/collapse action.

![Image](https://github.com/user-attachments/assets/14470bf4-364e-4bc5-a5d9-b8abb82e40a2)

**Actual**
Card expand only on click in icon (chevron right card side)

**Expected**
User need a way to expand card when chevron icon is not available (covered with some content or component)

**Solution**
Make card expand/collapse by click on title/header

PR #1261, opened 2024-07-30
fix(DetailsCard): expand card on label click

marcinsawicki:
Resolves: #1253 

## Description
Added the possibility of opening the card by clicking the label.
Added guard that will not trigger opening in case the user click the interactive element (input, button, etc.)

## Storybook

https://feature-1253--613a8e945a5665003a05113b.chromatic.com

## Checklist

**Obligatory:**

- [x] Add correct label


packages/react-components/src/components/DetailsCard/DetailsCard.tsx
```diff
@@ -81,6 +81,19 @@ export const DetailsCard: React.FC<IDetailsCardProps> = ({
...
+  const handleLabelClick = (
+    e: React.MouseEvent<HTMLDivElement, MouseEvent>
+  ) => {
+    const target = e.target as HTMLElement;
+    const targetTagName = target.tagName.toLowerCase();
+    const nonInteractiveTags = ['input', 'button', 'select', 'textarea', 'a'];
+
+    if (!nonInteractiveTags.includes(targetTagName)) {
+      setIsOpen((prevValue) => !prevValue);
+      onClick?.();
```
VadymBezpalko, line 93: we can pass the click event to the onClick callback

packages/react-components/src/components/DetailsCard/DetailsCard.tsx
```diff
@@ -81,6 +81,19 @@ export const DetailsCard: React.FC<IDetailsCardProps> = ({
     onClick?.();
   };
 
+  const handleLabelClick = (
+    e: React.MouseEvent<HTMLDivElement, MouseEvent>
+  ) => {
+    const target = e.target as HTMLElement;
+    const targetTagName = target.tagName.toLowerCase();
+    const nonInteractiveTags = ['input', 'button', 'select', 'textarea', 'a'];
```
VadymBezpalko, line 89: we can define those outside of the component's body

VadymBezpalko (approved):

vladko-uxds (approved):

Issue #1262, opened 2024-07-30
[BUG] Picker and keyboard issue


zavi007:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

When using search + keyboard in picker, the highlighted option is not actually applied on enter, seems like the focus is still on previous item

https://github.com/user-attachments/assets/7bf975d1-5c7a-453c-bc65-f2de0ad67434

### Steps To Reproduce

1. open picker
2. use keyboard to search for item
3. hit enter
4. search for another item (item should be highlighted, do not use key down)
5. hit enter 

### Expected Behavior

Highlighted item should be added on enter

### Design system package(s) version

examples:
  @livechat/design-system-react-components": "^1.29.0"

PR #1267, opened 2024-07-31
fix(Picker): fixed the issue with virtual focus when filtering the list

marcinsawicki:
Resolves: #1262 

## Description
Fixed the issue with incorrectly adding elements using keyboard controls when the first element is selected, and then the list is filtered.

## Storybook

https://feature-1262--613a8e945a5665003a05113b.chromatic.com

## Checklist

**Obligatory:**

- [x] Add correct label


VadymBezpalko (approved):
🔥

PR #1277, opened 2024-08-06
feat(AppFrame): Navigation top bar and alerts

ashbork:
## Description
Adds two components: `NavigationTopBar` and `NavigationTopBarAlert` (`NavigationTopBar.Alert`). Allows for adding alerts of different kinds with sensible defaults following the DS guidelines.

## Storybook

https://feature-navigation-top-bar--613a8e945a5665003a05113b.chromatic.com

## Checklist

**Obligatory:**

- [x] Add correct label


packages/react-components/src/components/AppFrame/components/NavigationTopBar/NavigationTopBar.module.scss
```diff
@@ -0,0 +1,112 @@
...
+    color: var(--content-locked-white);
+    font-size: 14px;
+
+    &--open {
+      padding-block: var(--spacing-2);
+      min-height: 42px;
+    }
+
+    &__wrapper {
+      // Grid trick: https://css-tricks.com/css-grid-can-do-auto-height-transitions/
```
ashbork, line 32: This may be confusing - recommend reading before asking abt the grid wrapper

packages/react-components/src/components/AppFrame/components/NavigationTopBar/NavigationTopBar.module.scss
```diff
@@ -0,0 +1,112 @@
...
+    &--error {
+      background-color: var(--surface-accent-emphasis-high-negative);
+    }
+
+    &__children-wrapper {
+      display: flex;
+      gap: var(--spacing-4);
+      align-items: center;
+      justify-content: center;
+      width: calc(100% - 48px);
```
ashbork, line 67: places the 'x' in a sensible location

packages/react-components/src/components/AppFrame/components/NavigationTopBar/NavigationTopBar.module.scss
```diff
@@ -0,0 +1,112 @@
...
+
+      @media screen and (width <= 768px) {
+        flex-direction: column;
+        gap: var(--spacing-2);
+      }
+    }
+
+    &__close-button {
```
ashbork, line 75: could use a DS button instead, but none of them are quite what we want - an inert icon btn with no hover animation

packages/react-components/src/components/AppFrame/components/NavigationTopBar/NavigationTopBar.tsx
```diff
@@ -0,0 +1,140 @@
...
+  className,
+  kind = 'info',
+  closeButton,
+  children,
+  primaryCta,
+  secondaryCta,
+  show = true,
+}) => {
+  const alertRef = React.useRef<HTMLDivElement>(null);
+  const { isMounted, isOpen } = useAppFrameAnimations({
```
ashbork, line 62: While this hook isn't specifically designed for this component, it does work

packages/react-components/src/components/AppFrame/components/NavigationTopBar/NavigationTopBar.tsx
```diff
@@ -0,0 +1,140 @@
...
+    isVisible: show,
+    elementRef: alertRef,
+  });
+
+  const isMobile = window.matchMedia('(max-width: 768px)').matches;
```
ashbork, line 67: There should be a better way, but I didn't find anything that we're using in DS currently - the check only updates on rerenders which isn't great, we don't want to `useEffect` either. CSS feels janky for this purpose. Thoughts?

vladko-uxds (approved):
👍

packages/react-components/src/components/AppFrame/AppFrame.mdx
```diff
@@ -160,6 +160,74 @@ const SideNavigationSwitcher: FC = () => {
...
+    <NavigationTopBarAlert kind="info">
+    Hello, this is an alert!
+    </NavigationTopBarAlert>
+  </NavigationTopBar>
+  App content
+</AppFrame>
+```
+
+
+### Top bar alert <a id="TopBarAlert" />
```
VadymBezpalko, line 188: We can include this section in "How to build the navigation" navigation links (line 20)

packages/react-components/src/components/AppFrame/components/NavigationTopBar/NavigationTopBar.module.scss
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    &__children-wrapper {
+      display: flex;
+      gap: var(--spacing-4);
+      align-items: center;
+      justify-content: center;
+      width: calc(100% - 48px);
+
+      @media screen and (width <= 768px) {
```
VadymBezpalko, line 76: we can store width value in the variable and reuse it in the code below

packages/react-components/src/components/AppFrame/components/NavigationTopBar/NavigationTopBar.tsx
marcinsawicki: This hook is designed for `AppFrame`, and top bar is a part of the whole component, so I don't see any problems with using it here :)

packages/react-components/src/components/AppFrame/AppFrame.mdx
```diff
@@ -160,6 +160,74 @@ const SideNavigationSwitcher: FC = () => {
...
+import { AppFrame, Navigation, NavigationTopBar, NavigationTopBarAlert } from '@livechat/design-system-react-components';
+
+<AppFrame
+  navigation={<Navigation>...</Navigation>}
+>
+  <NavigationTopBar>
+    <NavigationTopBarAlert kind="info">
+    Hello, this is an alert!
+    </NavigationTopBarAlert>
+  </NavigationTopBar>
```
marcinsawicki, line 182: Shouldn't this be placed in the `topBar` prop in this example? 🤔

packages/react-components/src/components/AppFrame/components/NavigationTopBar/NavigationTopBar.tsx
marcinsawicki: Why do we need to change the kind of button for mobile only? 🤔

marcinsawicki:
First of all, great job with adding the new sub-components for the `AppFrame` and improving how the top bar works. 🙌 
Left some comments.
Missing unit tests.

packages/react-components/src/components/AppFrame/components/NavigationTopBar/types.ts
```diff
@@ -0,0 +1,61 @@
...
+   * */
+  secondaryCta?: {
+    label: string;
+    onClick: () => void;
+  };
+
+  /**
+   * Show or hide the alert, defaults to `true`. Changes to this prop are animated - the alert will enter and leave smoothly.
+   * */
+  show?: boolean;
```
VadymBezpalko, line 60: Historically, our codebase commonly uses `isVisible` naming. Therefore, I suggest using similar naming to maintain a consistent pattern.
P.S.: Consistent naming patterns are something I highly desire, but for various reasons, we don't have them yet. 😄

VadymBezpalko (approved):

packages/react-components/src/components/AppFrame/components/NavigationTopBar/NavigationTopBar.tsx
ashbork: That's how it is in the design, presumably because the secondary button takes up much less space

packages/react-components/src/components/AppFrame/components/NavigationTopBar/types.ts
ashbork: > but for various reasons, we don't have them yet. 😄

sorry to have been one of these reasons! it's `isVisible` now

VadymBezpalko: @ashbork Haha, no, I meant legacy code from the previous DS version, since part of the new API is a clone of the old one. 😄 
You and the other contributors are actually the reason why the codebase keeps improving, so great job and thanks for your effort! 🏆

marcinsawicki (approved):

PR #1290, opened 2024-08-13
fix(AppFrame): Fixes and Improvements

VadymBezpalko:
## Description
- Adding `disabled` state to NavigationItem
- Making `url` and `label` fields optional
- Removing `disableTooltip` prop as it will no longer be needed after making `label` field optional

## Storybook

https://feature-appframe-fixes--613a8e945a5665003a05113b.chromatic.com

## Checklist

**Obligatory:**

- [ x Self review (use this as your final check for proposed changes before requesting the review)
- [x] Add correct label

### DIFF
--- a/packages/react-components/.storybook/main.ts
+++ b/packages/react-components/.storybook/main.ts
@@ -19,7 +19,7 @@ const config: StorybookConfig = {
   },
   docs: {},
   typescript: {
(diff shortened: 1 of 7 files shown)


ashbork (approved):

vladko-uxds (approved):
👍

PR #1293, opened 2024-08-14
fix(ProductSwitcher): lock in dark mode; add textUrl Prop

VadymBezpalko:
## Description
Adding textURL prop and locking switcher in dark mode.

## Storybook

https://feature-switcher-tweaks--613a8e945a5665003a05113b.chromatic.com

## Checklist

**Obligatory:**

- [x] Add correct label


mreszke (approved):

vladko-uxds (approved):
👍

PR #1294, opened 2024-08-14
fix(Picker): make customElement props optional

ashbork:
## Description
Allows you to only provide one of `selectedItemBody` and `listItemBody`.
Previously, you had to add both or the component would break while rendering

## Storybook

https://feature-picker-optional-customelement--613a8e945a5665003a05113b.chromatic.com

## Checklist

**Obligatory:**

- [x] Add correct label


VadymBezpalko (approved):
👍

Issue #1089, opened 2024-04-25
[Input] - AutoComplete


vladko-uxds:
LIst of additional properties:

- [ ] Single AutoComplete: As the user types, the system predicts and displays a single item (EX: email) suggestion directly in the input field.
- [ ] Multiple AutoComplete: Similar as single AutoComplete - system predicts and displays a list of matched items/suggestions and the user can select one item from the list to complete the input. Example [AutoComplete Ant](https://ant.design/components/auto-complete)

PR #1295, opened 2024-08-14
feat(Input): new AutoComplete component

ashbork:
Resolves: #1089

## Description
Adds a new component, `AutoComplete`, which is a wrapper around `Input` and augments it with autocomplete capabilities.

## Storybook

https://feature-autocomplete--613a8e945a5665003a05113b.chromatic.com

## Checklist

**Obligatory:**

- [x] Add correct label


packages/react-components/src/components/Input/AutoComplete.tsx
```diff
@@ -0,0 +1,225 @@
...
+  FloatingNode,
+  FloatingPortal,
+  Placement,
+  Strategy,
+} from '@floating-ui/react';
+
+import { IPickerListItem, PickerList } from '../Picker';
+import { DEFAULT_LIST_HEIGHT, MIN_LIST_HEIGHT } from '../Picker/constants';
+import { useFloatingPicker } from '../Picker/hooks/useFloatingPicker';
+import { usePickerItems } from '../Picker/hooks/usePickerItems';
```
ashbork, line 13: These hooks, while not built specifically for this, work perfectly fine for building a "headless" `PickerList` (one without the usual `PickerTrigger`)

packages/react-components/src/components/Picker/components/PickerList.tsx
```diff
@@ -32,6 +32,7 @@ export interface IPickerListProps {
     userProps?: React.HTMLProps<HTMLElement> | undefined
   ) => Record<string, unknown>;
   emptyStateText?: string;
+  hideWhenEmpty?: boolean;
```
ashbork, line 35: non-breaking API change: prop preventing the list from appearing when it's empty. I needed that for the AC component

packages/react-components/src/components/Input/AutoComplete.tsx
```diff
@@ -0,0 +1,250 @@
...
+      }
+      inputProps.onKeyDown?.(event);
+    };
+
+    const handleAutoComplete = (value: { key: string }[] | null) => {
+      if (value) {
+        const key = value[0].key;
+        handleValueChange({
+          target: { value: key },
+        } as React.ChangeEvent<HTMLInputElement>);
```
marcinsawicki, line 151: Do we need assertion here? 🤔

ashbork: Yes, otherwise it complains about a lack of 
```
addEventListener
dispatchEvent
removeEventListener
```

packages/react-components/src/components/Input/AutoComplete.tsx
```diff
@@ -0,0 +1,250 @@
...
+  items: IPickerListItem[],
+  single: boolean,
+  hideIfExactMatch: boolean,
+  inputValue: string
+): IPickerListItem[] => {
+  const isExactMatch = items.length === 1 && items[0].name === inputValue;
+  const shownItems = single ? items.slice(0, 1) : items;
+
+  return hideIfExactMatch && isExactMatch ? [] : shownItems;
+};
```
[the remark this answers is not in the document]
ashbork: Sure!

packages/react-components/src/components/Input/AutoComplete.tsx
```diff
@@ -0,0 +1,250 @@
...
+  maxListHeight?: number;
+  /** Placement of the list */
+  placement?: Placement;
+  /** Floating strategy */
+  floatingStrategy?: Strategy;
+  /** Type of the input. If true, only shows one matching item. */
+  single?: boolean;
+  /** If true, the option list will be hidden if there is only one option and it is an exact match to the input value. */
+  hideIfExactMatch?: boolean;
+}
```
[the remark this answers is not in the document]
ashbork: OK!

marcinsawicki (commented):
Left some comments.
In my opinion, this is the new component, and it should be placed in dedicated place in our files structure (`components/AutoComplete`), it will be easier to maintain it in the future. You can use `title: 'Forms/Input',` in the `.stories.` file to assign it to "Forms" in the SB.
Also, I'm not sure if "AutoComplete" is the good name 🤔 . What about "AutoCompleteInput"?

ashbork:
The current name seems like the industry standard:

https://ant.design/components/auto-complete
https://mui.com/material-ui/react-autocomplete/

packages/react-components/src/components/AutoComplete/AutoComplete.mdx
```diff
@@ -0,0 +1,84 @@
...
+```jsx
+<AutoComplete options={["Option 1", "Option 2", "Option 3"]} placeholder="Placeholder text" onChange={(e) => {}} />
+```
+
+## Component API <a id="ComponentAPI" />
+
+The `AutoComplete` component is a wrapper around the `Input` component that provides a list of suggestions based on the user's Input. It requires a list of unique `options` to be passed. All of `Input`'s props are available and the component can be used both controlled and uncontrolled.
+
+
+<ArgTypes of={AutoCompleteStories.Default} sort="requiredFirst" />
```
marcinsawicki, line 24: We keep the information about the API component at the end of the documentation, please change it to be the same as the others docs pages

packages/react-components/src/components/AutoComplete/AutoComplete.stories.tsx
```diff
@@ -0,0 +1,170 @@
...
+    },
+    single: {
+      description: 'If true, only shows the first matching item.',
+      table: {
+        defaultValue: { summary: 'false' },
+      },
+      control: {
+        type: 'boolean',
+      },
+    },
```
marcinsawicki, line 58: Do we need these descriptions for props here? 🤔 
Are the comments on the interface not enough?

packages/react-components/src/components/Picker/components/PickerList.tsx
marcinsawicki: No need to show any empty state? 🤔

marcinsawicki (approved):

packages/react-components/src/components/AutoComplete/AutoComplete.stories.tsx
ashbork: They were necessary before, when AutoComplete was in the Input story, but that's no longer the case.

packages/react-components/src/components/AutoComplete/index.ts
```diff
@@ -0,0 +1,5 @@
+import { AutoComplete } from './AutoComplete';
+import { AutoCompleteProps } from './types';
+
+export { AutoComplete };
+export type { AutoCompleteProps };
```
VadymBezpalko, line 5: You need to add export statements to the `src/index.ts` file to make the AutoComplete component available for usage.

packages/react-components/src/components/AutoComplete/types.ts
```diff
@@ -0,0 +1,35 @@
+import * as React from 'react';
+
+import { Placement, Strategy } from '@floating-ui/react';
+
+import { InputProps } from '../Input';
+import { IPickerListItem } from '../Picker';
+
+// selectedItemBody is unnecessary for AutoCompleteListItem, key should be === name
+export type IAutoCompleteListItem = Omit<
```
VadymBezpalko, line 9: What if the names would be duplicated in the options? Key attributes should be unique.

packages/react-components/src/components/AutoComplete/helpers.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+export const buildOptionsFromStrings = (options: string[]): IPickerListItem[] =>
+  options.map((option) => ({ name: option, key: option }));
+
+export const buildOptionsFromAutoCompleteListItems = (
+  options: IAutoCompleteListItem[]
+): IPickerListItem[] =>
+  options.map(({ customElement, name, ...rest }) => ({
+    name,
+    key: name,
```
VadymBezpalko, line 20: What if somebody passes options with duplicated names?

packages/react-components/src/components/AutoComplete/types.ts
ashbork: for this component's purposes they shouldn't be - it'd be impossible to differentiate between two options and typing in their name would show both of them

packages/react-components/src/components/AutoComplete/helpers.tsx
ashbork: what if someone passes options with duplicated keys to `Picker`? I don't think we check that anywhere

VadymBezpalko: Yes, but keys should be unique by design, while names don't need to be. I mentioned this because we previously encountered an issue with the Picker, where we used names as keys, which caused problems with items rendering.

VadymBezpalko: But taking into account your other reply I think we can leave it like this 👌

VadymBezpalko (approved):
Great job! 🔥 🏆

vladko-uxds:
@ashbork  
When use` keyboard to focus` field then 

Uploading Monosnap screencast 2024-09-02 12-10-10.mp4…

typing text - list with auto-complete options does't display

ashbork:
Fixed the issue @vladko-uxds mentioned

vladko-uxds (approved):
👍

PR #1310, opened 2024-08-29
feat(Accordion): new component

marcinsawicki:
Resolves: #{issue-number}

## Description
Accordion new component.

## Storybook

https://feature-accordion-new-component--613a8e945a5665003a05113b.chromatic.com/?path=/story/components-accordion--examples

## Checklist

**Obligatory:**

- [x] Add correct label


packages/react-components/src/components/Accordion/Accordion.mdx
```diff
@@ -0,0 +1,27 @@
...
+
+## Intro <a id="Intro" />
+
+Accordion is a simple component for building expandable tiles that display provided content when opened.
+
+<Canvas of={AccordionStories.Examples} sourceState="none" />
+
+#### Example implementation
+
+```jsx
```
VadymBezpalko, line 19: Please add an import path for this example

packages/react-components/src/components/Accordion/Accordion.module.scss
```diff
@@ -0,0 +1,88 @@
...
+      border-color: transparent;
+    }
+  }
+
+  &__chevron {
+    position: absolute;
+    top: 20px;
+    right: 20px;
+    transition: inherit;
+    z-index: 0;
```
VadymBezpalko, line 55: do we need to set it to 0 here?

packages/react-components/src/components/Accordion/Accordion.module.scss
```diff
@@ -0,0 +1,88 @@
...
+
+    &--open {
+      transform: rotate(180deg);
+    }
+  }
+
+  &__label {
+    z-index: 1;
```
VadymBezpalko, line 63: do we need this z-index here?

packages/react-components/src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,142 @@
...
+  const mergedClassName = cx(
+    styles[baseClass],
+    styles[`${baseClass}--${kind}`],
+    {
+      [styles[`${baseClass}--open`]]: isExpanded,
+    },
+    className
+  );
+
+  const handleStateChange = (state: boolean) => {
```
VadymBezpalko, line 52: I'd suggest avoiding "empty words" like "state", "value" etc, because those words alone are not providing any context to the function/variable name - because the word "state" could mean anything. So I'd suggest using in this case the word "visibility": `handleVisibilityChange`, `newVisibilityValue`.

packages/react-components/src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,142 @@
...
+          setSize(newSize);
+          previousSizeRef.current = newSize;
+        }
+      });
+
+      resizeObserver.observe(contentRef.current);
+
+      return () => resizeObserver.disconnect();
+    }
+  }, [contentRef]);
```
VadymBezpalko, line 92: Are you sure this effect works as intended? The contentRef is a reference so it's mutable and the useEffect between the renders will not detect its changes.

packages/react-components/src/components/Accordion/components/AccordionAnimatedLabel.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+      if (openRef) {
+        openRef.addEventListener(
+          'transitionend',
+          () => {
+            setIsOpenMounted(false);
+          },
+          { once: true }
+        );
+      }
```
VadymBezpalko, line 56: Both sides of the `if` statement are nearly identical and can be simplified using the ternary operator and the `isOpen` value itself

packages/react-components/src/components/Accordion/components/AccordionAnimatedLabel.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+      requestAnimationFrame(() => setIsClosedVisible(true));
+
+      if (openRef) {
+        openRef.addEventListener(
+          'transitionend',
+          () => {
+            setIsOpenMounted(false);
+          },
+          { once: true }
```
VadymBezpalko, line 54: This option removes the event listener only if it was invoked, but if not - we need to clear ourselves.

packages/react-components/src/components/Accordion/components/AccordionMultilineElement.tsx
```diff
@@ -0,0 +1,59 @@
...
+          setSize(newSize);
+          previousMultilineSizeRef.current = newSize;
+        }
+      });
+
+      resizeObserver.observe(multilineRef.current);
+
+      return () => resizeObserver.disconnect();
+    }
+  }, [multilineRef]);
```
VadymBezpalko, line 45: references in react a mutable, so the effect cannot detect the ref changes between the renders

VadymBezpalko, line 45: This code is nearly identical to the effect from `Accordion.tsx` file.

vladko-uxds:
Catch a bug with picker inside accordion:
- Can't pick a list option by using keyboard, only close and open but not pick an item
https://github.com/user-attachments/assets/647dc530-6102-4955-a628-6a7a774040a8

packages/react-components/src/components/Accordion/Accordion.module.scss
marcinsawicki: Yes, it is used in animation to fade between open/close label variants.

packages/react-components/src/components/Accordion/Accordion.module.scss
marcinsawicki: Yes, it is used in animation to fade between open/close label variants.

packages/react-components/src/components/Accordion/Accordion.module.scss
marcinsawicki: Yes, it makes the whole label clickable (which is expected) what is not covered by chevron icon.

packages/react-components/src/components/Accordion/Accordion.module.scss
marcinsawicki: Yes, it makes the whole label clickable (which is expected) what is not covered by chevron icon.

packages/react-components/src/components/Accordion/Accordion.module.scss
marcinsawicki: I will change that for `pointer-events`

packages/react-components/src/components/Accordion/Accordion.module.scss
marcinsawicki: I will change that for pointer-events

vladko-uxds (approved):
👍

VadymBezpalko (approved):

Issue #1309, opened 2024-08-29
[BUG] ActionMenu: Tooltip inherits opacity from disabled button 


dprzybylek:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

When I add a disabled button with a tooltip, the tooltip inherits the opacity from the disabled button, resulting in an opacity of 0.45 for the tooltip. This causes a visual issue. Could you please look into fixing this?

### Steps To Reproduce

Try to create a menu with disabled item and tooltip explaining why item is disabled

### Expected Behavior

Tooltip doesn't inherit opacity from the button

### Screenshots

<img width="569" alt="image" src="https://github.com/user-attachments/assets/74568213-8c0b-4fb8-ba28-2270e421fe8f">

### Design system package(s) version

"@livechat/design-system-react-components": "^1.34.1"

### Environment

```markdown
chrome, mac
```

PR #1313, opened 2024-09-04
fix(ActionMenu): change opacity to color token for disabled item

marcinsawicki:
Resolves: #1309 

## Description
Added color token for a disabled button to show disabled state. Removed the `opacity` prop, which will solve the problem of displaying tooltips for disabled elements.

## Storybook

https://feature-1309--613a8e945a5665003a05113b.chromatic.com

## Checklist

**Obligatory:**

- [x] Add correct label


VadymBezpalko (approved):

vladko-uxds (approved):
👍

Issue #1307, opened 2024-08-27
[Picker] Choosing picker option by 'Enter' press does not work


EmWa94:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

When `searchDisabled` prop is set to true, choosing picker options by pressing `Enter` doesn't work, e.g bellow ⬇️ 

<img width="408" alt="image" src="https://github.com/user-attachments/assets/ecf6e225-7ab4-417d-af5c-c009f1107419">

<img width="559" alt="image" src="https://github.com/user-attachments/assets/2dc6deae-bc51-4287-ab85-3ad8ceb1e84c">

### Steps To Reproduce

1. Focus Picker and open list of options - by clicking picker or enter press on a focused picker
2. Choose the new option to select by arrow down/up press - it works ok
3. Confirming the chosen option by `Enter` press - it does not work.the 

### Expected Behavior

Confirming focused/chosen list item by `Enter` press select this option.

### Design system package(s) version

"@livechat/design-system-react-components": "1.33.7"

PR #1315, opened 2024-09-05
fix(Picker): fixing keyboard handling when search is disabled

VadymBezpalko:
Resolves: #1307 

## Description
Adding keyboard handling to the picker list when the search input is disabled.

## Storybook

https://feature-1307--613a8e945a5665003a05113b.chromatic.com

## Checklist

**Obligatory:**

- [x] Add correct label


ashbork (approved):

vladko-uxds (approved):
👍

packages/react-components/src/components/Picker/Picker.stories.tsx
```diff
@@ -39,6 +39,8 @@ const PickerComponent = (args: IPickerProps) => {
   return (
     <Picker
       {...args}
+      openedOnInit={false}
+      searchDisabled={true}
```
marcinsawicki, line 43: Why do we set those props manually? 🤔

VadymBezpalko: nope, these are leftovers

marcinsawicki (approved):

Issue #1289, opened 2024-08-13
[AppFrame] - Mobile view & SubscriptionNavigationItem


vladko-uxds:
Based on slack https://text.slack.com/archives/CA945NPQR/p1723533987577029

There is a feature request from HelpDesk and Accounts teams to add a following features to the AppFrame functionality:
- There is a need of Mobile view handling in the AppFrame component. This feature should handle width change with rendering of specific mobile-view navigation items. This set can subset of existing items and also mobile-only nav items. There is an idea to add also a nav item that will work like "More" button and show a popup with additional items.
- There is a request for nav item, that will work like subscription item in AA - will show red background with days counter.

We must discuss these requests and decide if, when, and how we want to implement them.

PR #1319, opened 2024-09-05
feat(AppFrame): mobile view and MobileNavigation component

marcinsawicki:
Resolves: #1289 

## Description
Mobile view for `AppFrame`.
Added required prop `mobileNavigation` and `MobileNavigation` component (act as `Navigation` but for mobile view).
UI changes related to the mobile view (top bars overlap the view).

## Storybook

https://feature-1289--613a8e945a5665003a05113b.chromatic.com

## Checklist

**Obligatory:**

- [x] Add correct label


packages/react-components/src/components/AppFrame/components/NavigationItem/NavigationItem.tsx
```diff
@@ -44,6 +44,7 @@ export const NavigationItem: React.FC<INavigationItemProps> = ({
   isActive,
   onClick,
   className,
+  isMobile,
```
VadymBezpalko, line 47: Do we need to add this prop? Can't we imply this from its wrapper `MobileNavigation`?

packages/react-components/src/components/AppFrame/components/NavigationItem/NavigationItem.tsx
```diff
@@ -87,6 +89,7 @@ export const NavigationItem: React.FC<INavigationItemProps> = ({
             {...props}
           >
             {icon}
+            {isMobile && label}
```
VadymBezpalko, line 92: Instead of using separate prop isMobile, I'd use only the label and show it when this component placed inside of `MobileNavigation`.

packages/react-components/src/components/AppFrame/AppFrame.spec.tsx
```diff
@@ -30,6 +31,17 @@ const defaultProps = {
       />
     </Navigation>
   ),
+  mobileNavigation: (
+    <MobileNavigation data-testid="mobile-navigation">
+      <NavigationItem
+        id="item-1"
+        label="Item 1"
+        icon={<div>Icon</div>}
+        url="#"
```
VadymBezpalko, line 40: The `url` prop is optional, you do not need to pass it.

packages/react-components/src/components/AppFrame/AppFrame.tsx
```diff
@@ -30,30 +32,39 @@ const Frame = (props: IAppFrameProps) => {
     isVisible: isSideNavigationVisible,
     elementRef: sideNavWrapperRef,
   });
+  const { isMobile, handleResize } = useMobileViewDetector({
```
VadymBezpalko, line 35: To avoid confusing the name `handleResize` with a function, I'd suggest to add it for example: `handleResizeRef`

packages/react-components/src/components/AppFrame/types.ts
```diff
@@ -31,6 +35,10 @@ export interface IAppFrameProps extends ComponentCoreProps {
    * The CSS class for the content container
    */
   contentClassName?: string;
+  /**
```
VadymBezpalko, line 38: Please add information about default value

VadymBezpalko (approved):

vladko-uxds (approved):

packages/react-components/src/components/AppFrame/components/NavigationItem/NavigationItem.tsx
marcinsawicki: Moved that flag to the `AppFrameContext`

PR #1359, opened 2024-09-19
feat(Accordion): change label height based on child component

kubaczerwinski77:
Resolves: #{issue-number}

## Description
Define the label height based on child component

## Storybook

https://feature-1358--613a8e945a5665003a05113b.chromatic.com

## Checklist

**Obligatory:**

- [x] Add correct label

### DIFF
--- a/packages/react-components/src/components/Accordion/components/AccordionAnimatedLabel.module.scss
+++ b/packages/react-components/src/components/Accordion/components/AccordionAnimatedLabel.module.scss
@@ -5,7 +5,7 @@ $base-class: 'accordion-animated-label';
   position: relative;
   align-items: center;
   transition: all var(--transition-duration-fast-2) ease-in-out;
-  min-height: 21px;
+  min-height: 24px;
 
   &__open,
   &__close {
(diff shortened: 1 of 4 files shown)


packages/react-components/src/components/Accordion/components/AccordionAnimatedLabel.module.scss
```diff
@@ -14,7 +15,19 @@ $base-class: 'accordion-animated-label';
...
   }
 }
+
+@keyframes change-visibility {
+  0% {
+    opacity: 0;
+  }
+
+  70%,
+  100% {
```
mreszke, line 30: why it's done like that? Shouldn't we use shorter duration to make effect faster?

mreszke (approved):

marcinsawicki (approved):

packages/react-components/src/components/Accordion/components/AccordionAnimatedLabel.module.scss
kubaczerwinski77: I guess the results will be similar in both cases :P

PR #1363, opened 2024-09-23
feat(InviteAgents): new component

VadymBezpalko:
## Description
New component: [FIGMA](https://www.figma.com/design/koantnx83TCyrM1dH6LqLx/Refresh-WIP?node-id=3017-62928&node-type=frame&m=dev)

## Storybook

https://feature-invite-agents--613a8e945a5665003a05113b.chromatic.com

## Checklist

**Obligatory:**

- [x] Add correct label

### DIFF
--- a/packages/react-components/src/components/Avatar/Avatar.tsx
+++ b/packages/react-components/src/components/Avatar/Avatar.tsx
@@ -53,6 +53,10 @@ export interface AvatarProps extends ComponentCoreProps {
    * Displays rim
    */
   withRim?: boolean;
+  /**
+   * Custom style for the avatar
+   */
+  style?: React.CSSProperties;
 }
 
 const baseClass = 'avatar';
@@ -68,6 +72,7 @@ export const Avatar: React.FC<AvatarProps> = ({
   text,
   type,
   withRim = false,
+  style,
   ...props
 }) => {
   const isImproperImageSetup = type === 'image' && !src;
@@ -136,7 +141,11 @@ export const Avatar: React.FC<AvatarProps> = ({
   }, [isImproperImageSetup]);
 
   return (
-    <div className={mergedClassNames} style={backgroundStyle} {...props}>
+    <div
+      className={mergedClassNames}
+      style={{ ...backgroundStyle, ...style }}
+      {...props}
+    >
       {withRim && (
         <div
           data-testid={`${baseClass}__rim`}
--- a/packages/react-components/src/components/Avatar/Avatar.tsx
+++ b/packages/react-components/src/components/Avatar/Avatar.tsx
@@ -53,10 +53,6 @@ export interface AvatarProps extends ComponentCoreProps {
    * Displays rim
    */
   withRim?: boolean;
-  /**
-   * Custom style for the avatar
-   */
-  style?: React.CSSProperties;
 }
 
 const baseClass = 'avatar';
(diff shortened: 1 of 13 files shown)


mreszke:
IMO it would be better to remove `onSetUpChatbotClick` from the props and allow to pass an array of custom elements


return (
  <InviteAgents
    agents={agents}
    onAddTeammateClick={onAddTeammateClick}
    customElements={[
       onClick: onSetUpChatbotClick,
       icon: ChatbotIcon,
       text: 'Set up ChatBot'
    ]}
    animatedInviteButton={true}
    tooltipArrowOffset={13}
    className="invite-agents"
  />
);

packages/react-components/src/components/Avatar/Avatar.tsx
```diff
@@ -53,6 +53,10 @@ export interface AvatarProps extends ComponentCoreProps {
    * Displays rim
    */
   withRim?: boolean;
+  /**
+   * Custom style for the avatar
+   */
+  style?: React.CSSProperties;
```
marcinsawicki, line 59: Maybe this should be a part of `ComponentCoreProps`? 🤔

packages/react-components/src/components/InviteAgents/InviteAgents.mdx
```diff
@@ -0,0 +1,51 @@
...
+  {
+    name: 'Alice',
+    email: 'alice@example.com',
+    status: 'available',
+    avatar: 'https://via.placeholder.com/150',
+  },
+]; 
+
+const onAddTeammateClick = () => {
+  console.log('Add Teammate Clicked');
```
marcinsawicki, line 29: I think we shouldn't put the `console.log` in the docs. Simple `() => {}` is enough imo 🤔

packages/react-components/src/components/InviteAgents/InviteAgents.tsx
```diff
@@ -0,0 +1,177 @@
...
+                    +{additionalAgentsNumber}
+                  </Text>
+                )}
+              </div>
+            )
+          }
+        >
+          {hasOnlyUnavailableAgents ? (
+            <Interactive
+              header="No one's available to assist customers"
```
marcinsawicki, line 138: IMO a request to change copy here or elsewhere will appear sooner rather than later. 😅 Shouldn't we leave this open for custom string? 🤔

packages/react-components/src/components/InviteAgents/types.ts
```diff
@@ -0,0 +1,33 @@
+export interface Agent {
+  name: string;
+  email: string;
+  status: 'available' | 'unavailable' | 'unknown';
+  avatar: string;
+}
+
+export interface InviteAgentsProps {
```
marcinsawicki, line 8: extend `ComponentCoreProps`? 🤔

marcinsawicki (approved):
left some comments

VadymBezpalko:
> IMO it would be better to remove `onSetUpChatbotClick` from the props and allow to pass an array of custom elements
> 
> ```
>   <InviteAgents
>     agents={agents}
>     onAddTeammateClick={onAddTeammateClick}
>     customElements={[
>        onClick: onSetUpChatbotClick,
>        icon: ChatbotIcon,
>        text: 'Set up ChatBot'
>     ]}
>     animatedInviteButton={true}
>     tooltipArrowOffset={13}
>     className="invite-agents"
>   />
> ```

@mreszke The problem is that we have a 'Set up ChatBot' CTA in a dedicated tooltip, so it needs to be passed either way.
image

packages/react-components/src/components/InviteAgents/InviteAgents.tsx
VadymBezpalko: There are at least a few different places with this copy, so if we want to move it to props then other places should be covered too, so I'd leave it like that for now

PR #1372, opened 2024-09-30
feat(ProductSwitcher): Add prop to change the theme mode

jakubsikora:
Resolves: #{issue-number}

## Description
Add prop to change the theme mode. TeamChat doesnt have the app frame yet, so the light mode will work.

## Storybook

https://feature-product-switcher-theme-prop--613a8e945a5665003a05113b.chromatic.com

## Checklist

**Obligatory:**

- [x] Add correct label


VadymBezpalko (approved):
👏

Issue #1326, opened 2024-09-10
[Popover] - New variant for Research needs


vladko-uxds:
https://livechatinc.atlassian.net/browse/PP-11581

[Figma design](https://www.figma.com/design/EnrJJg0SP5K0am1JXD6otJ/%5BINT%5D-Modal-inviting-to-interview?node-id=21-3&t=VawJNekp19wioX1j-4)

PR #1376, opened 2024-10-02
tweak(PromoBannerV2): new kind "dark" for PromoBannerV2 

VadymBezpalko:
Resolves: #1326 

## Description
Adding new kind dark for PromoBannerV2

## Storybook

https://feature-1326--613a8e945a5665003a05113b.chromatic.com

## Checklist

**Obligatory:**

- [x] Add correct label

### DIFF
--- a/packages/react-components/src/components/PromoBannerV2/PromoBannerV2.stories.css
+++ b/packages/react-components/src/components/PromoBannerV2/PromoBannerV2.stories.css
@@ -38,3 +38,19 @@
 .text {
   font-size: 24px;
 }
+
+.dark-promo-banner {
+  max-width: 800px;
+}
+
+.dark-promo-header {
+  margin: 0 0 var(--spacing-4) 0;
+}
+
+.dark-promo-additional-content {
+  align-items: center;
+
+  img {
+    border-radius: 12px;
+  }
+}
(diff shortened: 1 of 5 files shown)


vladko-uxds (approved):
👍

packages/react-components/src/components/PromoBannerV2/PromoBannerV2.stories.css
```diff
@@ -38,3 +38,19 @@
...
+}
+
+.dark-promo-header {
+  margin: 0 0 var(--spacing-4) 0;
+}
+
+.dark-promo-additional-content {
+  align-items: center;
+
+  img {
```
marcinsawicki, line 53: Are we sure that we want to force the radiuses on the img? 🤔 
What about cases where this will be unexpected?

marcinsawicki (approved):

packages/react-components/src/components/PromoBannerV2/PromoBannerV2.stories.css
VadymBezpalko: Do we do not want to do that, so that's why radius is added in custom story styles and not in the component ;)

PR #1388, opened 2024-10-09
fix(InviteAgents): hiding bots info in tooltip option

VadymBezpalko:
Resolves: #{issue-number}

## Description
Hiding bots info in tooltip option

## Storybook

https://feature-invite-agents-tweak--613a8e945a5665003a05113b.chromatic.com

## Checklist

**Obligatory:**

- [ ] Add correct label


vladko-uxds (approved):
👍

Issue #1405, opened 2024-10-15
[FloatingPortal] - add a component that allows breaking out of current stacking context


ashbork:
### Description

Add a component allowing other elements, not just Modals (using `ModalPortal`) to be 'portaled' out of their DOM location

PR #1408, opened 2024-10-16
feat(FloatingPortal): add a FloatingPortal util component and allow using it with Tooltip

ashbork:
allows lifting the tooltip's floating part out of the hierarchy

Resolves: #1405

## Storybook

https://feature-floating-portal--613a8e945a5665003a05113b.chromatic.com

## Checklist

**Obligatory:**

- [x] Add correct label


packages/react-components/src/components/FloatingPortal/types.ts
```diff
@@ -0,0 +1,20 @@
+import * as React from 'react';
+
+export interface FloatingPortalProps {
```
VadymBezpalko, line 3: Maybe it would be better to use the original FloatingPortalProps here?

ashbork: floating-ui doesn't export them

VadymBezpalko (approved):

Issue #1238, opened 2024-07-09
Multiple components - broaden type definitions for usage flexibility 


ashbork:
### Description

[FormField](https://design.livechat.com/?path=/docs/forms-formfield--docs), [Radio button](https://design.livechat.com/?path=/docs/forms-radio-button--docs) and [Checkbox](https://design.livechat.com/?path=/docs/forms-checkbox--docs) have very restrictive typed for props like `description`. This makes it impossible to, for example, add a tooltip to the label of a form field without ignoring type errors, while the component renders with arbitrary `ReactNode`s correctly

PR #1411, opened 2024-10-17
fix(description): broading field type to ReactNode

VadymBezpalko:
Resolves: #1238 

## Description
Allowing to pass a react node type to description fields

## Storybook

https://feature-1238--613a8e945a5665003a05113b.chromatic.com

## Checklist

**Obligatory:**

- [x] Add correct label


marcinsawicki (approved):

vladko-uxds (approved):
👍

Issue #817, opened 2023-10-09
[Tooltip] - Interactive tooltip types


MrVorgoth:
### Description

I wanted to use icon inside `primaryButton` prop inside `Interactive` tooltip component, however I stumbled upon such props: `handleClick` / `label` / `kind (optional)`. I thought I will be able to just use any other `Button` component inside `primaryButton` prop, but I couldn't do it. Can you type that any `ReactNode` can go there?

Also I wonder why it was designed to have `<Interactive />` component placed inside `<Tooltip />` component. As a developer I thought I would use it the same way I use normal tooltip: `<InteractiveTooltip (all props from normal tooltip + props from interactive tooltip />`. Can you think about such implementation?

UPDATE: Please also export a `TooltipButton` type.

vladko-uxds:
Improve storybook doc : how to use formfield +input

PR #1412, opened 2024-10-18
tweak(Tooltip): better docs; marking closeWithX as deprecated

VadymBezpalko:
Resolves: #817

## Description

- Marking `closeWithX` as deprecated prop. Using `handleCloseAction` should be enough to show the close button.
- Preparing better docs with examples for Tooltip subcomponents and their use cases.

## Storybook

https://feature-817--613a8e945a5665003a05113b.chromatic.com

## Checklist

**Obligatory:**

- [x] Add correct label


vladko-uxds (approved):
👍

PR #1423, opened 2024-11-06
fix(ProductSwitcher): update Text Accounts name and icon

### LINKED ISSUE #1415: [ProductSwitcher] - Updates
Copy update
https://text.slack.com/archives/C07CCL0CAPP/p1729591147790789

robertp-lc:
Resolves: #1415

## Description
- updated name for Text Accounts and changed icon

Note: I used existing icon in DS instead of new assets available on cdn server. I initially wanted to but it would require converting svg from url to react component which will be done anyway in the future when wrapper component for those assets will be introduced in DS. Hope it will be good enough for the time being. 

## Storybook

https://feature-1415--613a8e945a5665003a05113b.chromatic.com

## Checklist

**Obligatory:**

- [X] Self review (use this as your final check for proposed changes before requesting the review)
- [X] Add correct label
- [X] Assign pull request with the correct issue

### DIFF
--- a/packages/react-components/src/components/ProductSwitcher/constants.ts
+++ b/packages/react-components/src/components/ProductSwitcher/constants.ts
@@ -5,7 +5,7 @@ import {
   KnowledgeBaseMono,
   LiveChatMono,
   OpenWidgetMono,
-  TextAccounts,
+  TextLogo,
 } from '@livechat/design-system-icons';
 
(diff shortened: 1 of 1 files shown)


VadymBezpalko (approved):

vladko-uxds (approved):
👍

Issue #1486, opened 2025-01-15
[User Guide V2] - new look and feel


vladko-uxds:
Explore and prepare new version of User guide component "Conversational UI"

Based on draft screens:
![Image](https://github.com/user-attachments/assets/5d270f83-b966-492e-82e5-96d38b330ecb)
![Image](https://github.com/user-attachments/assets/1223d3e7-3f19-4e7e-8b38-58039c3f258b)

More details coming soon
- Animated cursor moving
- Animation tour position

PR #1487, opened 2025-01-16
feat(UserGuide): update UserGuide component with new highlight functionality

JoannaSikora:
Resolves: #1486 

## Description
Updated version of `UserGuide`:
- `UserGuideStep` component - UI update with new designs
- `UserGuide` component - UI update with new designs (floating element with animated arrow), updated the component API by removing unnecessary properties

I moved `UserGuide` to its own stories as it no longer relates to the `Tooltip`.
Added advanced example of implementation.

## Storybook

https://feature-user-guide--613a8e945a5665003a05113b.chromatic.com/?path=/story/components-userguide--example

## Checklist

**Obligatory:**

- [ ] Add correct label

## Release Notes

- **New Features**
  - Added `AnimatedTextContainer` component for creating typing animations.
  - Introduced `UserGuide` component with step-by-step guidance functionality.
  - Added `UserGuideBubbleStep` and `UserGuideStep` components for interactive user guidance.

- **Improvements**
  - Enhanced component accessibility by adding optional `id` props across multiple components.
  - Added new transition duration CSS variables for more granular animation control.

- **Styling**
  - Introduced new shadow and transition CSS variables.
  - Added SCSS modules for `UserGuide` and related components.

- **Removals**
  - Removed previous user guide tooltip implementation.
  - Deleted unused user guide components and related files.

- **Documentation**
  - Added comprehensive documentation for new components.
  - Created Storybook stories for `AnimatedTextContainer` and `UserGuide`.


packages/react-components/src/components/AppFrame/components/NavigationItem/NavigationItem.tsx
```diff
@@ -77,6 +78,7 @@ export const NavigationItem: React.FC<INavigationItemProps> = ({
         triggerRenderer={
           <>
             <a
+              // id={id}
               tabIndex={disabled ? -1 : 0}
```
JoannaSikora, line 82: Please remove unnecessary comment

packages/react-components/src/components/AppFrame/components/NavigationTopBar/NavigationTopBar.tsx
```diff
@@ -35,11 +35,12 @@ export const NavigationTopBar = ({
   children,
   className,
   additionalNodes,
+  id,
```
JoannaSikora, line 38: `Id` shouldn't be optional here like in the other components?

packages/react-components/src/components/UserGuide/UserGuide.module.scss
```diff
@@ -0,0 +1,153 @@
...
+        transform: rotate(-50deg);
+      }
+
+      &--right {
+        top: calc(50% - #{$arrow-size / 2});
+        transform: rotate(-90deg);
+      }
+
+      &--right-end {
+        top: calc(100% - 45px - 9px);
```
JoannaSikora, line 50: why not `100% - 54px` ?

packages/react-components/src/components/UserGuide/UserGuide.tsx
```diff
@@ -0,0 +1,247 @@
...
+          >
+            <div ref={contentRef} className={styles[`${baseClass}__guide`]}>
+              <div
+                className={cx(
+                  styles[`${baseClass}__guide__arrow`],
+                  styles[`${baseClass}__guide__arrow--${placement}`],
+                  styles[`${baseClass}__guide__arrow--${cursorTiming}`]
+                )}
+              >
+                <svg
```
JoannaSikora, line 158: Can we extract this svg to a separate component?

packages/react-components/src/components/UserGuide/stories-helpers.css
```diff
@@ -0,0 +1,84 @@
...
+}
+
+.column-header.center {
+  border-bottom: 0;
+}
+
+.chats-list-item {
+  display: flex;
+  flex-direction: row;
+  padding: 19px 24px;
```
JoannaSikora, line 53: 24px => --spacing-6

packages/react-components/src/components/UserGuide/stories-helpers.css
```diff
@@ -0,0 +1,84 @@
...
+}
+
+.chats-list-item-content {
+  display: flex;
+  flex-direction: column;
+  margin-left: 16px;
```
JoannaSikora, line 59: 16px => --spacing-4

JoannaSikora (commented):
👍🏼

packages/react-components/src/components/UserGuide/UserGuide.tsx
```diff
@@ -0,0 +1,247 @@
...
+      />
+    );
+  };
+
+  return parentElement ? (
+    <>
+      <div
+        className={cx(styles[`${baseClass}__overlay`])}
+        style={{
+          display: isVisible ? 'block' : 'none',
```
MateuszRomek, line 130: Can we add an option for a custom z-index on the overlay? Sometimes some elements have bigger zindex than overlay itself. I assume that the same index will be used for cloned element

packages/react-components/src/components/UserGuide/UserGuide.tsx
```diff
@@ -0,0 +1,247 @@
...
+
+const baseClass = 'user-guide';
+
+const virtualReference = (element: Element, padding: number = 0) =>
+  new VirtualReference(element, padding);
+
+export const UserGuide: FC<PropsWithChildren<IUserGuide>> = ({
+  className,
+  children,
+  cursorPosition = 'bottom',
```
MateuszRomek, line 20: In this case, do we completely abandon the previous look and behavior of UserGuide?

I ask about this because Agent App uses the current UserGuide component in its current app tours (which will stay), so this is partly a breaking change and a possible regression for all tours in the application

packages/react-components/src/components/AppFrame/components/NavigationTopBar/NavigationTopBar.tsx
marcinsawicki: It's defined as optional in the interface

packages/react-components/src/components/UserGuide/UserGuide.tsx
marcinsawicki: AFAIK all tours should be verified so this breaking change should go hand in hand with this verification. 🤷

MateuszRomek (approved):

packages/react-components/src/components/UserGuide/UserGuide.tsx
MrVorgoth: Wouldn't it be better to prepare new component like `v2` in this case? I can see that there are many tours to check:
`export const enum UserGuidedTourType {
  ChatCannedResponses = 'ChatCannedResponses',
  CustomersList = 'CustomersList',
  CustomSegments = 'CustomSegments',
  TicketPrivacy = 'TicketPrivacy',
  TicketFormSettings = 'TicketFormSettings',
  TicketStatus = 'TicketStatus',
  TagsUsage = 'TagsUsage',
  TargetedMessages = 'TargetedMessages',
  HomeScreenExplore = 'HomeScreenExplore',
  OpenChat = 'OpenChat',
  UnassignedChat = 'UnassignedChat',
  RichGreeting = 'RichGreeting',
  ReportViews = 'ReportViews',
  RoutingRules = 'RoutingRules',
  Traffic = 'Traffic',
}`

MateuszRomek: ☝️ What's the reasoning for changing the current component and not creating the V2 version like with PromoBanner? It automatically creates dependencies and possible blockers for product teams to introduce this

packages/react-components/src/components/UserGuide/UserGuide.mdx
```diff
@@ -0,0 +1,205 @@
...
+
+<Canvas of={Stories.Default} />
+
+The above example only shows the way of implementation, state management is on the implementation side. 
+You can manage it in any way, the most important is to place in `UserGuide` an element that is to be 
+displayed next to the element the guide points to. We suggest using `UserGuideStep`.
+
+## Style highlighted element <a id="style" />
+
+Component allows to easly style the highlighted element, by passing own styles properties. To do this, you need to enchant the `UserGuide` component with styles object using the `elementStyles` prop:
```
MrVorgoth, line 175:
```suggestion
Component allows to easily style the highlighted element, by passing own styles properties. To do this, you need to enchant the `UserGuide` component with styles object using the `elementStyles` prop:
```

packages/react-components/src/components/UserGuide/types.ts
```diff
@@ -0,0 +1,36 @@
...
+   */
+  cursorPosition?: Placement;
+  /**
+   * The timing for the floating element transition
+   */
+  cursorTiming?: CursorTiming;
+  /**
+   * The id of the element to highlight
+   */
+  parentElementName?: string;
```
MrVorgoth, line 27: Shouldn't `parentElementName`, `cursorPosition` and `isVisible` be required props? 🤔

packages/react-components/src/components/UserGuide/UserGuide.tsx
```diff
@@ -0,0 +1,182 @@
...
+      };
+    }
+  }, [parentElement]);
+
+  useEffect(() => {
+    const handleResize = () => update();
+    window.addEventListener('resize', handleResize);
+
+    return () => window.removeEventListener('resize', handleResize);
+  }, [update]);
```
MrVorgoth, line 66:
```suggestion
  useEffect(() => {
    window.addEventListener('resize', update);

    return () => window.removeEventListener('resize', update);
  }, [update]);
```

packages/react-components/src/components/UserGuide/UserGuide.tsx
```diff
@@ -0,0 +1,182 @@
...
+      Object.assign(clonedElement.style, elementStyles);
+      containerRef.current.appendChild(clonedElement);
+    }
+
+    return () => {
+      if (containerRef.current) {
+        containerRef.current.innerHTML = '';
+      }
+    };
+  }, [parentElement, containerRef.current, isVisible]);
```
MrVorgoth, line 96: Why there is `isVisible` in dependency array? It's not used in above effect 🤔

packages/react-components/src/components/UserGuide/UserGuide.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+  useEffect(() => {
+    const handleResize = () => update();
+    window.addEventListener('resize', handleResize);
+
+    return () => window.removeEventListener('resize', handleResize);
+  }, [update]);
+
+  useEffect(() => {
+    if (parentElementName) {
```
MrVorgoth, line 69: I wonder if we should allow to have optional props for `parentElementName`? Is tour without highlighting elements even an option and can happen? 🤔

MrVorgoth (approved):
Nice job 🎉

packages/react-components/src/components/UserGuide/UserGuide.tsx
dprzybylek: @MrVorgoth @MateuszRomek  I think most of these tours might have only one step. From my experience, we tend to forget about tours and don't test them - so in my opinion, it's a good idea to introduce one breaking change and clean up the whole app. Recently, I deleted two or three tours just because they didn't make sense. Let's check the list; for example, I see that tickets should be removed as we will replace tickets with helpdesk.

MateuszRomek: @dprzybylek it's not about changing these tours but it's design decision. Marta will talk to Vlad about this as this must be product designers decision to make all of the tours like this onboarding tour and remove tooltip approach:) 
AFAIK there were no discussions about this

packages/react-components/src/components/UserGuide/components/UserGuideStep.module.scss
```diff
@@ -0,0 +1,40 @@
+$base-class: 'user-guide-step';
+
+.#{$base-class} {
+  display: flex;
+  flex-direction: column;
+  border-radius: var(--radius-4);
+  background: linear-gradient(45deg, #cddef9, #e7e5fd, #f9f2ff);
```
dprzybylek, line 7: Shouldn't it be a theme variable?

packages/react-components/src/components/UserGuide/types.ts
marcinsawicki: Good point, but we currently working on the "first step" and "last step" features, which will not require the element to highlight, as well as other positioning props.

MichalPaszowski (approved):

packages/react-components/src/components/UserGuide/UserGuide.tsx
marcinsawicki: Explained above

PR #1713, opened 2026-05-05
feat(icons): change TextLogoFull to new one

jpyzio123:
Resolves: #{issue-number}

## Description
Update TextLogoFull to the new one, same as it is in Text Design System. 

OLD:
image

NEW: 
image

## Storybook

https://feature-ac-2963-text-logo--613a8e945a5665003a05113b.chromatic.com

## Checklist

**Obligatory:**


JoannaSikora (approved):
I think that we should also update TextLogoFilled alongside it cc: @vladko-uxds 

Image

